Issue #42, opened 2014-08-05
add explicit device selection


camlorn:
I'm looking at using Cubeb for my own project, and one of the features I wish to support is explicit device selection.  If a standard API can be decided upon, I can do the work for the Windows backends, but am not yet in the position to do so for Linux, mac, and Android.  I'm seriously considering doing the work on Windows for my own project; consequently, I want to hash out what would be needed to get it upstream.

ieei:
1y later - I've been working on an enumeration API. This could then again be used for device selection. (Which is the next thing I was thinking of doing.) See https://github.com/ieei/cubeb/commits/dev_enum

kinetiknz:
This is in now.

PR #58, opened 2015-09-10
Enumerate devices

ieei:
This PR adds:
- New API for enumerating devices.
  * DeviceID, friendly name, group and [vendor]
  * Channels, sample rate and format
  * State (disable, unplugged, enabled)
  * Preferred device
  * Latency
- a demo/example/test app; test_devices
- implementation for the backends:
  * audiounit (tested on OSX)
  * pulse (tested on Ubuntu)
  * winmm
  * wasapi

The plan behind this is to later add API for opening a stream on a specific device. (Issue #42) Also API for opening a capture stream or full duplex stream as we have been discussing in the WebRTC group for Firefox.

Output from test_devices:
[OSX - audiounit]
```
Enumerating input devices for backend audiounit
Found 1 input devices
dev: "AppleHDAEngineInput:1B,0,1,0:1" (PREFERRED)
	Name:    "Built-in Microphone (Internal Microphone)"
	Group:   "AppleHDAEngineInput:1B,0,1,0:1"
	Vendor:  "Apple Inc."
	Type:    input
	State:   enabled
	Ch:      2
	Format:  S16LE S16BE F32LE F32BE (0x3030) (default: F32LE)
	Rate:    44100 - 96000 (default: 44100)
	Latency: lo 1ms, hi 94ms
Enumerating output devices for backend audiounit
Found 2 output devices
dev: "AppleHDAEngineOutput:1B,0,1,1:0" (PREFERRED)
	Name:    "Built-in Output (Headphones)"
	Group:   "AppleHDAEngineOutput:1B,0,1,1:0"
	Vendor:  "Apple Inc."
	Type:    output
	State:   enabled
	Ch:      2
	Format:  S16LE S16BE F32LE F32BE (0x3030) (default: F32LE)
	Rate:    44100 - 96000 (default: 44100)
	Latency: lo 2ms, hi 95ms
dev: "AppleHDAEngineOutputDP:0,1,0,1,3:0:{2D4C-0BFD-00000000}"
	Name:    "DisplayPort (SE790C)"
	Group:   "AppleHDAEngineOutputDP:0,1,0,1,3:0:{2D4C-0BFD-00000000}"
	Vendor:  "Apple Inc."
	Type:    output
	State:   enabled
	Ch:      2
	Format:  S16LE S16BE F32LE F32BE (0x3030) (default: F32LE)
	Rate:    32000 - 48000 (default: 48000)
	Latency: lo 2ms, hi 87ms
```
[Win7 - wasapi]
```
Found 3 input devices
dev: "{0.0.1.00000000}.{35b0dcdc-747a-4103-972f-7d57f2033e8a}"
        Name:    "Internal AUX Jack (High Definition Audio Device)"
        Group:   "HDAUDIO\FUNC_01&VEN_8384&DEV_7680&SUBSYS_83847680&REV_1034\4&31e60982&0&0001"
        Vendor:  "(null)"
        Type:    input
        State:   disabled
        Ch:      0
        Format:  F32LE (0x1000) (default: F32LE)
        Rate:    0 - 0 (default: 0)
        Latency: lo 0ms, hi 0ms
dev: "{0.0.1.00000000}.{dcc7610d-6fda-4167-bdcc-24812d4854d1}"
        Name:    "CD Audio (High Definition Audio Device)"
        Group:   "HDAUDIO\FUNC_01&VEN_8384&DEV_7680&SUBSYS_83847680&REV_1034\4&31e60982&0&0001"
        Vendor:  "(null)"
        Type:    input
        State:   disabled
        Ch:      2
        Format:  F32LE (0x1000) (default: F32LE)
        Rate:    44100 - 44100 (default: 44100)
        Latency: lo 0ms, hi 0ms
dev: "{0.0.1.00000000}.{fe4076a9-92dd-49e2-9428-47b0a4f318ef}" (PREFERRED)
        Name:    "Microphone (High Definition Audio Device)"
        Group:   "HDAUDIO\FUNC_01&VEN_8384&DEV_7680&SUBSYS_83847680&REV_1034\4&31e60982&0&0001"
        Vendor:  "(null)"
        Type:    input
        State:   enabled
        Ch:      2
        Format:  F32LE (0x1000) (default: F32LE)
        Rate:    44100 - 44100 (default: 44100)
        Latency: lo 3ms, hi 10ms
Enumerating output devices for backend wasapi
Found 2 output devices
dev: "{0.0.0.00000000}.{14960cf0-7a7d-40f7-a72b-16808cf1654c}" (PREFERRED)
        Name:    "Speakers (High Definition Audio Device)"
        Group:   "HDAUDIO\FUNC_01&VEN_8384&DEV_7680&SUBSYS_83847680&REV_1034\4&31e60982&0&0001"
        Vendor:  "(null)"
        Type:    output
        State:   enabled
        Ch:      2
        Format:  F32LE (0x1000) (default: F32LE)
        Rate:    44100 - 44100 (default: 44100)
        Latency: lo 3ms, hi 10ms
dev: "{0.0.0.00000000}.{3f23870d-2b99-4b5c-9873-e2da2cc79dfa}"
        Name:    "Headphones (High Definition Audio Device)"
        Group:   "HDAUDIO\FUNC_01&VEN_8384&DEV_7680&SUBSYS_83847680&REV_1034\4&31e60982&0&0001"
        Vendor:  "(null)"
        Type:    output
        State:   unplugged
        Ch:      0
        Format:  F32LE (0x1000) (default: F32LE)
        Rate:    0 - 0 (default: 0)
        Latency: lo 0ms, hi 0ms
```


ieei:
I'm open to discussion on the API itself.
- I'm not sure about the type of the device_id. Some backends have this is a string, others an index.
- I've added some new types, which could be reused/shared. Like sample format.

ieei:
I've been using MSVC to build the windows backends. I see that the CI mingw distro lacks some of wasapi/wincoreaudio headers or parts...

kinetiknz:
Thanks for doing this!  Some questions and comments below.

Interface:

- I think it'd be nicer to use an index-based device query, where you'd have an enumerate call that returns a count and a way to fetch a device by index.  This avoids exposing a list type via the API/ABI.  A stream can then be created on a particular device by specifying either the index or passing a reference to the device returned by the index-query function.

- It's not really clear to me what cubeb_device_id_to_str is intended to be used for.
  Also, there may be some complexity around expecting the caller to free memory allocated within a DLL on Win32 (e.g. different allocators), but I'm not sure if this is still an issue with modern systems.

- cubeb_register_device_list_changed/cubeb_deregister_device_list_changed should be merged into a single function called cubeb_register_device_list_changed_callback where registering a NULL callback has the effect of deregistering any existing callback -- this matches the behaviour of the existing function cubeb_stream_register_device_changed_callback.

- The sample formats/channels/rates exposed in cubeb_device_info may not be able to represent all configurations adequately.  e.g. for channels and rate the current representation implies that all values between the min and max may be supported, but a given backend may only be able to support a discrete set.  It also seems like it's possible that a particular sample format or number of channels may only be supported at a particular sample rate.  Depending on the intended use case, maybe it's simpler to leave this information out and/or make it explicit in the documentation that there's no guarantee that an arbitrary sample format/rate/channel/latency configuration within the specified min/max is supported.
  - Change the representation
  - Add an is_format_supported query that returns the devices that can support the requested format
    (Or can be called on a specific device and returns a yes/no answer (and possibly a nearest supported recommendation))
  - Remove these details until we have a use for them, then reinstate with a different representation

- Documentation of intended use of cubeb_device_info fields, particularly I'm unclear about vendor_name, but I also wonder if group_id would be better as a standalone equality function rather than expecting the caller to use strcmp/memcmp -- e.g. cubeb_device_groups_equal(device1, device2)


Implementation:

- Case in Doxygen comments
- No spaces between function name and (
- Seems like the KAI backend could have NULLs instead of stub functions in the ops struct

- Can caller expect device_list_changed_callback to be called if state/attributes of devices change, e.g. if the default sample rate for a device changed or a device is disconnected, do/should you get a callback?

ieei:
The API is sort of implemented with getUserMedia and w3c media capture and streams in the back of my head. http://www.w3.org/TR/mediacapture-streams/#device-info

vendor_name could be removed, it is anyway only informative, and shouldn't be needed for anything of value. w3c defines both deviceId and groupId as strings, which makes sense to me at least.

An index based device query is exactly what I think we should avoid!
1) Newer audio APIs doesn't use this. So the backend would need to emulate indices.
2) The idea with index based device id is truly flawed. What happens when a device is removed? Say index 1 on a system with three devices. Is index 2 now index 1?

So I added a cubeb_devid which is supposed to be an opaque type. This type/handle would be what could be used to create a stream with a particular device. Going further I was thinking that we should add a new function to create a new stream and keep cubeb_stream_init() which create the stream using the default device.
cubeb_device_id_to_str() takes the opaque type and returns a c string which I'm using in the test application is used to print out device id for the particular backend. It could also be used for MediaDeviceInfo.deviceId. I should probably have added means to free it as well, as you point out. 
I understand this might seem like it is complicating the API, but it will cover the legacy APIs actually using index based devices ids, and the newer ones where you use an unique device ID for identification.
Anyway, if you really want me to change to index based device IDs, I'll just go ahead and change it. Let me know.

Regarding cubeb_register_device_list_changed():
I haven't done too much thinking around this yet. Just added the interface, which I probably shouldn't have done. Although your suggestion about one function to register/install or deregister/uninstall a callback seems odd. This would mean you only get to register one callback. But yeah, one is probably enough anyway.

Good points about format/rate/channels. I think a new query for supported format should be added! Although I like the idea of indicating what formats, rates and channels are supported, as it is now. But thats probably because I like seeing such values in the test application. Should I remove max/min rates and max channels?

kinetiknz:
I'm okay with leaving vendor_name present if it's useful for the MediaDeviceInfo stuff.  Makes more sense with that context, thanks.

I'm not sure there's a huge distinction between index vs list, except that an index allows you to avoid exposing a list type.  The indices and list/iterators would become invalid when a device is removed; I don't see how the current API allows the list to mutate safely while it's being iterated by a library caller anyway.  My thinking here is mostly influenced by existing libraries (PortAudio, libsoundio, etc. use indexed devices, libsoundio's are reference counted as a way to deal with device list mutation during iteration).

A new variation of cubeb_stream_init that allows you to specify a device makes sense, I assumed that'd come later as a natural extension of your work on the device enumeration stuff.  Leaving cubeb_stream_init to use the default sounds fine, but we could also add a new argument and use NULL for the default... the API is not particularly stable.

A single callback doesn't seem like a large limitation and we've already somewhat committed to this route without any problems so far with the existing callback registration.  Multiple API users are sort-of expected to use separate cubeb contexts in the first place, so they'd have their own slot to register a callback.  We can also change the existing interface to fit your suggested model... I don't care which we use so much as I care that they're consistent.

It's fine to leave the existing format/rate/channels stuff so long as it's clear that it's informational and not a canonical list of supported formats (which we can add/extend later as necessary).

ieei:
When I say index based API, I mean what's given to the upcoming `cubeb_stream_init_on_device` (or whatever clever name we come up with) which identifies the device. Port audio uses an index `PaDeviceIndex`. libsoundio uses a pointer to it's internal created  `struct SoundIoDevice`.

**port audio** is creating its internal array when initializing. So it doesn't handle devices being added (or removed for that matter). Using a device that's no longer present will probably just fail. So in this case it makes sense to use an index based API. It lacks device changed notification, but that's probably ok. Adding device notification will be a problem for port audio as far as I can tell. I guess the only way to do it now is to deinitialize and then initialize the library again. Which probably means stopping all audio streams.
 
**libsound.io** does it by exposing it's internal cached list/array of devices in the form of `SoundIoDevice` which is, yes - reference counted, but also queried by an index. The internal `safe_devices_info` will only update when you call either `soundio_flush_events` or `soundio_wait_events` as it says in the documentation. So it is thread safe in this sense. There is also [soundio_force_device_scan](http://libsound.io/doc/soundio_8h.html#a7c04afbb72a9ae76e1934b62916670cb). It looks good to me, but from my point of view, this API is only usable from one thread. If you want to call soundio API from multiple threads you should be very careful. Which is probably the philosophy of the API. BTW, libsound.io also has the notion of device id.
> Exposes both device id and friendly name. id you could save in a config file because it persists between devices becoming plugged and unplugged, while friendly name is suitable for exposing to users. 

With libsoundio if you query default (input/output) device and store it, and then call something that will update the internal state, you basically have to update the cached default device. Also in theory (especially on osx) the default device might change at any time. E.g. when you are plugging in your Apple Cinema display. In my opinion, opening the default device should be different from opening a specific device. Just as we were discussing/planning.

**cubeb** (or the API I'm proposing) gives you a snapshot in the form of a list. (I can obviously change it to an array. I don't see the real value or issue with a list vs array) The list is owned by the caller and should be freed whenever needed, so it will never mutate. It doesn't expose anything internal that are subject to change at any time. So what I am proposing is that the `cubeb_devid` type could be used when creating/initializing a stream. So the `cubeb_device_info` struct is totally isolated to enumeration and the field `devid` might be used as input to `cuebb_init_stream_on_device` (or whatever it will be). Obviously a devices could be removed/unplugged between the time of enumeration and initializing a stream on this device. Then the init call will fail.

kinetiknz:
Thanks, I think we're pretty much on the same page now.

I still think there's small but worthwhile value in not exposing a list as it makes the C API smaller and cleaner which has minor benefits for FFI and future maintainability.  If we can find a way to address that and the callback registration issue, I'm pretty happy with the rest of the API design and code changes.

ieei:
I pushed some patches before the weekend.
I changed the list to a collection/array. So now it is very similar to WASAPI actually. Tell me what you think! :)

kinetiknz:
Thanks, looks great!

PR #63, opened 2016-01-13
Preparatory work for the input and duplex code

padenot:
This is changing all the signatures of the `cubeb_stream_init` implementations,
the signature of the `data_callback` type, so that cubeb can support audio
input.

`cubeb_stream_init` now has two `cubeb_stream_params` pointers, one for input,
one for output. If two pointers are passed, a "duplex" stream is opened. If only
one pointer is passed, an input-only or output-only stream is created.

Duplex streams have the same sample rate, and sample type. They don't have to
have the same number of channels.

`data_callback` now has two pointers to audio buffers: an input buffer (`NULL`
if this is an output-only stream) containing input data (e.g. a microphone), and
an output buffer, to be filled, as usual, with the audio frames to play. The
two buffers always have the exact same number of audio frames, and are
temporally correlated in a way that ensures the minimal loop-back latency on
the system if one directly copies the input buffer to the output buffer.

No functionnal changes are present in this patch, just signature changes.
Asserts have been added to prevent users to try to use the input code path for
now.

Actual implementations with the input code for different platforms will follow.

Green `mozilla-central` push:
<https://treeherder.mozilla.org/#/jobs?repo=try&revision=15b4dd3cbbe8>

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -221,17 +221,21 @@ typedef struct {
 } cubeb_device_collection;
 
 /** User supplied data callback.
-    @param stream
-    @param user_ptr
-    @param buffer
-    @param nframes
-    @retval Number of frames written to buffer, which must equal nframes except
-            at end of stream.
+    @param stream The stream for which this callback fired
+    @param user_ptr The pointer passed to cubeb_stream_create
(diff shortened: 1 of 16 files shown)


kinetiknz:
Thanks!

PR #64, opened 2016-01-15
PulseAudio implementaion for full-duplex

achronop:
This implement the full-duplex functionality for PulseAudio back-end in the way that described in kinetiknz/cubeb#63

Test files has been updated as required and the test-suite runs successfully under Linux/Pulse.

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -118,6 +118,8 @@ typedef enum {
 } cubeb_stream_type;
 #endif
 
+typedef void * cubeb_devid;
+
 /** Stream format initialization parameters. */
 typedef struct {
   cubeb_sample_format format; /**< Requested sample format.  One of
@@ -127,6 +129,7 @@ typedef struct {
 #if defined(__ANDROID__)
   cubeb_stream_type stream_type; /**< Used to map Android audio stream types */
 #endif
+  cubeb_devid devid;
 } cubeb_stream_params;
 
 /** Output device description */
@@ -172,8 +175,6 @@ typedef enum {
 
 /** An opaque handle used to refer a particular input or output device
  *  accross calls. */
-typedef void * cubeb_devid;
-
 /**
  * Architecture specific sample type.
  */
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -118,6 +118,8 @@ typedef enum {
 } cubeb_stream_type;
 #endif
 
+/** An opaque handle used to refer a particular input or output device
+ *  across calls. */
 typedef void * cubeb_devid;
 
 /** Stream format initialization parameters. */
@@ -173,8 +175,6 @@ typedef enum {
   CUBEB_DEVICE_STATE_ENABLED /**< The device is enabled. */
 } cubeb_device_state;
 
-/** An opaque handle used to refer a particular input or output device
- *  accross calls. */
 /**
  * Architecture specific sample type.
  */
--- a/src/cubeb_alsa.c
+++ b/src/cubeb_alsa.c
@@ -85,9 +85,7 @@ struct cubeb_stream {
   snd_pcm_uframes_t write_position;
   snd_pcm_uframes_t last_position;
   snd_pcm_uframes_t buffer_size;
-  snd_pcm_uframes_t period_size;
-  cubeb_stream_params input_params;
-  cubeb_stream_params output_params;
+  cubeb_stream_params params;
 
   /* Every member after this comment is protected by the owning context's
      mutex rather than the stream's mutex, or is only used on the context's
@@ -318,14 +316,14 @@ alsa_refill_stream(cubeb_stream * stm)
   if (got > 0) {
     snd_pcm_sframes_t wrote;
 
-    if (stm->output_params.format == CUBEB_SAMPLE_FLOAT32NE) {
+    if (stm->params.format == CUBEB_SAMPLE_FLOAT32NE) {
       float * b = (float *) p;
-      for (uint32_t i = 0; i < got * stm->output_params.channels; i++) {
+      for (uint32_t i = 0; i < got * stm->params.channels; i++) {
         b[i] *= stm->volume;
       }
     } else {
       short * b = (short *) p;
-      for (uint32_t i = 0; i < got * stm->output_params.channels; i++) {
+      for (uint32_t i = 0; i < got * stm->params.channels; i++) {
         b[i] *= stm->volume;
       }
     }
@@ -340,7 +338,7 @@ alsa_refill_stream(cubeb_stream * stm)
   }
   if (got != avail) {
     long buffer_fill = stm->buffer_size - (avail - got);
-    double buffer_time = (double) buffer_fill / stm->output_params.rate;
+    double buffer_time = (double) buffer_fill / stm->params.rate;
 
     /* Fill the remaining buffer with silence to guarantee one full period
        has been written. */
@@ -782,19 +780,17 @@ alsa_destroy(cubeb * ctx)
 static void alsa_stream_destroy(cubeb_stream * stm);
 
 static int
-alsa_stream_init(cubeb * ctx,
-                 cubeb_stream ** stream,
-                 char const * stream_name,
+alsa_stream_init(cubeb * ctx, cubeb_stream ** stream, char const * stream_name,
                  cubeb_stream_params * input_stream_params,
                  cubeb_stream_params * output_stream_params,
                  unsigned int latency,
-                 cubeb_data_callback data_callback,
-                 cubeb_state_callback state_callback,
+                 cubeb_data_callback data_callback, cubeb_state_callback state_callback,
                  void * user_ptr)
 {
   cubeb_stream * stm;
   int r;
   snd_pcm_format_t format;
+  snd_pcm_uframes_t period_size;
 
   assert(ctx && stream);
 
@@ -834,8 +830,7 @@ alsa_stream_init(cubeb * ctx,
   stm->data_callback = data_callback;
   stm->state_callback = state_callback;
   stm->user_ptr = user_ptr;
-  stm->input_params = *input_stream_params;
-  stm->output_params = *output_stream_params;
+  stm->params = *output_stream_params;
   stm->state = INACTIVE;
   stm->volume = 1.0;
 
@@ -859,14 +854,14 @@ alsa_stream_init(cubeb * ctx,
   }
 
   r = snd_pcm_set_params(stm->pcm, format, SND_PCM_ACCESS_RW_INTERLEAVED,
-                         stm->output_params.channels, stm->output_params.rate, 1,
+                         stm->params.channels, stm->params.rate, 1,
                          latency * 1000);
   if (r < 0) {
     alsa_stream_destroy(stm);
     return CUBEB_ERROR_INVALID_FORMAT;
   }
 
-  r = snd_pcm_get_params(stm->pcm, &stm->buffer_size, &stm->period_size);
+  r = snd_pcm_get_params(stm->pcm, &stm->buffer_size, &period_size);
   assert(r == 0);
 
   stm->nfds = snd_pcm_poll_descriptors_count(stm->pcm);
--- a/src/cubeb_alsa.c
+++ b/src/cubeb_alsa.c
@@ -85,7 +85,9 @@ struct cubeb_stream {
   snd_pcm_uframes_t write_position;
   snd_pcm_uframes_t last_position;
   snd_pcm_uframes_t buffer_size;
-  cubeb_stream_params params;
+  snd_pcm_uframes_t period_size;
+  cubeb_stream_params input_params;
+  cubeb_stream_params output_params;
 
   /* Every member after this comment is protected by the owning context's
      mutex rather than the stream's mutex, or is only used on the context's
@@ -316,14 +318,14 @@ alsa_refill_stream(cubeb_stream * stm)
   if (got > 0) {
     snd_pcm_sframes_t wrote;
 
-    if (stm->params.format == CUBEB_SAMPLE_FLOAT32NE) {
+    if (stm->output_params.format == CUBEB_SAMPLE_FLOAT32NE) {
       float * b = (float *) p;
-      for (uint32_t i = 0; i < got * stm->params.channels; i++) {
+      for (uint32_t i = 0; i < got * stm->output_params.channels; i++) {
         b[i] *= stm->volume;
       }
     } else {
       short * b = (short *) p;
-      for (uint32_t i = 0; i < got * stm->params.channels; i++) {
+      for (uint32_t i = 0; i < got * stm->output_params.channels; i++) {
         b[i] *= stm->volume;
       }
     }
@@ -338,7 +340,7 @@ alsa_refill_stream(cubeb_stream * stm)
   }
   if (got != avail) {
     long buffer_fill = stm->buffer_size - (avail - got);
-    double buffer_time = (double) buffer_fill / stm->params.rate;
+    double buffer_time = (double) buffer_fill / stm->output_params.rate;
 
     /* Fill the remaining buffer with silence to guarantee one full period
        has been written. */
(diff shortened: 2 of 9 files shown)


include/cubeb/cubeb.h
```diff
@@ -172,8 +175,6 @@ typedef enum {
 
 /** An opaque handle used to refer a particular input or output device
  *  accross calls. */
```
kinetiknz: Move the comment with the code.

src/cubeb_alsa.c
```diff
@@ -85,7 +85,9 @@ struct cubeb_stream {
   snd_pcm_uframes_t write_position;
   snd_pcm_uframes_t last_position;
   snd_pcm_uframes_t buffer_size;
-  cubeb_stream_params params;
+  snd_pcm_uframes_t period_size;
```
kinetiknz: I can't see a reason for this being reintroduced.  Looks like a merge error has reverted a change made on master.  This applies to every other occurrence of period_size in the changeset.

kinetiknz: I'd rather see the ALSA changes separately.  I don't think they should be part of this PR.

src/cubeb_pulse.c
```diff
@@ -170,46 +188,52 @@ stream_state_callback(pa_stream * s, void * u)
...
 
   towrite = nbytes;
-
+  read_offset = 0;
   while (towrite) {
     size = towrite;
     r = WRAP(pa_stream_begin_write)(s, &buffer, &size);
+    if (r < 0) {
+      // Never get here in normal scenario
+      LOG("Unexpected error. Debugging with rr causes this\n");
```
kinetiknz: This needs to be tracked down.  What error are you seeing returned here?

The docs don't seem to have any documented error codes, but looking at pulseaudio/src/pulse/stream.c, it can return PA_ERR_FORKED, PA_ERR_BADSTATE, or PA_ERR_INVALID.

src/cubeb_pulse.c
```diff
@@ -170,46 +188,52 @@ stream_state_callback(pa_stream * s, void * u)
...
+      WRAP(pa_stream_cancel_write)(s);
+      stm->shutdown = 1;
+      return;
+    }
     assert(r == 0);
     assert(size > 0);
     assert(size % frame_size == 0);
 
-    got = stm->data_callback(stm, stm->user_ptr, NULL, buffer, size / frame_size);
+    LOG("Trigger user callback with output buffer size %zd, read_offset=%zd\n", size, read_offset);
```
kinetiknz: Be consistent about "var=%" or "var %" in logging.

src/cubeb_pulse.c
```diff
@@ -247,6 +271,79 @@ stream_request_callback(pa_stream * s, size_t nbytes, void * u)
...
 
 static int
+read_from_input(pa_stream * s, const void ** buffer, size_t * size)
+{
+  if (WRAP(pa_stream_readable_size)(s) > 0) {
+    if (WRAP(pa_stream_peek)(s, buffer, size) < 0) {
+      return -1;
+    }
+  }
+  return WRAP(pa_stream_readable_size)(s);
```
kinetiknz: Can you explain the logic here? I don't understand why pa_stream_readable_size is being called twice.

src/cubeb_pulse.c
```diff
@@ -247,6 +271,79 @@ stream_request_callback(pa_stream * s, size_t nbytes, void * u)
...
+{
+  LOG("Input callback buffer size %zd\n", nbytes);
+
+  cubeb_stream * stm;
+  stm = u;
+  if (stm->shutdown) {
+    return;
+  }
+
+  const void *read_data=NULL;
```
kinetiknz: Spaces: const void * read_data = NULL;

src/cubeb_pulse.c
```diff
@@ -247,6 +271,79 @@ stream_request_callback(pa_stream * s, size_t nbytes, void * u)
...
+  size_t read_size;
+  while (read_from_input(s, &read_data, &read_size) > 0) {
+    size_t in_frame_size = WRAP(pa_frame_size)(&stm->input_sample_spec);
+    size_t read_frames = read_size / in_frame_size;
+
+    cubeb_stream * stm = u;
+    if (stm->output_stream) {
+      // input/capture + output/playback operation
+      void * read_buffer = (void*)read_data;
```
kinetiknz: Spaces: (void *) read_data;

src/cubeb_pulse.c
```diff
@@ -170,46 +188,52 @@ stream_state_callback(pa_stream * s, void * u)
...
+      stm->shutdown = 1;
+      return;
+    }
     assert(r == 0);
     assert(size > 0);
     assert(size % frame_size == 0);
 
-    got = stm->data_callback(stm, stm->user_ptr, NULL, buffer, size / frame_size);
+    LOG("Trigger user callback with output buffer size %zd, read_offset=%zd\n", size, read_offset);
+    got = stm->data_callback(stm, stm->user_ptr, (uint8_t *)input_data + read_offset, buffer, size / frame_size);
```
kinetiknz: Spaces: (uint8_t *) input_data

src/cubeb_pulse.c
```diff
@@ -247,6 +271,79 @@ stream_request_callback(pa_stream * s, size_t nbytes, void * u)
...
+    if (stm->output_stream) {
+      // input/capture + output/playback operation
+      void * read_buffer = (void*)read_data;
+      size_t out_frame_size = WRAP(pa_frame_size)(&stm->output_sample_spec);
+      size_t write_size = read_frames * out_frame_size;
+      // Offer full duplex data for writing
+      trigger_user_callback(stm->output_stream, read_buffer, write_size, stm);
+    } else {
+      // input/capture only operation. Call callback directly
+      void * read_buffer = (void*)read_data;
```
kinetiknz: Ditto.

src/cubeb_pulse.c
```diff
@@ -247,6 +271,79 @@ stream_request_callback(pa_stream * s, size_t nbytes, void * u)
...
+  stm = u;
+  if (stm->shutdown) {
+    return;
+  }
+
+  const void *read_data=NULL;
+  size_t read_size;
+  while (read_from_input(s, &read_data, &read_size) > 0) {
+    size_t in_frame_size = WRAP(pa_frame_size)(&stm->input_sample_spec);
+    size_t read_frames = read_size / in_frame_size;
```
kinetiknz: Since read_frames is passed as a long, it should be declared as a long and any computations checked to ensure they fit into a long.

src/cubeb_pulse.c
```diff
@@ -247,6 +271,79 @@ stream_request_callback(pa_stream * s, size_t nbytes, void * u)
...
+      void * read_buffer = (void*)read_data;
+      size_t out_frame_size = WRAP(pa_frame_size)(&stm->output_sample_spec);
+      size_t write_size = read_frames * out_frame_size;
+      // Offer full duplex data for writing
+      trigger_user_callback(stm->output_stream, read_buffer, write_size, stm);
+    } else {
+      // input/capture only operation. Call callback directly
+      void * read_buffer = (void*)read_data;
+      long got = stm->data_callback(stm, stm->user_ptr, read_buffer, NULL, read_frames);
+      size_t u_got = got;
```
kinetiknz: If read_frames was a long, as mentioned above, this wouldn't be necessary.

src/cubeb_pulse.c
```diff
@@ -247,6 +271,79 @@ stream_request_callback(pa_stream * s, size_t nbytes, void * u)
...
+      if ( got < 0 || u_got != read_frames) {
```
kinetiknz: No space: (got

src/cubeb_pulse.c
```diff
@@ -290,16 +404,25 @@ operation_wait(cubeb * ctx, pa_stream * stream, pa_operation * o)
 }
 
 static void
-stream_cork(cubeb_stream * stm, enum cork_state state)
+cork_io_stream(cubeb_stream * stm, pa_stream * io_stream, enum cork_state state)
```
kinetiknz: I'm not sure I see the value in splitting stream_cork (and wait_until_stream_ready) up like this.

src/cubeb_pulse.c
```diff
@@ -290,16 +404,25 @@ operation_wait(cubeb * ctx, pa_stream * stream, pa_operation * o)
...
 {
-  pa_operation * o;
-
-  WRAP(pa_threaded_mainloop_lock)(stm->context->mainloop);
-  o = WRAP(pa_stream_cork)(stm->stream, state & CORK, stream_success_callback, stm);
+  pa_operation * o = NULL;
```
kinetiknz: No need to initialize this, it's clearly initialized on first use.

src/cubeb_pulse.c
```diff
@@ -485,91 +636,153 @@ pulse_destroy(cubeb * ctx)
...
        until some point after initialization has completed. */
-    o = WRAP(pa_stream_update_timing_info)(stm->stream, stream_success_callback, stm);
-    if (o) {
-      r = operation_wait(stm->context, stm->stream, o);
-      WRAP(pa_operation_unref)(o);
-    }
+    r = stream_update_timing_info(stm);
   }
+
+#ifdef LOGGING_ENABLED
```
kinetiknz: It seems like it'd be better to log this after dropping the lock and testing r != 0.

src/cubeb_resampler.h
```diff
@@ -54,7 +54,7 @@ cubeb_resampler * cubeb_resampler_create(cubeb_stream * stream,
  */
 long cubeb_resampler_fill(cubeb_resampler * resampler,
                           void * input_buffer, 
-						  void * output_buffer, long frames_needed);
+                          void * output_buffer, long frames_needed);
```
kinetiknz: This change and the test changes should be in a separate PR as well.

src/cubeb_pulse.c
```diff
@@ -308,6 +431,34 @@ stream_cork(cubeb_stream * stm, enum cork_state state)
...
+{
+  int r = -1;
+  pa_operation * o = NULL;
+  if (stm->output_stream) {
+    o = WRAP(pa_stream_update_timing_info)(stm->output_stream, stream_success_callback, stm);
+    if (o) {
+      r = operation_wait(stm->context, stm->output_stream, o);
+      WRAP(pa_operation_unref)(o);
+    }
+    // if this fail abort
```
kinetiknz: Comment seems unnecessary.

src/cubeb_pulse.c
```diff
@@ -485,91 +636,153 @@ pulse_destroy(cubeb * ctx)
 
 static void pulse_stream_destroy(cubeb_stream * stm);
 
+pa_sample_format_t
+cube_to_pulse_format(cubeb_sample_format format)
```
kinetiknz: cubeb not cube

src/cubeb_pulse.c
```diff
@@ -485,91 +636,153 @@ pulse_destroy(cubeb * ctx)
...
+    return PA_SAMPLE_FLOAT32BE;
+  default:
+    return PA_SAMPLE_INVALID;
+  }
+}
+
+static pa_stream *
+create_pa_stream(cubeb_stream * stm, cubeb_stream_params * stream_params, char const * stream_name)
+{
+  assert(stm || stream_params);
```
kinetiknz: Should be &&

src/cubeb_pulse.c
```diff
@@ -485,91 +636,153 @@ pulse_destroy(cubeb * ctx)
...
+  pa_sample_spec ss;
+  ss.format = cube_to_pulse_format(stream_params->format);
```
kinetiknz: If cubeb_to_pulse_format returns PA_SAMPLE_INVALID, we still try to create a stream.  We also lose the old CUBEB_ERROR_INVALID_FORMAT result from pulse_stream_init which could be used to try fallback formats.

src/cubeb_pulse.c
```diff
@@ -485,91 +636,153 @@ pulse_destroy(cubeb * ctx)
...
+                                     PA_STREAM_START_CORKED | PA_STREAM_ADJUST_LATENCY,
+                                     NULL, NULL);
   }
-  WRAP(pa_stream_set_state_callback)(stm->stream, stream_state_callback, stm);
-  WRAP(pa_stream_set_write_callback)(stm->stream, stream_request_callback, stm);
-  WRAP(pa_stream_connect_playback)(stm->stream, NULL, &battr,
+
+  // Set up input stream
+  if (input_stream_params) {
+    stm->input_stream = create_pa_stream(stm, input_stream_params, "cubeb input");
```
kinetiknz: This should use the same stream name as the output so they're paired correctly.  If we can't use an identical string, appending "(input)" or something to stream_name would be better than using a generic name.

src/cubeb_pulse.c
```diff
@@ -485,91 +636,153 @@ pulse_destroy(cubeb * ctx)
...
+    WRAP(pa_stream_set_read_callback)(stm->input_stream, stream_read_callback, stm);
+
+    battr = set_buffering_attribute(latency, &stm->input_sample_spec);
+    WRAP(pa_stream_connect_record)(stm->input_stream,
+                                   input_stream_params->devid,
+                                   &battr,
                                    PA_STREAM_AUTO_TIMING_UPDATE | PA_STREAM_INTERPOLATE_TIMING |
-                                   PA_STREAM_START_CORKED,
-                                   NULL, NULL);
+                                   PA_STREAM_NOT_MONOTONIC |
```
kinetiknz: NOT_MONOTONIC is only documented to affect the result of pa_stream_get_time, which I can't see ever being called on the input_stream.  Why is the flag being used here?

include/cubeb/cubeb.h
```diff
@@ -127,6 +129,7 @@ typedef struct {
 #if defined(__ANDROID__)
   cubeb_stream_type stream_type; /**< Used to map Android audio stream types */
 #endif
+  cubeb_devid devid;
```
kinetiknz: At the very least: this needs to be documented.

I don't like adding this to stream_params, because it's pretty dangerous if the caller forgets to initialize it (which we've already seen happen during integration with Gecko), and we have no way to sanity check this value unlike the other values in this struct.

I think the original plan was to pass a devid in as a new parameter to cubeb_stream_params.  I guess one each for the input and output stream.  At least it's impossible to forget to pass a value in that case, and we're already breaking the API by adding the input stream parameters.

kinetiknz:
Thanks for working on this Alex.

I think for merging the final changes I'd prefer to see the history squashed so each commit contains a single logic change, etc. rather than pulling in the development history.

src/cubeb_pulse.c
achronop: I am afraid the story is more complicated here. The bug experienced once by jesup while debugging with rr. The problem was that write stream was in state PA_STREAM_CREATING (so I guess the error was PA_ERR_BADSTATE) which should not be possible because we wait until stream is in PA_STREAM_READY state inside stream_init. Unfortunately I am not able to reproduce it here in order to track it and I prefer to check and report than failing silently.

src/cubeb_pulse.c
achronop: I thought it's cheap enough to call it again. I see it is confusing though and I will change it to call it just once.

src/cubeb_pulse.c
achronop: This is another API change for cubeb. Since we have also to deal with devid issue we can add there this change.

src/cubeb_pulse.c
achronop: `read_frames` take part to additional computation inside if statement. Instead of changing that i created a method to calculate the size that does not produce overflow. I found it easier to consume all the data in this way. Please take a look and if you think that's not the right approach let me know.

src/cubeb_pulse.c
achronop: I updated the code to bring back this functionality.

src/cubeb_pulse.c
achronop: I use the same name now and seems to work. The documentation also does not mention anything about unique name.

achronop:
Thank you very much for your time!

I have updated and/or comment back most of the comments. I've left behind those that will change due to interface changes. They will be updated by rebase. Unfortunately check fails now.

I've also left squash for then. When we have API changes landed I will fix this too. I know it looks ugly now, sorry for that.

In the mean time feel free to comment on anything of the above. Thanks again!

kinetiknz:
I addressed the my remaining comments, merged and pushed the squashed version here: https://github.com/kinetiknz/cubeb/commit/79ff75467fd439b435843d98da0d124babc6233b

PR #65, opened 2016-01-20
Signature changes for stream init and data callback

achronop:
Change on the signature of cubeb_stream_init to add the device input/output that will be connected. If NULL provided the default device should be chosen. 

Change on the signature for data_callback to make the input buffer const since it is not safe for some implementations to treat the buffer as non-const.


kinetiknz:
What's the expected API behavior if the requested device isn't available? I think, because we have device enumeration, the call needs to fail so that the caller can try a different device.

So for every backend changed, until it supports device selection, it should return an error if the device ID is not NULL.  We probably need to add a new error code for this.

kinetiknz:
I merged this and made all of the changes.  Thanks.

PR #66, opened 2016-01-21
API change for device notifications

achronop:
This pull request contains the API changes for device notification.

Currently method cubeb_register_device_collection_changed() does not has a parameter to indicate whether to notify for collection changes on sinks, sources or both. This request introduces a new method parameter of type cubeb_device_type which allow the user to choose that:

CUBEB_DEVICE_TYPE_INPUT --> for sources
CUBEB_DEVICE_TYPE_OUTPUT --> for sinks
CUBEB_DEVICE_TYPE_INPUT | CUBEB_DEVICE_TYPE_OUTPUT --> for both sinks and sources

This is a similar logic that applies in the cubeb_enumerate_devices() method.
Thanks!

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -480,12 +480,14 @@ int cubeb_device_info_destroy(cubeb_device_info * info);
 /** Registers a callback which is called when the system detects
     a new device or a device is removed.
     @param context
+    @param devtype device type to include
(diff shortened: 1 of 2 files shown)


kinetiknz:
Merged in 67b21f4451481830246c14abb64fe728009f647e

Thanks!

PR #67, opened 2016-01-22
Device notification for PulseAudio

jesup:
I'm probably doing this wrong....

### DIFF
--- a/src/cubeb.c
+++ b/src/cubeb.c
@@ -439,8 +439,13 @@ int cubeb_register_device_collection_changed(cubeb * context,
                                              cubeb_device_collection_changed_callback callback,
                                              void * user_ptr)
 {
-  if ((devtype & (CUBEB_DEVICE_TYPE_INPUT | CUBEB_DEVICE_TYPE_OUTPUT)) == 0)
+  if (context == NULL || (devtype & (CUBEB_DEVICE_TYPE_INPUT | CUBEB_DEVICE_TYPE_OUTPUT)) == 0)
     return CUBEB_ERROR_INVALID_PARAMETER;
-  return CUBEB_ERROR_NOT_SUPPORTED;
+
+  if (!context->ops->register_device_collection_changed) {
+    return CUBEB_ERROR_NOT_SUPPORTED;
+  }
+
+  return context->ops->register_device_collection_changed(context, devtype, callback, user_ptr);
 }
 
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -76,6 +76,9 @@
   X(pa_stream_peek)                             \
   X(pa_stream_drop)                             \
   X(pa_stream_get_buffer_attr)                  \
+  X(pa_stream_get_device_name)                  \
+  X(pa_context_set_subscribe_callback)          \
+  X(pa_context_subscribe)                       \
 
 #define MAKE_TYPEDEF(x) static typeof(x) * cubeb_##x;
 LIBPULSE_API_VISIT(MAKE_TYPEDEF);
@@ -101,6 +104,8 @@ struct cubeb {
   pa_sink_info * default_sink_info;
   char * context_name;
   int error;
+  cubeb_device_collection_changed_callback collection_changed_callback;
+  void * collection_changed_user_ptr;
 };
 
 struct cubeb_stream {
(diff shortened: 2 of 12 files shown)


src/cubeb.c
```diff
@@ -435,9 +435,18 @@ int cubeb_device_info_destroy(cubeb_device_info * info)
 }
 
 int cubeb_register_device_collection_changed(cubeb * context,
+                                             cubeb_device_type devtype,
                                              cubeb_device_collection_changed_callback callback,
                                              void * user_ptr)
 {
-  return CUBEB_ERROR_NOT_SUPPORTED;
+  if (!context || !callback) {
```
kinetiknz: The intention is for a NULL callback to unregister an existing callback.

This applies to cubeb_stream_register_device_changed_callback too, which also currently rejects it with INVALID_PARAMETER.

src/cubeb_pulse.c
```diff
@@ -101,6 +104,8 @@ struct cubeb {
   pa_sink_info * default_sink_info;
   char * context_name;
   int error;
+  cubeb_device_collection_changed_callback collection_changed_callback;
+  void * user_ptr;
```
kinetiknz: Call this collection_changed_user_ptr or something

src/cubeb_pulse.c
```diff
@@ -1174,6 +1179,100 @@ pulse_enumerate_devices(cubeb * context, cubeb_device_type type,
...
+int pulse_stream_get_current_device(cubeb_stream * stm, cubeb_device ** const device)
+{
+#if PA_CHECK_VERSION(0, 9, 8)
+  *device = calloc(1, sizeof(cubeb_device));
+
+  if (*device == NULL)
+    return CUBEB_ERROR;
+
+  if (stm->input_stream)
+    (*device)->input_name = strdup(WRAP(pa_stream_get_device_name)(stm->input_stream));
```
kinetiknz: pa_stream_get_device_name returns NULL on error and it's not valid to call strdup(NULL), so need to check the result first.

src/cubeb_pulse.c
```diff
@@ -1174,6 +1179,100 @@ pulse_enumerate_devices(cubeb * context, cubeb_device_type type,
...
+  return CUBEB_OK;
+#else
+  return CUBEB_ERROR_NOT_SUPPORTED;
+#endif
+}
+
+int pulse_stream_device_destroy(cubeb_stream * stream,
+                                cubeb_device * device)
+{
+  if (device->input_name) {
```
kinetiknz: free(NULL) is fine so there's no need to test the pointer first.  And since we free the containing struct, there's no need to NULL the members after calling free().

src/cubeb_pulse.c
```diff
@@ -1174,6 +1179,100 @@ pulse_enumerate_devices(cubeb * context, cubeb_device_type type,
...
+
+  WRAP(pa_context_set_subscribe_callback)(context->context, pulse_subscribe_callback, context);
+
+  // Set subscription mask according to dev type
+  pa_subscription_mask_t mask;
+  if (devtype == CUBEB_DEVICE_TYPE_INPUT)
+    mask = PA_SUBSCRIPTION_MASK_SOURCE;
+  else if (devtype == CUBEB_DEVICE_TYPE_OUTPUT)
+    mask = PA_SUBSCRIPTION_MASK_SINK;
+  else
```
kinetiknz: So passing CUBEB_DEVICE_TYPE_UNKNOWN maps to INPUT | OUTPUT? That doesn't seem right.

src/cubeb_pulse.c
```diff
@@ -1174,6 +1179,100 @@ pulse_enumerate_devices(cubeb * context, cubeb_device_type type,
...
+    mask = PA_SUBSCRIPTION_MASK_SINK | PA_SUBSCRIPTION_MASK_SOURCE;
+
+  pa_operation *o = NULL;
+  o = WRAP(pa_context_subscribe)(context->context, mask, NULL, NULL/*user_ptr*/);
```
kinetiknz: user_ptr comment seems unnecessary, you're not passing a success_cb to pa_context_subscribe

src/cubeb_pulse.c
```diff
@@ -1174,6 +1179,100 @@ pulse_enumerate_devices(cubeb * context, cubeb_device_type type,
...
+  // Set subscription mask according to dev type
+  pa_subscription_mask_t mask;
+  if (devtype == CUBEB_DEVICE_TYPE_INPUT)
+    mask = PA_SUBSCRIPTION_MASK_SOURCE;
+  else if (devtype == CUBEB_DEVICE_TYPE_OUTPUT)
+    mask = PA_SUBSCRIPTION_MASK_SINK;
+  else
+    mask = PA_SUBSCRIPTION_MASK_SINK | PA_SUBSCRIPTION_MASK_SOURCE;
+
+  pa_operation *o = NULL;
```
kinetiknz: Space

src/cubeb_pulse.c
```diff
@@ -1174,6 +1179,100 @@ pulse_enumerate_devices(cubeb * context, cubeb_device_type type,
...
+    device->output_name = NULL;
+  }
+
+  free(device);
+  return CUBEB_OK;
+}
+
+void pulse_subscribe_callback(pa_context *ctx,
+                              pa_subscription_event_type_t t,
+                              uint32_t index, void *userdata)
```
kinetiknz: Spaces between * and param names

src/cubeb_pulse.c
```diff
@@ -1174,6 +1179,100 @@ pulse_enumerate_devices(cubeb * context, cubeb_device_type type,
...
+{
+  cubeb * context = userdata;
+
+  switch (t & PA_SUBSCRIPTION_EVENT_FACILITY_MASK) {
+  case PA_SUBSCRIPTION_EVENT_SINK:
+    if ((t & PA_SUBSCRIPTION_EVENT_TYPE_MASK) == PA_SUBSCRIPTION_EVENT_REMOVE)
+      printf("Removing sink index %d\n", index);
```
kinetiknz: Change to LOG() or remove

src/cubeb_pulse.c
```diff
@@ -1174,6 +1179,100 @@ pulse_enumerate_devices(cubeb * context, cubeb_device_type type,
...
+    context->collection_changed_callback(context, context->user_ptr);
+    break;
+
+  case PA_SUBSCRIPTION_EVENT_SOURCE:
+    if ((t & PA_SUBSCRIPTION_EVENT_TYPE_MASK) == PA_SUBSCRIPTION_EVENT_REMOVE)
+      printf("Removing source index %d\n", index);
+    else if ((t & PA_SUBSCRIPTION_EVENT_TYPE_MASK) == PA_SUBSCRIPTION_EVENT_NEW ){
+      printf("Adding source index %d\n", index);
+    }
+    context->collection_changed_callback(context, context->user_ptr);
```
kinetiknz: At/after this point, the caller is expected to cubeb_device_collection_destroy and call cubeb_enumerate_devices to get the updated device list, right? Just want to make sure I understand the intent.

kinetiknz:
Need rebasing against current master, plus comments addressed.

achronop:
I am afraid this code not exactly there for landing. I have made some modifications from that point.

kinetiknz:
No problem, jesup asked me to take a look and give some feedback, hopefully some of which is useful.

src/cubeb_pulse.c
achronop: To my understanding the callback notify the fact that the collection is invalidated and it's up to the user to create a new collection inside the callback. I have this impression because the signature of the callback does not offer a collection out argument. Actually I made a similar question to padenot on irc and he also pointed there. But it makes sense to discuss it more.

Not directly relative to this comment but worth to mention the fact that enumerate_devices cannot be called directly from the callback but need to be set in a separate thread in order to work. This is because callback is executed in the pulse thread and the implementation of the method contains `wait_operation`s which will freeze the pulse thread. It's kind of odd since it would be the most expected thing to do someone inside that callback (given my understanding of use).

src/cubeb_pulse.c
achronop: You are right. I see one of your latest commits protects that case and now we can have only INPUT | OUTPUT in here. Do you still want me to add extra safeguards here?

achronop:
Thank you for taking the time to provide initial comments they helped a lot. I pushed a better version of the code. Take a look when possible and let me know what you think.

kinetiknz:
Thanks!

PR #70, opened 2016-01-27
pulse: Drop now-redundant assert in stream read callback

ford-prefect:
Following to @jesup's last PR


kinetiknz:
Good catch, thanks!

PR #71, opened 2016-01-27
Pulse: Add missing synchronizations

achronop:
Add missing mutex protection to methods that communicate with PulseAudio thread. The methods are 
- pulse_enumerate_devices
- pulse_register_device_collection_changed

The problem reproduced by firing a number of threads that call the methods.


kinetiknz:
Thanks!

PR #73, opened 2016-02-06
Refactor the resampler code to a new, duplex, architecture

padenot:
This is only the resampling code, not the WASAPI code. I feel I need to work a bit more on the windows-specific part before sending the PR, but it's working just fine on my windows box as it is, just not very beautiful.

This is supposed to be very generic, we'll be able to use it for OpenSL probably, maybe other.

This PR does not include [controlled adaptive resampling](http://lac.linuxaudio.org/2012/download/adaptive-resampling-slides.pdf) for now (for example to compensate for different audio interface drift because their clock is not synced), because we're only going to use it for Windows, and, from my (limited) testing, it does some sort of sync internally (or maybe I was lucky and my interfaces don't drift ?) However, there is room to add that in the future if needed.

I've also added a number of tests, and enabled more tests, but I think I was a bit over-optimistic about Travis' capabilities regarding audio things, so it probably makes sense to disable them. Also, some tests are very thorough (see the nested for-loops in `test_resamplers.cpp`), and so very very long to run, I'll check how long it takes on Travis and treeherder and probably will end up making a "light" version of them covering the basics.


Makefile.am
```diff
@@ -1,7 +1,7 @@
 AUTOMAKE_OPTIONS = foreign 1.10 dist-bzip2 subdir-objects
 ACLOCAL_AMFLAGS = -I m4
 
-AM_CPPFLAGS = -I$(top_srcdir)/include -I$(top_builddir)/include -I. -I$(top_srcdir)/src
+AM_CPPFLAGS = -I$(top_srcdir)/include -I$(top_builddir)/include -I. -I$(top_srcdir)/src -std=c++11
```
kinetiknz: AM_CPPFLAGS is C pre-processor flags.  Add AM_CXXFLAGS for this.

Makefile.am
```diff
@@ -97,7 +101,18 @@ test_test_latency_LDADD = -lm src/libcubeb.la $(platform_lib)
...
+test_test_resampler_SOURCES = test/test_resampler.cpp
+test_test_resampler_LDADD = -lm src/libcubeb.la $(platform_lib) src/cubeb_resampler.o
+
+test_test_duplex_SOURCES = test/test_duplex.cpp
+test_test_duplex_LDADD = -lm src/libcubeb.la $(platform_lib)
+
+test_test_record_SOURCES = test/test_record.cpp
+test_test_record_LDADD = -lm src/libcubeb.la $(platform_lib)
+
+test_test_resampler_SOURCES = test/test_utils.cpp
```
kinetiknz: test_test_utils_SOURCES

configure.ac
```diff
@@ -46,7 +46,7 @@ AM_PROG_CC_C_O
 AC_LIBTOOL_WIN32_DLL
 AM_PROG_LIBTOOL
 
-NEED_SPEEX=0
+NEED_SPEEX=1
```
kinetiknz: If we're going to require the libspeex resampler to build at all (which it looks like we are, since the default has changed and no backends are opting out), remove all of the NEED_SPEEX stuff from the build system.

src/cubeb_resampler.cpp
```diff
@@ -69,184 +32,280 @@ to_speex_quality(cubeb_resampler_quality q)
...
-                 void * ptr)
-    : stream(s)
-    , data_callback(cb)
-    , user_ptr(ptr)
-  {
-  }
+template<typename T>
+cubeb_resampler_speex_one_way<T>::cubeb_resampler_speex_one_way(int32_t channels,
+                                                                int32_t source_rate,
+                                                                int32_t target_rate,
```
kinetiknz: These might as well all be uint32_t, since that's what speex_resampler_init takes and negative values are non-sensical.

kinetiknz:
This PR doesn't compile and pass tests on Travis or locally.  Fixing the issue Travis reports, the AM_CXXFLAGS issue, and a compile error in test_record.cpp seems to be sufficient, but make check results in test_devices and test_duplex aborting and test_resampler segfaulting.  These need to be addressed before merging, along with any specific comments on the code.

test/test_record.cpp
```diff
@@ -0,0 +1,129 @@
...
+  if (r != CUBEB_OK) {
+	  fprintf(stderr, "Error getting the device collection\n");
+	  return r;
+  }
+
+  params.format = STREAM_FORMAT;
+  params.rate = SAMPLE_FREQUENCY;
+  params.channels = 1;
+  assert(collection->count >= 3);
+  params.devid = collection->device[5]->devid;
```
kinetiknz: devid was removed from cubeb_stream_params.

There's no guarantee that device[5] exists.  There's also no guarantee count will be >= 3. The test needs to be able to run (at least, not assert) even if the required resources aren't available.

test/test_duplex.cpp
```diff
@@ -0,0 +1,137 @@
...
+#ifdef NDEBUG
+#undef NDEBUG
+#endif
+#define _XOPEN_SOURCE 500
+#include <stdio.h>
+#include <stdlib.h>
+#include <math.h>
+#include <assert.h>
+
+#include "../include/cubeb/cubeb.h"
```
kinetiknz: "cubeb/cubeb.h" should be sufficient here and in the other tests.

src/cubeb_utils.h
```diff
@@ -0,0 +1,199 @@
...
+  /** The underlying storage */
+  T * data_;
+  /** The size, in number of elements, of the storage. */
+  size_t capacity_;
+  /** The number of elements the array contains. */
+  size_t length_;
+};
+
+template<typename T>
+class auto_ptr
```
kinetiknz: Is there a reason not to use std::unique_ptr instead?

.gitignore
```diff
@@ -47,8 +51,13 @@ test/test_tone
 test/test_tone.exe
 test/test_devices
 test/test_devices.exe
+test/test_resampler
+test/test_resampler.exe
+test/test_record
+test/test_record.exe
+test/test_utils
+test/test_utils.exe
```
kinetiknz: Add test_duplex{,.exe} too.

src/cubeb_resampler.cpp
```diff
@@ -69,184 +32,280 @@ to_speex_quality(cubeb_resampler_quality q)
...
-  virtual long fill(void * input_buffer, void * output_buffer, long frames_needed)
-  {
-    long got = data_callback(stream, user_ptr, input_buffer, output_buffer, frames_needed);
-    assert(got <= frames_needed);
-    return got;
+long noop_resampler::fill(void * input_buffer, long * input_frames_count,
+                          void * output_buffer, long output_frames)
+{
+  assert(input_buffer && output_buffer &&
+         *input_frames_count >= output_frames||
```
kinetiknz: Space between output_frames and ||

src/cubeb_utils.h
```diff
@@ -0,0 +1,199 @@
...
+    length_ = 0;
+  }
+
+   /** Change the storage of this auto array, copying the elements to the new
+    * storage.
+    * @returns true in case of success
+    * @returns false if the new capacity is not big enough to accomodate for the
+    *                elements in the array.
+    */
+  bool resize(size_t new_capacity)
```
kinetiknz: I think this would be better named reserve since it matches the behaviour of reserve (increase capacity but not length, i.e. doesn't add new items; never shrinks buffer) rather than resize from other vector-like classes I've seen.

src/cubeb_utils.h
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+   /** Append `length` elements to the end of the array, resizing the array if
+    * needed.
+    * @parameter elements the elements to append to the array.
+    * @parameter length the number of elements to append to the array.
+    */
+  void push(const T * elements, size_t length)
+  {
+    if (length_ + length > capacity_) {
```
kinetiknz: This check (and the same one in the other push() overload) seems redundant since resize(length_ + length) will do the expected thing anyway.

src/cubeb_utils.h
```diff
@@ -0,0 +1,199 @@
...
+    }
+    PodCopy(data_ + length_, elements, length);
+    length_ += length;
+  }
+
+  /** Append `length` zero-ed elements to the end of the array, resizing the
+   * array if needed.
+   * @parameter length the number of elements to append to the array.
+   */
+  void push(size_t length)
```
kinetiknz: This could probably be renamed resize(), although then it might need to allow shrinking the array too.

test/test_duplex.cpp
```diff
@@ -0,0 +1,137 @@
...
+struct user_state
+{
+  bool seen_noise;
+};
+
+long data_cb(cubeb_stream *stream, void *user, const void * inputbuffer, void *outputbuffer, long nframes)
+{
+  user_state * u = reinterpret_cast<user_state*>(user);
+  float *ib = (float *)inputbuffer;
+  float *ob = (float *)outputbuffer;
```
kinetiknz: These aren't floats on non-Win32.

test/test_duplex.cpp
```diff
@@ -0,0 +1,137 @@
...
+  if (r != CUBEB_OK) {
+    fprintf(stderr, "Error initializing cubeb library\n");
+    return r;
+  }
+
+  /* typical user-case: mono input, stereo output, low latency. */
+  input_params.format = CUBEB_SAMPLE_FLOAT32LE;
+  input_params.rate = 48000;
+  input_params.channels = 1;
+  output_params.format = CUBEB_SAMPLE_FLOAT32LE;
```
kinetiknz: This should either use the SAMPLE_FORMAT define from the top of the file, or be prepared to fail gracefully (i.e. not fail the test) if F32 isn't available for a given backend.

src/cubeb_resampler.cpp
```diff
@@ -69,184 +32,280 @@ to_speex_quality(cubeb_resampler_quality q)
...
-    return CUBEB_ERROR;
-  }
+  assert(!input_buffer && !input_frames_count &&
+         output_buffer && output_frames_needed);
 
-  uint32_t in_frames = leftover_frame_count + got;
-  uint32_t out_frames = frames_needed;
-  uint32_t old_in_frames = in_frames;
+  long got = 0;
+T * out_unprocessed = nullptr;
```
kinetiknz: Missing indent.

src/cubeb_resampler.cpp
```diff
@@ -69,184 +32,280 @@ to_speex_quality(cubeb_resampler_quality q)
...
-                                            out_buffer, &out_frames);
+  uint32_t output_frames_before_processing =
+    output_processor->input_needed_for_output(output_frames_needed);
+
+  /* fill directly the input buffer of the output processor to save a copy */
+  out_unprocessed =
+    output_processor->input_buffer(output_frames_before_processing);
+
+  got = data_callback(stream, user_ptr,
+    nullptr, nullptr,
```
kinetiknz: Indent to the (

src/cubeb_resampler.h
```diff
@@ -28,18 +28,15 @@ typedef enum {
...
- *                           the size of buffer allocated for resampling.
  * @param user_ptr User data supplied to the data callback.
  * @param quality Quality of the resampler.
  * @retval A non-null pointer if success.
  */
 cubeb_resampler * cubeb_resampler_create(cubeb_stream * stream,
-                                         cubeb_stream_params params,
+                                         cubeb_stream_params * input_params,
+                                         cubeb_stream_params * output_params,
                                          unsigned int out_rate,
```
kinetiknz: target_rate?

src/cubeb_resampler_internal.h
```diff
@@ -0,0 +1,488 @@
...
+    uint32_t out_len = output_frame_count;
+
+    speex_resample(resampling_in_buffer.data(), &in_len,
+                   output_buffer, &out_len);
+
+    /* This shifts back any unresampled samples to the beginning of the input
+       buffer. */
+    resampling_in_buffer.pop(nullptr, frames_to_samples(in_len));
+
+    // assert(resampling_in_buffer.length() == 0);
```
kinetiknz: Is this commented out because speex doesn't always consume the requested number of samples?  Since drain() is documented to empty the input buffer, need to handle this or change the documentation.

src/cubeb_resampler_internal.h
```diff
@@ -0,0 +1,488 @@
...
+      resampling_out_buffer.resize(frames_to_samples(output_frame_count));
+    }
+
+    uint32_t in_len = samples_to_frames(resampling_in_buffer.length());
+    uint32_t out_len = output_frame_count;
+
+    speex_resample(resampling_in_buffer.data(), &in_len,
+                   resampling_out_buffer.data(), &out_len);
+
+    assert(out_len == output_frame_count );
```
kinetiknz: Trailing space before )

src/cubeb_resampler_internal.h
```diff
@@ -0,0 +1,488 @@
...
+    , length(frames)
+  {
+    /* Fill the delay line with some silent frames to add latency. */
+    delay_input_buffer.push(frames * channels);
+  }
+  /* Add some latency to the delay line.
+   * @param frames the number of frames of latency to add. */
+  void add_latency(size_t frames)
+  {
+    length+=frames;
```
kinetiknz: Space around +=

src/cubeb_resampler_internal.h
```diff
@@ -0,0 +1,488 @@
...
+   * @parameter frames_needed the number of frame that will be written. */
+  void output(T * buffer, uint32_t frame_needed)
+  {
+    delay_input_buffer.pop(buffer, frames_to_samples(frame_needed));
+  }
+  /** Pop some frames from the internal buffer, into a internal output buffer.
+   * @parameter frames_needed the number of frames to be returned.
+   * @return a buffer containing the delayed frames. The consumer should not
+   * hold onto the pointer. */
+  T* output(uint32_t frames_needed)
```
kinetiknz: Space between T and *

src/cubeb_resampler.cpp
```diff
@@ -69,184 +32,280 @@ to_speex_quality(cubeb_resampler_quality q)
...
+  T * out_unprocessed = nullptr;
+  size_t output_frames_before_processing;
+  /* The number of frames returned from the callback. */
+  long got = 0;
+
+  /* We need to determine how much frames to present to the consumer.
+   * - If we have a two way stream, but we're only resampling input, we resample
+   * the input to the number of output frames.
+   * - If we have a two way stream, but we're only resampling the output, we
+   * resize the input buffer of the outut resampler to the number of input
```
kinetiknz: Typo: output

src/cubeb_resampler_internal.h
```diff
@@ -0,0 +1,488 @@
...
+  /** Returns the number of frames one needs to input into the delay line to get
+   * #frames_needed frames back.
+   * @parameter frames_needed the number of frames one want to write into the
+   * delay_line
+   * @returns the number of frames one will get. */
+  size_t input_needed_for_output(uint32_t frames_needed)
+  {
+    return frames_needed;
+  }
+  size_t output_for_input(uint32_t input_frames) {
```
kinetiknz: New line between ) and {

src/cubeb_resampler_internal.h
```diff
@@ -0,0 +1,488 @@
...
+    }
+  }
+
+  if (input_params && (input_params->rate != target_rate)) {
+    input_resampler = new cubeb_resampler_speex_one_way<T>(input_params->channels,
+                                                           input_params->rate,
+                                                           target_rate,
+                                                           to_speex_quality(quality));
+    if (!input_resampler) {
+      return NULL;
```
kinetiknz: output_resampler could leak here

src/cubeb_resampler_internal.h
```diff
@@ -0,0 +1,488 @@
...
+    }
+  }
+
+  /* If we resample only one direction but we have a duplex stream, insert a
+   * delay line with a length equal to the resampler latency of the
+   * other direction so that the streams are synchronized. */
+  if (input_resampler && !output_resampler && input_params && output_params) {
+    output_delay = new delay_line<T>(input_resampler->latency(),
```
kinetiknz: input_delay and output_delay aren't checked to see if new succeeded.

src/cubeb_resampler.cpp
```diff
@@ -69,184 +32,280 @@ to_speex_quality(cubeb_resampler_quality q)
...
-                                             void * ptr)
-  : speex_resampler(r)
+  return data_callback(stream, user_ptr,
+                       input_buffer, output_buffer, output_frames);
+}
+
+namespace {
+
+long
+frame_count_at_rate(long frame_count, float rate)
```
kinetiknz: This doesn't seem to be used anywhere.

src/cubeb_resampler_internal.h
```diff
@@ -0,0 +1,488 @@
...
+    resampling_in_buffer.push(frames_to_samples(frame_count));
+    return resampling_in_buffer.data() + prev_length;
+  }
+private:
+  /** Wrapper for the speex resampling functions to have a typed
+    * interface. */
+  void speex_resample(float * input_buffer, uint32_t * input_frame_count,
+                      float * output_buffer, uint32_t * output_frame_count)
+  {
+    speex_resampler_process_interleaved_float(speex_resampler,
```
kinetiknz: It's probably worth asserting that these return RESAMPLER_ERR_SUCCESS.

src/cubeb_resampler_internal.h
```diff
@@ -0,0 +1,488 @@
...
+    assert(out_len == output_frame_count );
+
+    /* This shifts back any unresampled samples to the beginning of the input
+       buffer. */
+    resampling_in_buffer.pop(nullptr, frames_to_samples(in_len));
+
+    return resampling_out_buffer.data();
+  }
+
+  /** Get the l atency of the resampler, in output frames. */
```
kinetiknz: Extra space in "latency"

src/cubeb_resampler_internal.h
```diff
@@ -0,0 +1,488 @@
...
+  uint32_t latency() const
+  {
+    /* The documentation of the resampler talks about "samples" here, but it
+     * only consider a single channel here so it's the same number of frames. */
+    return speex_resampler_get_output_latency(speex_resampler) + additional_latency;
```
kinetiknz: Maybe assert the result of get_output_latency is >= 0 since it returns an int.

src/cubeb_resampler_internal.h
```diff
@@ -0,0 +1,488 @@
...
+      (output_params && !input_params && (output_params->rate == target_rate))) {
+    return new noop_resampler(stream, callback, user_ptr);
+  }
+
+  /* Determine if we need to resampler one or both directions, and create the
+     resamplers. */
+  if (output_params && (output_params->rate != target_rate)) {
+    /* The target param here is the platforms stream's rate, since we're
+     * resampling from the user's chosen sample rate to the platforms's
+     * mandatory sample rate. */
```
kinetiknz: While this is true in practice with current users, I don't think any of the resampling code cares about this fact and it could be used with some other rate without any issues.  So maybe move the comment up to the caller where the comment applies or remove it altogether?

kinetiknz:
Thanks, generally looks really good, appreciate you adding the tests.  I think we're good to merge after the issues I've raised have been addressed, although it would be good to look over it all again as I'm just coming back up to speed after being out sick for a couple of days.

test/test_record.cpp
padenot: Oops, that's some test code I was using to target a specific device on my windows box.

src/cubeb_utils.h
padenot: Too much Gecko work, I forgot the STL even existed :-).

src/cubeb_utils.h
padenot: We don't want to new/delete if we have enough room. In general, we try to only grow the buffers. In our case, their size stabilizes after a couple callbacks.

src/cubeb_utils.h
padenot: We really want to convey the fact that we're pushing silent frames here, I've renamed it to make it obvious.

padenot: Yeah I reckon we're sometimes using it to allocate and change the size to directly access the buffer, resize is better.

padenot: Hrm in fact it really appends some frames, I'm not sure, and I'll keep it as is, with a new, more obvious, name.

padenot:
I think it will probably be good to have those fixes squashed before merging, but this makes it clear that I fixed everything. The behavioral changes don't have "Review comments" at the beginning of the message.

kinetiknz:
Thanks!  Super easy to review with the broken up commits.  Squashing sounds good before merge.

The changes build locally, but test_record and test_duplex are both aborting on my Linux/PA system, both hit assert(stream_state.seen_noise)... the tests should probably bail early if there's no capture device available, I suspect Travis will have the same issue.

The build failure on Travis should be a trivial fix, updating the args for test_resampler_one_way to be unsigned and also the sample_rate array.

padenot:
This now passes on Linnux (gcc + clang) after some serious head scratching. I'm working on getting mingw up and running now, of course it does not work, the linker is probably the wrong one or something. I know close to nothing about how mingw works so I need to scratch my head some more.

padenot:
>  Have you sent this upstream, and does the Gecko in-tree version need the same fix?

Not yet, I need to talk to Jean-Marc, and will send it upstream. Also, yes, I believe Gecko needs the same fix.

padenot:
Matthew, this is green, squashed, and address all review comments.

I've switched to a more recent distro on travis to get c++11, because that gives us gcc-4.8 and recent mingw to cross-compile. I believe Travis will update their image soon, so we'll be able to switch back to a non-beta `.travis.yml`. In addition, it pins the version of the compiler on Linux.

Also, as you suggested, we now skip the tests that need audio input when no available input device is present.

kinetiknz:
Thanks!  I appreciate you sorting the Travis stuff out.

PR #75, opened 2016-02-19
Implement input and duplex streams in the WASAPI backend.

padenot:
With the resampler bits landed, this is rather straightforward to implement.


kinetiknz:
Strange, I didn't get any notifications from GitHub that you had pushed updates.  Might be a good idea to post a comment in the PR in future as well.

Updates all look good, thanks for that.

PR #77, opened 2016-03-02
Unnecessary to check Android version if android api level > 10

makotokato:
Android/arm64 doesn't export __system_property_get on NDK.  So we shouldn't this API to get API level on 64bit Android, and, API level is already 10+, it is unnecessary to check whether JB MR1.


kinetiknz:
Thanks!

PR #78, opened 2016-03-04
Implement full-duplex for audiounit backend.

achronop:
The PR implements the full-duplex functionality for audiounit. The idea is to configure two different audiounit (AU) one for the input and one for the output. The input AU operates in device sampling rate thus the new full-duplex cubeb_resampler is used to modify the input when necessary. Output AU operates correctly in the user selected sampling rate.

In addtion AU operates on different threads. Thread sequence is not guaranteed thus a primitive ring array was implemented to hold the incoming buffers and feed them to the output in the right order. For this reason a new header file is added with the corresponding test file.

Last thing to note is that the newly added devid data member of cubeb_device_info struct set to free inside the general method cubeb_device_info_destroy(). This is based on the convention that the devid data will be reallocated before set in every implementation. The reason for this is that the destroy method is general (not available in back ends) so there is no other way to free the memory in the right place.

Thanks!

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -221,7 +221,7 @@ typedef enum {
  *  `cubeb_enumerate_devices`, and must be destroyed using
  *  `cubeb_device_info_destroy`. */
 typedef struct {
-  cubeb_devid devid;          /**< Device identifier handle. */
+  cubeb_devid devid;          /**< Device identifier handle. Always fresh allocate and copy it's value. */
   char * device_id;           /**< Device identifier which might be presented in a UI. */
   char * friendly_name;       /**< Friendly device name which might be presented in a UI. */
   char * group_id;            /**< Two devices have the same group identifier if they belong to the same physical device; for example a headset and microphone. */
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -221,7 +221,7 @@ typedef enum {
  *  `cubeb_enumerate_devices`, and must be destroyed using
  *  `cubeb_device_info_destroy`. */
 typedef struct {
-  cubeb_devid devid;          /**< Device identifier handle. Always fresh allocate and copy it's value. */
+  cubeb_devid devid;          /**< Device identifier handle. Required deep copy of the value. */
   char * device_id;           /**< Device identifier which might be presented in a UI. */
   char * friendly_name;       /**< Friendly device name which might be presented in a UI. */
   char * group_id;            /**< Two devices have the same group identifier if they belong to the same physical device; for example a headset and microphone. */
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -221,7 +221,7 @@ typedef enum {
  *  `cubeb_enumerate_devices`, and must be destroyed using
  *  `cubeb_device_info_destroy`. */
 typedef struct {
-  cubeb_devid devid;          /**< Device identifier handle. Required deep copy of the value. */
+  cubeb_devid devid;          /**< Device identifier handle. */
   char * device_id;           /**< Device identifier which might be presented in a UI. */
   char * friendly_name;       /**< Friendly device name which might be presented in a UI. */
   char * group_id;            /**< Two devices have the same group identifier if they belong to the same physical device; for example a headset and microphone. */
--- a/src/cubeb_audiounit.c
+++ b/src/cubeb_audiounit.c
@@ -61,16 +61,6 @@
 #define LOG(...)
 #endif
 
-#ifdef DEBUG
-#define ASSERT_LOCKED(mutex)           \
-do {                                   \
-  int rv = pthread_mutex_lock(&mutex); \
-  assert(rv == EDEADLK);               \
-} while (0);
-#else
-#define ASSERT_LOCKED(mutex)
-#endif
-
 static struct cubeb_ops const audiounit_ops;
 
 struct cubeb {
@@ -165,7 +155,6 @@ audiounit_input_callback(void * user_ptr,
   void * input_buffer = NULL;
 
   pthread_mutex_lock(&stm->mutex);
-  ASSERT_LOCKED(stm->mutex);
 
   assert(stm->input_unit != NULL);
   assert(AU_IN_BUS == bus);
@@ -255,7 +244,6 @@ audiounit_output_callback(void * user_ptr,
   void * output_buffer = NULL, * input_buffer = NULL;
 
   pthread_mutex_lock(&stm->mutex);
-  ASSERT_LOCKED(stm->mutex);
 
   if (stm->shutdown) {
     audiounit_make_silent(outBufferList);
@@ -941,11 +929,7 @@ audiounit_stream_init(cubeb * context,
 
   pthread_mutexattr_t attr;
   pthread_mutexattr_init(&attr);
-#ifdef DEBUG
-  pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_ERRORCHECK);
-#else
   pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
-#endif
   r = pthread_mutex_init(&stm->mutex, &attr);
   assert(r == 0);
   pthread_mutexattr_destroy(&attr);
--- a/src/cubeb_audiounit.c
+++ b/src/cubeb_audiounit.c
@@ -88,7 +88,6 @@ struct cubeb_stream {
   pthread_mutex_t mutex;
   /* Hold the input samples in every
    * input callback iteration */
-  AudioBuffer input_buffer[RING_ARRAY_CAPACITY];
   ring_array input_buffer_array;
   /* Frames on input buffer */
   uint32_t input_buffer_frames;
@@ -164,12 +163,12 @@ audiounit_input_callback(void * user_ptr,
   }
 
   /* Get next store buffer from ring array */
-  AudioBuffer * input_buffer = ring_array_get_next_free_buffer(&stm->input_buffer_array);
+  AudioBuffer * input_buffer = ring_array_get_free_buffer(&stm->input_buffer_array);
   if (input_buffer == NULL) {
     LOG("input: Ring array is full drop one buffer\n");
-    ring_array_get_first_data_buffer(&stm->input_buffer_array);
+    ring_array_get_data_buffer(&stm->input_buffer_array);
 
-    input_buffer = ring_array_get_next_free_buffer(&stm->input_buffer_array);
+    input_buffer = ring_array_get_free_buffer(&stm->input_buffer_array);
     assert(input_buffer);
   }
 
@@ -204,7 +203,7 @@ audiounit_input_callback(void * user_ptr,
   /* Input only. Call the user callback through resampler.
      Resampler will deliver input buffer in the correct rate. */
   frames = input_frames;
-  input_buffer = ring_array_get_first_data_buffer(&stm->input_buffer_array);
+  input_buffer = ring_array_get_data_buffer(&stm->input_buffer_array);
   assert(input_buffer && "fetch buffer is null in the input");
   outframes = cubeb_resampler_fill(stm->resampler,
                                    input_buffer->mData,
@@ -280,7 +279,7 @@ audiounit_output_callback(void * user_ptr,
       return noErr;
     }
     /* Input samples stored previously in input callback. */
-    input_aud_buf = ring_array_get_first_data_buffer(&stm->input_buffer_array);
+    input_aud_buf = ring_array_get_data_buffer(&stm->input_buffer_array);
     if (input_aud_buf == NULL) {
       LOG("Requested more output than input. "
              "This is either a hole or we are after a stream stop and input thread stopped before output\n");
@@ -291,7 +290,7 @@ audiounit_output_callback(void * user_ptr,
 
       /* Avoid here to allocate new memory since we are inside callback. Use the existing
        * allocated buffers since the ring array is empty and the buffer is not used. */
-      input_aud_buf = &stm->input_buffer[0];
+      input_aud_buf = ring_array_get_dummy_buffer(&stm->input_buffer_array);
       audiounit_make_silent(input_aud_buf);
     }
     input_buffer = input_aud_buf->mData;
@@ -810,59 +809,19 @@ audiounit_create_unit(AudioUnit * unit,
 }
 
 static int
-audiounit_init_audio_buffer(AudioBuffer * buffer,
-                            uint32_t bytesPerFrame,
-                            uint32_t channelsPerFrame,
-                            uint32_t frames)
-{
-  size_t size = bytesPerFrame * frames;
-
-  buffer->mData = calloc(1, size);
-  if (buffer->mData == NULL) {
-    return CUBEB_ERROR;
-  }
-
-  buffer->mNumberChannels = channelsPerFrame;
-  buffer->mDataByteSize = size;
-
-  return CUBEB_OK;
-}
-
-static int
 audiounit_init_input_buffer_array(cubeb_stream * stream)
 {
-  /* Init ring array. */
-  ring_array_init(&stream->input_buffer_array);
-  /* Initialize buffer list. */
-  memset(&stream->input_buffer, 0, sizeof(stream->input_buffer));
-  /* Create input descriptor. */
-  AudioStreamBasicDescription src_desc = stream->input_desc;
-  src_desc.mSampleRate = stream->input_hw_rate;
-  /* Set data to ring array. */
-  for (int i = 0; i < RING_ARRAY_CAPACITY; ++i) {
-    /* Allocate the data (AudioBufferList here). */
-    if (audiounit_init_audio_buffer(&stream->input_buffer[i],
-                                    src_desc.mBytesPerFrame,
-                                    src_desc.mChannelsPerFrame,
-                                    stream->input_buffer_frames) != CUBEB_OK) {
-      return CUBEB_ERROR;
-    }
-    /* Set the data in the array. */
-    ring_array_set_data(&stream->input_buffer_array,
-                        &stream->input_buffer[i], i);
-  }
-  return CUBEB_OK;
+  int r = ring_array_init(&stream->input_buffer_array,
+                          stream->input_desc.mBytesPerFrame,
+                          stream->input_desc.mChannelsPerFrame,
+                          stream->input_buffer_frames);
+
+  return r;
 }
 
 static void
 audiounit_destroy_input_buffer_array(cubeb_stream * stream)
 {
-  /* Destroy the data in the array first*/
-  for (int i = 0; i < RING_ARRAY_CAPACITY; ++i) {
-    if (stream->input_buffer[i].mData) {
-      free(stream->input_buffer[i].mData);
-    }
-  }
   ring_array_destroy(&stream->input_buffer_array);
 }
 
--- a/src/cubeb_audiounit.c
+++ b/src/cubeb_audiounit.c
@@ -160,32 +160,32 @@ audiounit_input_callback(void * user_ptr,
                          UInt32 input_frames,
                          AudioBufferList * bufs)
 {
-  cubeb_stream * stream = user_ptr;
+  cubeb_stream * stm = user_ptr;
   long outframes, frames;
   void * input_buffer = NULL;
 
-  pthread_mutex_lock(&stream->mutex);
-  ASSERT_LOCKED(stream->mutex);
+  pthread_mutex_lock(&stm->mutex);
+  ASSERT_LOCKED(stm->mutex);
 
-  assert(stream->input_unit != NULL);
+  assert(stm->input_unit != NULL);
   assert(AU_IN_BUS == bus);
 
-  if (stream->shutdown) {
-    pthread_mutex_unlock(&stream->mutex);
+  if (stm->shutdown) {
+    pthread_mutex_unlock(&stm->mutex);
     return noErr;
   }
 
   /* Get next store buffer from ring array */
-  AudioBufferList * store_input_buffer_list = ring_array_store_buffer(&stream->input_buffer_list_array);
+  AudioBufferList * store_input_buffer_list = ring_array_store_buffer(&stm->input_buffer_list_array);
   if (store_input_buffer_list == NULL) {
     LOG("input: Ring array is full drop one buffer\n");
-    ring_array_fetch_buffer(&stream->input_buffer_list_array);
+    ring_array_fetch_buffer(&stm->input_buffer_list_array);
 
-    store_input_buffer_list = ring_array_store_buffer(&stream->input_buffer_list_array);
+    store_input_buffer_list = ring_array_store_buffer(&stm->input_buffer_list_array);
     assert(store_input_buffer_list);
   }
   /* Render input samples */
-  OSStatus r = AudioUnitRender(stream->input_unit,
+  OSStatus r = AudioUnitRender(stm->input_unit,
                                flags,
                                tstamp,
                                bus,
@@ -196,41 +196,41 @@ audiounit_input_callback(void * user_ptr,
       store_input_buffer_list->mBuffers[0].mDataByteSize, store_input_buffer_list->mBuffers[0].mNumberChannels, input_frames);
 
   assert(input_frames > 0);
-  stream->frames_read += input_frames;
+  stm->frames_read += input_frames;
 
   // Full Duplex. We'll call data_callback in the AudioUnit output callback.
-  if (stream->output_unit != NULL) {
+  if (stm->output_unit != NULL) {
     // User callback will be called by output callback
-    pthread_mutex_unlock(&stream->mutex);
+    pthread_mutex_unlock(&stm->mutex);
     return noErr;
   }
 
   /* Input only. Call the user callback through resampler.
      Resampler will deliver input buffer in the correct rate. */
   frames = input_frames;
-  AudioBufferList * fetch_input_buffer_list = ring_array_fetch_buffer(&stream->input_buffer_list_array);
+  AudioBufferList * fetch_input_buffer_list = ring_array_fetch_buffer(&stm->input_buffer_list_array);
   assert(fetch_input_buffer_list && "fetch buffer is null in the input");
   input_buffer = fetch_input_buffer_list->mBuffers[0].mData;
-  outframes = cubeb_resampler_fill(stream->resampler,
+  outframes = cubeb_resampler_fill(stm->resampler,
                                    input_buffer,
                                    &frames,
                                    NULL,
                                    0);
 
   if (outframes < 0 || outframes != input_frames) {
-    stream->shutdown = 1;
-    pthread_mutex_unlock(&stream->mutex);
+    stm->shutdown = 1;
+    pthread_mutex_unlock(&stm->mutex);
     return noErr;
   }
 
-  pthread_mutex_unlock(&stream->mutex);
+  pthread_mutex_unlock(&stm->mutex);
   return noErr;
 }
 
 static void
 audiounit_make_silent(AudioBufferList * ioData)
 {
-  for(UInt32 i = 0; i<ioData->mNumberBuffers;i++) {
+  for(UInt32 i = 0; i < ioData->mNumberBuffers; i++) {
     memset(ioData->mBuffers[i].mData, 0, ioData->mBuffers[i].mDataByteSize);
   }
 }
@@ -250,29 +250,29 @@ audiounit_output_callback(void * user_ptr,
       outBufferList->mNumberBuffers, outBufferList->mBuffers[0].mDataByteSize,
       outBufferList->mBuffers[0].mNumberChannels, output_frames);
 
-  cubeb_stream * stream = user_ptr;
+  cubeb_stream * stm = user_ptr;
   long outframes = 0, input_frames = 0;
   void * output_buffer = NULL, * input_buffer = NULL;
 
-  pthread_mutex_lock(&stream->mutex);
-  ASSERT_LOCKED(stream->mutex);
+  pthread_mutex_lock(&stm->mutex);
+  ASSERT_LOCKED(stm->mutex);
 
-  if (stream->shutdown) {
+  if (stm->shutdown) {
     audiounit_make_silent(outBufferList);
-    pthread_mutex_unlock(&stream->mutex);
+    pthread_mutex_unlock(&stm->mutex);
     return noErr;
   }
 
-  stream->current_latency_frames = audiotimestamp_to_latency(tstamp, stream);
-  if (stream->draining) {
-    OSStatus r = AudioOutputUnitStop(stream->output_unit);
+  stm->current_latency_frames = audiotimestamp_to_latency(tstamp, stm);
+  if (stm->draining) {
+    OSStatus r = AudioOutputUnitStop(stm->output_unit);
     assert(r == 0);
-    if (stream->input_unit) {
-      r = AudioOutputUnitStop(stream->input_unit);
-      assert(r==0);
+    if (stm->input_unit) {
+      r = AudioOutputUnitStop(stm->input_unit);
+      assert(r == 0);
     }
-    stream->state_callback(stream, stream->user_ptr, CUBEB_STATE_DRAINED);
-    pthread_mutex_unlock(&stream->mutex);
+    stm->state_callback(stm, stm->user_ptr, CUBEB_STATE_DRAINED);
+    pthread_mutex_unlock(&stm->mutex);
     audiounit_make_silent(outBufferList);
     return noErr;
   }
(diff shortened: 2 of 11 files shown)


kinetiknz:
> Last thing to note is that the newly added devid data member of cubeb_device_info struct set to free inside the general method cubeb_device_info_destroy(). This is based on the convention that the devid data will be reallocated before set in every implementation. The reason for this is that the destroy method is general (not available in back ends) so there is no other way to free the memory in the right place.

It doesn't look like this change is safe to make without making further fixes, e.g.
wasapi: devid is the same as device_id, so this is now a double-free.
winmm: devid isn't a heap pointer.

The former could be fixed by a (variant of) strdup, but the latter can't be handled with this approach.  It looks like we'll need to move cubeb_device_info_destroy to be backend-specific.

include/cubeb/cubeb.h
```diff
@@ -221,7 +221,7 @@ typedef enum {
  *  `cubeb_enumerate_devices`, and must be destroyed using
  *  `cubeb_device_info_destroy`. */
 typedef struct {
-  cubeb_devid devid;          /**< Device identifier handle. */
+  cubeb_devid devid;          /**< Device identifier handle. Always fresh allocate and copy it's value. */
```
kinetiknz: Can you reword this? I'm not entirely sure what it means.

src/cubeb_audiounit.c
```diff
@@ -33,11 +34,42 @@
...
+#ifdef LOGGING_ENABLED
+#define LOG(...) do {                           \
+    fprintf(stderr, __VA_ARGS__);               \
+  } while(0)
+#else
+#define LOG(...)
+#endif
+
+#ifdef DEBUG
+#define ASSERT_LOCKED(mutex)           \
```
kinetiknz: I don't understand the purpose of this.  The only two places it's used are right after a call to pthread_mutex_lock - why not check the return value of that instead?

src/cubeb_audiounit.c
```diff
@@ -614,65 +914,176 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
   stm->state_callback = state_callback;
   stm->user_ptr = user_ptr;
   stm->device_changed_callback = NULL;
 
-  stm->sample_spec = ss;
-
   pthread_mutexattr_t attr;
   pthread_mutexattr_init(&attr);
+#ifdef DEBUG
+  pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_ERRORCHECK);
```
kinetiknz: I don't think it makes sense to change the type of mutex based on debug/release mode... use of a non/recursive mutex results in a different program design.

src/cubeb_audiounit.c
```diff
@@ -16,16 +16,17 @@
 #include <CoreAudio/AudioHardware.h>
 #include <CoreAudio/HostTime.h>
 #include <CoreFoundation/CoreFoundation.h>
-#else
+#endif
```
kinetiknz: Have all of the TARGET_OS_IPHONE changes been tested?

Same question for the MACOSX_LESS_THAN_106 changes.

src/cubeb_audiounit.c
```diff
@@ -538,73 +725,186 @@ audiounit_destroy(cubeb * ctx)
...
+                                 &devid, sizeof(AudioDeviceID));
+  if (err != noErr) {
+    return CUBEB_ERROR;
+  }
+
+  return CUBEB_OK;
+}
+
+static int
+audiounit_buflst_init_single_buffer (AudioBufferList * buflst,
```
kinetiknz: No space between buffer and (

src/cubeb_audiounit.c
```diff
@@ -538,73 +725,186 @@ audiounit_destroy(cubeb * ctx)
...
+                      void * user_ptr)
+{
+  cubeb_stream * stm;
+  AudioUnit input_unit;
+  AudioUnit output_unit;
+  int ret;
+  AURenderCallbackStruct aurcbs_in;
+  AURenderCallbackStruct aurcbs_out;
+  UInt32 size;
+#if 0
```
kinetiknz: Is there a reason this has been left in?

src/cubeb_audiounit.c
```diff
@@ -538,73 +725,186 @@ audiounit_destroy(cubeb * ctx)
...
+                      cubeb_stream_params * output_stream_params,
+                      unsigned int latency,
+                      cubeb_data_callback data_callback,
+                      cubeb_state_callback state_callback,
+                      void * user_ptr)
+{
+  cubeb_stream * stm;
+  AudioUnit input_unit;
+  AudioUnit output_unit;
+  int ret;
```
kinetiknz: r not ret for int return values (ret for OSStatus values)

src/cubeb_audiounit.c
```diff
@@ -694,42 +1105,67 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
+  cubeb_stream_params input_unconverted_params;
+  if (input_stream_params) {
+    input_unconverted_params = *input_stream_params;
+    /* Use the rate of the input device. */
+    input_unconverted_params.rate = stm->input_hw_rate;
+  }
+
+  /* Create resampler. Output params are unchanged
+   * because we do not need conversion on the output. */
+  stm->resampler = cubeb_resampler_create(stm,
```
kinetiknz: This is never freed anywhere.

src/cubeb_audiounit.c
```diff
@@ -694,42 +1105,67 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
+                                          input_stream_params ? &input_unconverted_params : NULL,
+                                          output_stream_params,
+                                          target_sample_rate,
+                                          stm->data_callback,
+                                          stm->user_ptr,
+                                          CUBEB_RESAMPLER_QUALITY_DESKTOP);
+  if (!stm->resampler) {
+    LOG("Could not get a resampler\n");
```
kinetiknz: The convention is to call audiounit_stream_destroy to clean up any already-allocated resources.

src/cubeb_audiounit.c
```diff
@@ -614,65 +914,176 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
+      audiounit_stream_destroy(stm);
+      return ret;
+    }
+
+    AudioStreamBasicDescription src_desc = stm->input_desc;
+    /* Input AudioUnit must be configured with device's sample rate. */
+    if (src_desc.mSampleRate != stm->input_hw_rate) {
+      /* Set the sample rate of the device we will resample inside input callback. */
+      src_desc.mSampleRate = stm->input_hw_rate;
+    }
```
kinetiknz: Why bother checking before assigning? The end result will be equal, so just set it unconditionally.

src/cubeb_audiounit.c
```diff
@@ -614,65 +914,176 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
-  r = OpenAComponent(comp, &stm->unit);
-#else
-  r = AudioComponentInstanceNew(comp, &stm->unit);
-#endif
-  if (r != 0) {
-    audiounit_stream_destroy(stm);
-    return CUBEB_ERROR;
+  /* Setup Input Stream! */
+  if (input_stream_params != NULL) {
+
```
kinetiknz: No newline

src/cubeb_audiounit.c
```diff
@@ -614,65 +914,176 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
+                             &size) != 0) {
+      audiounit_stream_destroy(stm);
+      return CUBEB_ERROR;
+    }
+
+    if (AudioUnitSetProperty(stm->input_unit,
+                             kAudioDevicePropertyBufferFrameSize,
+                             kAudioUnitScope_Output,
+                             AU_IN_BUS,
+                             &stm->input_buffer_frames,
```
kinetiknz: The previous AudioUnitGetProperty set stm->input_buffer_frames and this call clobbers the value without using it, so one of these must be wrong.

src/cubeb_audiounit.c
```diff
@@ -614,65 +914,176 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
-  input.inputProc = audiounit_output_callback;
-  input.inputProcRefCon = stm;
-  r = AudioUnitSetProperty(stm->unit, kAudioUnitProperty_SetRenderCallback,
-                           kAudioUnitScope_Global, 0, &input, sizeof(input));
-  if (r != 0) {
-    audiounit_stream_destroy(stm);
-    return CUBEB_ERROR;
+  /* Setup Output Stream! */
+  if (output_stream_params != NULL) {
+
```
kinetiknz: No newline

src/cubeb_audiounit.c
```diff
@@ -614,65 +914,176 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
+      /* Set the sample rate of the device we will resample inside input callback. */
+      src_desc.mSampleRate = stm->input_hw_rate;
+    }
+
+    if (AudioUnitSetProperty(stm->input_unit,
+                             kAudioUnitProperty_StreamFormat,
+                             kAudioUnitScope_Output,
+                             AU_IN_BUS,
+                             &src_desc,
+                             sizeof(AudioStreamBasicDescription)) !=0) {
```
kinetiknz: Space between != and 0

src/cubeb_audiounit.c
```diff
@@ -614,65 +914,176 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
   pthread_mutexattr_t attr;
   pthread_mutexattr_init(&attr);
+#ifdef DEBUG
+  pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_ERRORCHECK);
+#else
   pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
-  r = pthread_mutex_init(&stm->mutex, &attr);
+#endif
+  ret = pthread_mutex_init(&stm->mutex, &attr);
+  assert(0 == ret);
```
kinetiknz: r not ret for int return values
and r == 0, not reverse, please follow existing code styles rather than introducing new ones

src/cubeb_audiounit.c
```diff
@@ -614,65 +914,176 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
-  desc.componentManufacturer = kAudioUnitManufacturer_Apple;
-  desc.componentFlags = 0;
-  desc.componentFlagsMask = 0;
-#if MACOSX_LESS_THAN_106
-  comp = FindNextComponent(NULL, &desc);
-#else
-  comp = AudioComponentFindNext(NULL, &desc);
-#endif
-  assert(comp);
+  if (input_stream_params != NULL) {
```
kinetiknz: Don't double nest ifs here

src/cubeb_audiounit.c
```diff
@@ -614,65 +914,176 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
+    if ((ret = audiounit_create_unit (&input_unit, true,
+            input_stream_params, input_device)) != CUBEB_OK)
+      return ret;
+  }
 
-  stm = calloc(1, sizeof(*stm));
+  if (output_stream_params != NULL) {
```
kinetiknz: As above

src/cubeb_audiounit.c
```diff
@@ -614,65 +914,176 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
+  assert(0 == ret);
   pthread_mutexattr_destroy(&attr);
-  assert(r == 0);
 
+  stm->draining = false;
+  stm->shutdown = 0;
   stm->frames_played = 0;
   stm->frames_queued = 0;
+  stm->frames_read = 0;
   stm->current_latency_frames = 0;
```
kinetiknz: The struct is calloced, so it's not necessary to zero values

src/cubeb_audiounit.c
```diff
@@ -748,25 +1184,29 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
-#if MACOSX_LESS_THAN_106
-    CloseComponent(stm->unit);
-#else
-    AudioComponentInstanceDispose(stm->unit);
-#endif
+  if (stm->input_unit != NULL) {
+    AudioOutputUnitStop(stm->input_unit);
+    AudioUnitUninitialize(stm->input_unit);
+    AudioComponentInstanceDispose(stm->input_unit);
+    stm->input_unit = NULL;
```
kinetiknz: stm is about to be freed, don't NULL values

src/cubeb_audiounit.c
```diff
@@ -748,25 +1184,29 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
+  }
+
+  audiounit_destroy_input_buffer_list_array(stm);
+
+  if (stm->output_unit != NULL) {
+    AudioOutputUnitStop(stm->output_unit);
+    AudioUnitUninitialize(stm->output_unit);
+    AudioComponentInstanceDispose(stm->output_unit);
+    stm->output_unit = NULL;
```
kinetiknz: As above

src/cubeb_resampler_internal.h
```diff
@@ -154,10 +151,22 @@ class cubeb_resampler_speex_one_way : public processor {
   cubeb_resampler_speex_one_way(uint32_t channels,
                                 uint32_t source_rate,
                                 uint32_t target_rate,
-                                int quality);
+                                int quality)
```
kinetiknz: 1. Why is this being moved?
2. Why is this included in audiounit changes?

src/cubeb_ring_array.h
```diff
@@ -0,0 +1,133 @@
...
+
+#ifndef CUBEB_RING_ARRAY_H
+#define CUBEB_RING_ARRAY_H
+
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+/** Ring array of pointers is used to hold buffers. In case that
+    asynchronus producer/consumer callbacks do not arrive in a
```
kinetiknz: typo

src/cubeb_ring_array.h
```diff
@@ -0,0 +1,133 @@
...
+{
+  ra->capacity = RING_ARRAY_CAPACITY;
+  ra->tail = 0;
+  ra->count = 0;
+  memset(ra->pointer_array, 0, sizeof(ra->pointer_array));
+
+  pthread_mutexattr_t attr;
+  pthread_mutexattr_init(&attr);
+  pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_DEFAULT);
+  int ret = pthread_mutex_init(&ra->mutex, &attr);
```
kinetiknz: Just pass a NULL attr, there's no point doing a bunch of work to request the default.

src/cubeb_audiounit.c
```diff
@@ -101,70 +149,209 @@ audiotimestamp_to_latency(AudioTimeStamp const * tstamp, cubeb_stream * stream)
...
+                         UInt32 bus,
+                         UInt32 input_frames,
+                         AudioBufferList * bufs)
 {
-  cubeb_stream * stm;
-  unsigned char * buf;
-  long got;
-  OSStatus r;
-  float panning;
+  cubeb_stream * stream = user_ptr;
```
kinetiknz: stm not stream, please follow existing conventions

src/cubeb_ring_array.h
```diff
@@ -0,0 +1,133 @@
...
+    repeated order the ring array stores the buffers and fetch
+    them in the correct order. */
+#define RING_ARRAY_CAPACITY 8
+
+typedef struct {
+  void* pointer_array[RING_ARRAY_CAPACITY];   /**< Array that hold pointers of the allocated space for the buffers. */
+  unsigned int tail;                          /**< Index of the last element (first to deliver). */
+  int count;                                  /**< Number of elements in the array. */
+  unsigned int capacity;                      /**< Total length of the array. */
+  pthread_mutex_t mutex;                      /**< Mutex to synchronize store/fetch. */
```
kinetiknz: Why does this need internal synchronization? It'd be clearer to let the user of the data structure manage it.  It looks like existing uses are already holding the stream lock when this is accessed anyway.

src/cubeb_ring_array.h
```diff
@@ -0,0 +1,133 @@
...
+/** Set the allocated space to store the data.
+    This must be done before store/fetch.
+    @param ra The ring_array pointer.
+    @param data Pointer to allocated space of buffers.
+    @param index Index between 0 and capacity-1 to store the allocated data buffer.
+    @retval The data pointer on success or NULL on failure. */
+void *
+ring_array_set_data(ring_array * ra, void * data, unsigned int index)
+{
+  if (index < ra->capacity) {
```
kinetiknz: Does it ever make sense for set_data to be called with an index >= RING_ARRAY_CAPACITY?  I'd assert index is valid, set the array, and avoid having to return an error.

src/cubeb_ring_array.h
```diff
@@ -0,0 +1,133 @@
...
+ring_array_destroy(ring_array * ra)
+{
+  pthread_mutex_destroy(&ra->mutex);
+}
+
+/** Get the allocated buffer to be stored with fresh data.
+    @param ra The ring_array pointer.
+    @retval Pointer of the allocated space to be stored with fresh data or NULL if full. */
+void *
+ring_array_store_buffer(ring_array * ra)
```
kinetiknz: store_buffer and fetch_buffer are ambiguous names, please rename them to make it clearer what they do.

src/cubeb_ring_array.h
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/** Set the allocated space to store the data.
+    This must be done before store/fetch.
+    @param ra The ring_array pointer.
+    @param data Pointer to allocated space of buffers.
+    @param index Index between 0 and capacity-1 to store the allocated data buffer.
+    @retval The data pointer on success or NULL on failure. */
+void *
+ring_array_set_data(ring_array * ra, void * data, unsigned int index)
```
kinetiknz: Is the caller expected/required to call this for each indice between 0 and RING_ARRAY_CAPACITY? Where is this documented/checked?

src/cubeb_audiounit.c
```diff
@@ -101,70 +149,209 @@ audiotimestamp_to_latency(AudioTimeStamp const * tstamp, cubeb_stream * stream)
...
-    size_t rem_bytes = (nframes - got) * stm->sample_spec.mBytesPerFrame;
+  pthread_mutex_unlock(&stream->mutex);
+  return noErr;
+}
 
-    stm->draining = 1;
+static void
+audiounit_make_silent(AudioBufferList * ioData)
+{
+  for(UInt32 i = 0; i<ioData->mNumberBuffers;i++) {
```
kinetiknz: Spaces...

src/cubeb_audiounit.c
```diff
@@ -538,73 +725,186 @@ audiounit_destroy(cubeb * ctx)
...
+  /* Initialize buffer list. */
+  memset(&stream->input_buffer_list, 0, sizeof(stream->input_buffer_list));
+  /* Create input descriptor. */
+  AudioStreamBasicDescription src_desc = stream->input_desc;
+  src_desc.mSampleRate = stream->input_hw_rate;
+  /* Set data to ring array. */
+  for (int i = 0; i < RING_ARRAY_CAPACITY; ++i) {
+    /* Allocate the data (AudioBufferList here). */
+    if (audiounit_buflst_init_single_buffer(&stream->input_buffer_list[i],
+                                            &src_desc,
```
kinetiknz: This should just take a size rather than being forced to know about AudioStreamBasicDescriptions.

src/cubeb_ring_array.h
```diff
@@ -0,0 +1,133 @@
...
+/** Ring array of pointers is used to hold buffers. In case that
+    asynchronus producer/consumer callbacks do not arrive in a
+    repeated order the ring array stores the buffers and fetch
+    them in the correct order. */
+#define RING_ARRAY_CAPACITY 8
+
+typedef struct {
+  void* pointer_array[RING_ARRAY_CAPACITY];   /**< Array that hold pointers of the allocated space for the buffers. */
+  unsigned int tail;                          /**< Index of the last element (first to deliver). */
+  int count;                                  /**< Number of elements in the array. */
```
kinetiknz: Why is this an int?

src/cubeb_ring_array.h
```diff
@@ -0,0 +1,133 @@
...
+#endif
+
+/** Ring array of pointers is used to hold buffers. In case that
+    asynchronus producer/consumer callbacks do not arrive in a
+    repeated order the ring array stores the buffers and fetch
+    them in the correct order. */
+#define RING_ARRAY_CAPACITY 8
+
+typedef struct {
+  void* pointer_array[RING_ARRAY_CAPACITY];   /**< Array that hold pointers of the allocated space for the buffers. */
```
kinetiknz: I would like to understand why this wouldn't be simpler if input_buffer_list and input_buffer_list_array were merged so this could own the data.

src/cubeb_audiounit.c
```diff
@@ -101,70 +149,209 @@ audiotimestamp_to_latency(AudioTimeStamp const * tstamp, cubeb_stream * stream)
...
-  if (got < 0) {
-    /* XXX handle this case. */
-    assert(false);
-    pthread_mutex_unlock(&stm->mutex);
+  /* Input only. Call the user callback through resampler.
+     Resampler will deliver input buffer in the correct rate. */
+  frames = input_frames;
+  AudioBufferList * fetch_input_buffer_list = ring_array_fetch_buffer(&stream->input_buffer_list_array);
+  assert(fetch_input_buffer_list && "fetch buffer is null in the input");
+  input_buffer = fetch_input_buffer_list->mBuffers[0].mData;
```
kinetiknz: Get rid of input_buffer and pass fetch_input_buffer_list->mBuffers[0].mData directly

fetch_input_buffer_list and store_input_buffer_list variable names are too long and not clear enough, this naming issue flows on from the naming problem with the ring buffer access.

src/cubeb_audiounit.c
```diff
@@ -101,70 +149,209 @@ audiotimestamp_to_latency(AudioTimeStamp const * tstamp, cubeb_stream * stream)
...
+                                   &input_frames,
+                                   output_buffer,
+                                   output_frames);
+
+  /* Cleanup the input buffer to make sure that we have fresh data. */
+  if (input_buffer) {
+    audiounit_make_silent(fetch_input_buffer_list);
   }
 
+
```
kinetiknz: Unwanted extra newline

src/cubeb_audiounit.c
```diff
@@ -101,70 +149,209 @@ audiotimestamp_to_latency(AudioTimeStamp const * tstamp, cubeb_stream * stream)
...
+    pthread_mutex_unlock(&stream->mutex);
+    return noErr;
+  }
+
+  AudioFormatFlags outaff;
+  float panning;
+  size_t outbpf;
+
+  outbpf = stream->output_desc.mBytesPerFrame;
+  stream->draining = (outframes < output_frames);
```
kinetiknz: Unnecessary parens

src/cubeb_audiounit.c
```diff
@@ -538,73 +725,186 @@ audiounit_destroy(cubeb * ctx)
...
-
-  assert(!input_stream_params && "not supported");
-  if (input_device || output_device) {
-    /* Device selection not yet implemented. */
-    return CUBEB_ERROR_DEVICE_UNAVAILABLE;
-  }
-
-  assert(context);
-  *stream = NULL;
+  memset(ss, 0, sizeof(AudioStreamBasicDescription));
```
kinetiknz: Just initialize every field correctly, then there's no need to memset up front.

src/cubeb_audiounit.c
```diff
@@ -538,73 +725,186 @@ audiounit_destroy(cubeb * ctx)
...
 static int
-audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stream_name,
-                      cubeb_devid input_device,
-                      cubeb_stream_params * input_stream_params,
-                      cubeb_devid output_device,
-                      cubeb_stream_params * output_stream_params,
-                      unsigned int latency,
-                      cubeb_data_callback data_callback, cubeb_state_callback state_callback,
-                      void * user_ptr)
+audio_stream_desc_init (AudioStreamBasicDescription  * ss,
```
kinetiknz: No space between init and (

src/cubeb_audiounit.c
```diff
@@ -101,70 +149,209 @@ audiotimestamp_to_latency(AudioTimeStamp const * tstamp, cubeb_stream * stream)
...
+
+  if (outframes < 0) {
+    stream->shutdown = 1;
+    pthread_mutex_unlock(&stream->mutex);
+    return noErr;
+  }
+
+  AudioFormatFlags outaff;
+  float panning;
+  size_t outbpf;
```
kinetiknz: Declare as they're initialized, no point doubling the number of lines.

src/cubeb_audiounit.c
```diff
@@ -694,42 +1105,67 @@ audiounit_stream_init(cubeb * context, cubeb_stream ** stream, char const * stre
...
 
-  r = AudioUnitSetProperty(stm->unit, kAudioUnitProperty_StreamFormat, kAudioUnitScope_Input,
-                           0, &ss, sizeof(ss));
-  if (r != 0) {
-    audiounit_stream_destroy(stm);
+  /* We use a resampler because input AudioUnit operates
+   * reliable only in the capture device sample rate.
+   * Resampler will convert it to the user sample rate
+   * and deliver it to the callback. */
+  int32_t target_sample_rate;
```
kinetiknz: cubeb_stream_params rate is an unsigned int, why is this a different type?

src/cubeb_audiounit.c
```diff
@@ -538,73 +725,186 @@ audiounit_destroy(cubeb * ctx)
...
+}
+
+static int
+audiounit_buflst_init_single_buffer (AudioBufferList * buflst,
+                                     const AudioStreamBasicDescription * desc,
+                                     uint32_t frames)
+{
+  size_t size = desc->mBytesPerFrame * frames;
+
+  if ((buflst->mBuffers[0].mData = malloc(size)) == NULL) {
```
kinetiknz: Separate the assignment from the test.  This applies to every case where you assign and check a value in a single if clause.

kinetiknz:
This needs further review, but please address the issues I've raised so far.

It would be easier to review future patches if they were split up into separate logical changes within a PR.

src/cubeb_audiounit.c
achronop: No. Actually I haven't made any iphone change. I took a partial ready implementation from Haakon which has made those changes. Unfortunately I did not make him the same question. I'll try now to contact him and ask the status.

src/cubeb_audiounit.c
achronop: I am not sure I get it here. `AudioStreamBasicDescription` is used as an input to `audiounit_buflst_init_single_buffer()` method. Do you propose to change this latter method to accept size inputs and stop using `AudioStreamBasicDescription` inside `audiounit_init_input_buffer_list_array()` method. In this case we need two size numbers the byte per frame and channels per frame.

src/cubeb_resampler.cpp
```diff
@@ -33,36 +33,19 @@ to_speex_quality(cubeb_resampler_quality q)
...
+         !input_buffer && *input_frames_count == 0 ||
          !output_buffer && output_frames== 0);
 
-  if (*input_frames_count != output_frames) {
+  if (output_buffer == nullptr) {
+    output_frames = *input_frames_count;
+  }
+
+  if (input_buffer && *input_frames_count != output_frames) {
     assert(*input_frames_count > output_frames);
```
achronop: Pull request #80 handle the same (or better similar) changes inside `noop_resampler::fill()` method.

src/cubeb_resampler_internal.h
achronop: 1. The implementation method moved in the header due to link error when building test_resampler.cpp file. The problem is that `cubeb_resampler_speex_one_way` is a template class and any traslation unit that wants to use template class need to have access to the whole definition. The weird thing is that padenot with (gcc 4.8) does not get a compile (link) error for that. I had a link error (compiler version Apple LLVM version 7.0.2 (clang-700.1.81)) which solved by moving the implementation above in the header file.

2. All changes in resampler are errors found during audiounit testing and that's why they are including in that patch. Some of them are also included in new pull request #80. If you want me to move them in a separate PR let me know and I will follow.

src/cubeb_ring_array.h
achronop: My idea was to create something general in order to be used from other platforms too. Currently it is in a rather primitive state but still adequate for audiounit case. If we need something more sophisticated in the future we can build on the top of that. That's why I did not merge it with audiounit buffers and made it not to own any data. I just explain my approach, if you believe that it's better to change it to something else please let me know and I will follow.

src/cubeb_ring_array.h
achronop: This is because I can assert for non-negative value in order to verify correct operation.

src/cubeb_ring_array.h
achronop: You are right internal synchronization is redundant in our case. I implemented it in order to keep it thread safe and cover the general usage. I will remove the mutex and we can reconsider when we need it.

src/cubeb_ring_array.h
achronop: Inside get_*_buffer methods it is asserted that the the pointer_array contains at least one non-null data. Also the comments that describing the method mention that.

src/cubeb_audiounit.c
achronop: I see your point. Probably it was started that way (and continued by me) because the api requires buffer list. If we change it we need to create a buffer list every time we want to call a method from the api and attach the buffer. I think this is small compering to other changes required to replace AudioBufferList with AudioBuffer to the whole file. I will finish with all other updates and start checking this one

achronop:
Thank you for the comments. I have finished with the update. I provide answers/clarifications when necessary otherwise the source has been updated.

Sorry for the one long patch. I applied the comments in separated commits in order to be more readable. I will squash at the end. If it is still helpful I can reformat the whole history before review it again.

achronop:
> It doesn't look like this change is safe to make without making further fixes, e.g.
> wasapi: devid is the same as device_id, so this is now a double-free.
> winmm: devid isn't a heap pointer.

Before that the question is whether you agree with the proposed change. If yes I could update missing parts. Its hard to test though. The change is based on the assumption that every backend have to maintain a deep copy of devid.  

On the other hand if we do not want this approach I need to find another alternative for audiounit. The only one I can see is something similar to winmm approach, since devid is also a small integer. It is painless since does not affect any other backend. Please let me know what you think.

src/cubeb_resampler_internal.h
kinetiknz: Yeah, anything that wasn't included in #80 please separate out and submit separately.

src/cubeb_audiounit.c
kinetiknz: Yeah, I'd prefer to pass just bytes per frame and channels per frame rather than a whole ASBD.

src/cubeb_ring_array.h
kinetiknz: I think the only place this matters is where ra->count is decremented and then asserted to be >= 0.  You could make it the same type as tail, remove all of the int casts, and assert it's >= 1 before subtracting.

src/cubeb_ring_array.h
kinetiknz: Since it's only used in one place, and given the current design/API, I think it'd be clearer to not share this and rework it to own the buffers.  The current design is a bit too complex and rather than trying to solve that for a hypothetical second user, let's simplify it as much as possible for the current use.

kinetiknz:
> Before that the question is whether you agree with the proposed change. If yes I could update missing parts. Its hard to test though. The change is based on the assumption that every backend have to maintain a deep copy of devid.

Okay, I wasn't clear that it was a question/proposal because you changed some but not all of the backends.

> On the other hand if we do not want this approach I need to find another alternative for audiounit. The only one I can see is something similar to winmm approach, since devid is also a small integer. It is painless since does not affect any other backend. Please let me know what you think.

I don't have a strong opinion, but my preference is to make cubeb_device_info_destroy backend-specific so that backends can do what they like.  That keeps the reasoning about the behaviour of allocations within cubeb_device_info local to each backend, which seems simpler.

kinetiknz:
I think the Travis CI issues were server-side.  I kicked off a rerun this morning and it timed out again, but I just ran another one and it passed.

achronop:
Thank you for the extra comments. I pushe all the updates. Unfortunately Travis failed which does not happen locally. I will continue working on that. In the mean time you can take a look at the changes.

> I don't have a strong opinion, but my preference (if we were doing this over) is to make cubeb_device_info_destroy backend-specific so that backends can do what they like. That keeps the reasoning about the behavior of allocations within cubeb_device_info local to each backend, which seems simpler. It's a trade-off, because that approach duplicates a little code, but reduces dependencies. I'm fine with keeping your proposed approach for now.

In order to make the `cubeb_device_info_destroy` backend-specific we need to enhance it to get cubeb context argument. This is another interface change. After check the winmm approach I realized I could take the same approach (never thought before) which does not involve other backend change. So I did that and removed the free statement. Please let me know what you think.

src/cubeb_ring_array.h
achronop: I reworked this to own the buffers. I also changed the array allocation to be dynamic. This is good for capture only scenario in which we need just one buffer.

kinetiknz:
Thanks Alex, all looks good!

PR #80, opened 2016-03-07
Resampler fixes

padenot:
Those are fixes for the resampler patches, although I still need to write regression tests.


padenot:
Those tests would have caught the issues.

PR #82, opened 2016-03-14
Audiounit device change notifications

achronop:
This PR is the implementation of device change notifications. The idea is that every time a device is plugged or unplugged the corresponding callback will be called.

One thing to note in the implementation is that Apple framework does not differentiate if the device is input or output device. I worked around this by keeping the number of devices in the desired type and compering with the new updated number. A second thing to note is that the current implementation does not accept multiply register of the notify callback without deregister the previous one. If that is not acceptable please let me know.

Thanks!

### DIFF
--- a/src/cubeb_audiounit.c
+++ b/src/cubeb_audiounit.c
@@ -1730,7 +1730,7 @@ audiounit_number_of_devices(cubeb_device_type devtype)
   }
 
   AudioObjectID devices[count];
-  ret = AudioObjectGetPropertyData(kAudioObjectSystemObject, &adr, 0, NULL, &size, (void *)&devices);
+  ret = AudioObjectGetPropertyData(kAudioObjectSystemObject, &adr, 0, NULL, &size, &devices);
   if (ret != noErr) {
     return 0;
   }
@@ -1742,7 +1742,7 @@ audiounit_number_of_devices(cubeb_device_type devtype)
   uint32_t dev_count = 0;
   for(uint32_t i = 0; i < count; ++i) {
     /* For device in the given scope channel must be > 0. */
-    dev_count += !!audiounit_get_channel_count(devices[i], scope);
+    dev_count += audiounit_get_channel_count(devices[i], scope) > 0;
   }
 
   return dev_count;
--- a/src/cubeb_audiounit.c
+++ b/src/cubeb_audiounit.c
@@ -70,6 +70,8 @@ struct cubeb {
   int limit_streams;
   cubeb_device_collection_changed_callback collection_changed_callback;
   void * collection_changed_user_ptr;
+  cubeb_device_type collection_changed_devtype;
+  uint32_t devtype_device_count;
 };
 
 struct cubeb_stream {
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.c
```diff
@@ -1707,6 +1711,125 @@ audiounit_enumerate_devices(cubeb * context, cubeb_device_type type,
...
+  }
+  /* Total number of input and output devices. */
+  uint32_t count = (uint32_t)(size / sizeof(AudioObjectID));
+
+  if (devtype == (CUBEB_DEVICE_TYPE_INPUT | CUBEB_DEVICE_TYPE_OUTPUT)) {
+    return count;
+  }
+
+  AudioObjectID devices[count];
+  ret = AudioObjectGetPropertyData(kAudioObjectSystemObject, &adr, 0, NULL, &size, (void *)&devices);
```
kinetiknz: Is the void* cast necessary?

src/cubeb_audiounit.c
```diff
@@ -1707,6 +1711,125 @@ audiounit_enumerate_devices(cubeb * context, cubeb_device_type type,
...
+  }
+
+  AudioObjectPropertyScope scope = (devtype == CUBEB_DEVICE_TYPE_INPUT) ?
+                                         kAudioObjectPropertyScopeInput :
+                                         kAudioObjectPropertyScopeOutput;
+
+  uint32_t dev_count = 0;
+  for(uint32_t i = 0; i < count; ++i) {
+    /* For device in the given scope channel must be > 0. */
+    dev_count += !!audiounit_get_channel_count(devices[i], scope);
```
kinetiknz: `dev_count += audiounit_get_channel_count(devices[i], scope) > 0;`
seems clearer to me.

src/cubeb_audiounit.c
```diff
@@ -1707,6 +1711,125 @@ audiounit_enumerate_devices(cubeb * context, cubeb_device_type type,
...
+    /* Listener removed while waiting in mutex, abort. */
+    pthread_mutex_unlock(&context->mutex);
+    return noErr;
+  }
+
+  /* Differentiate input from output changes. */
+  if (context->collection_changed_devtype == CUBEB_DEVICE_TYPE_INPUT ||
+      context->collection_changed_devtype == CUBEB_DEVICE_TYPE_OUTPUT) {
+    uint32_t new_number_of_devices = audiounit_number_of_devices(context->collection_changed_devtype);
+    if (context->devtype_device_count == new_number_of_devices) {
```
kinetiknz: If I disconnect a device of type A and connect a device of type B, doesn't this mean that we won't notice because the device count doesn't change, even though the type of device attached may have?

kinetiknz:
Thanks, looks good, just a couple of questions above for you.

> A second thing to note is that the current implementation does not accept multiply register of the notify callback without deregister the previous one. If that is not acceptable please let me know.

That's fine, the current cubeb API is intentionally designed that way for simplicity.

src/cubeb_audiounit.c
achronop: The way that it works is that the `devtype_device_count` data member holds the number of devices of type `collection_changed_devtype` and not the total number of devices. Since we are inside the if branch, stored `collection_changed_devtype` expected to be either input or output not both. So if we register for a specific type and remove or add a device of that type the `new_number_of_devices` will be something different that the stored count. On the other hand if the change was from a device of the other type the `new_number_of_devices` will be the same with the stored one. So we are not affected by the number of devices of the other type.

achronop:
This looks like a infra issue. On my private build only one failed https://travis-ci.org/achronop/cubeb/builds/116095445 and also it successes locally. Can you please re-trigger the execution in case the error is not permanent?

achronop:
This looks like a infra issue. On my private build passed after sometime https://travis-ci.org/achronop/cubeb/builds/116095445 . Can you please re-trigger the execution? Thanks!

src/cubeb_audiounit.c
kinetiknz: Right, I realize that.  I wasn't clear enough in my original comment -- "type A" and "type B" don't refer to different devtypes. simply different (types of) devices.  I'm asking what happens if, for a given devtype, I disconnect a device and connect a new device (e.g. unplug USB headphones, connect USB speakers).  The number of devices of the specified devtype remains the same, but the type of devices attached are now different... and we don't send a device change notification in that case.  Or is that not a problem for some reason I'm missing?  Are we *guaranteed* to get a distinct callback for a disconnect and new connection (causing the count to drop and rise), or is it possible to miss them in the situation I've described?

kinetiknz: I can't find documentation for this stuff online (it used to be on developer.apple.com, but I'm failing to find the actual API docs there now).

Looking at similar device change/enumeration code (e.g. module-coreaudio-detect.c in PulseAudio), rather than simply checking the device count they iterate the devices comparing the device id (from AudioObjectGetPropertyData) with each cached device to detect add/remove/change.

achronop: Sorry I got confused by meaning of type. I considered this scenario and I though it was exaggerated. This is because device plug or unplug requires human activity which makes it very unlikely to have two different events in the exact same moment and even if this heppens once the API might provide distinct callback. But since this is your concern too (which makes it a valid scenario) and I am not able to verify the API behavior through documentation I will make my implementation more robust by comparing the device ids for changes when the count is the same.

kinetiknz:
I'll merge this as-is, but we still need to address https://github.com/kinetiknz/cubeb/pull/82#discussion_r56318044 in some way.

PR #83, opened 2016-03-17
Document Windows build prerequisites

padenot:
Part of all the issues I've had on Windows would have been caught by running the tests, but I was running a hacked up visual studio solution so I kind of forgot to run them all. I decided to set up a proper development environment to address this, and faced quite a number of issues (knowing close to nothing about mingw and all).

This is what I now use to work on the windows backends, and it works fine, it can build and run/debug the tests, maybe it would be good to have it in the INSTALL document ?


kinetiknz:
Thanks, looks similar to my setup.

I signed up for AppVeyor so we can run the CI on Windows machines, but haven't had a chance to set the environment up yet.

Issue #85, opened 2016-03-17
audiounit: Handle device change callback coalescing


kinetiknz:
The concern is that device change callbacks could be coalesced (speculation), which would cause the count-based device tracking to miss some device changes.

See the discussion at https://github.com/kinetiknz/cubeb/pull/82#discussion_r56318044.

I think @achronop is already working on fixing this.

kinetiknz:
Fixed by #86.

PR #86, opened 2016-03-17
audiounit: Handle device change callback coalescing

achronop:
Handle the case that audio device of the same type coalescing. The general idea is that devids are stored in arrays which are compared when the number of devices before and after change is the same. Also added to unregister callback in cubeb_destroy if necessary.  

PR is addressing issue #85


kinetiknz:
Thanks Alex!

PR #89, opened 2016-03-23
More small fixes for the resampler and wasapi code

padenot:
Getting there, it's almost green.

### DIFF
--- a/test/test_resampler.cpp
+++ b/test/test_resampler.cpp
@@ -5,9 +5,6 @@
  * accompanying file LICENSE for details.
  */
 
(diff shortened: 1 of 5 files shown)


test/test_resampler.cpp
```diff
@@ -5,9 +5,6 @@
  * accompanying file LICENSE for details.
  */
 
-#define OUTSIDE_SPEEX
-#define RANDOM_PREFIX speex
-
```
kinetiknz: This breaks the build if a system speex isn't available.

PR #91, opened 2016-03-24
More little fixes

padenot:
Mostly warning as error and asserts.


kinetiknz:
I think enabling asserts caused test_resampler to fail on Travis? Can you check that please.

PR #95, opened 2016-03-26
Correct gecko macros

achronop:
Change the version of _XOPEN_SOURCE used from 5 to 6. Version 5 caused compilation error in gecko. Also correct the xpcom scoped name of test_duplex.


achronop:
Travis expected to be green after merging #94

PR #100, opened 2016-04-18
Make the unique_ptr replacment technique compatible with libc+

ehsan:


ehsan:
Mozilla bug for landing this: https://bugzilla.mozilla.org/show_bug.cgi?id=1265541

PR #101, opened 2016-04-21
Input device change callback

achronop:
Register the listener for default input device. Also it check if input, output or both is activated in order to register for the corresponding default devices.


achronop:
New commits pushed

src/cubeb_audiounit.c
```diff
@@ -467,48 +465,84 @@ audiounit_property_listener_callback(AudioObjectID id, UInt32 address_count,
   return noErr;
 }
 
+OSStatus
+audiounit_add_listener(cubeb_stream * stm, AudioDeviceID id, AudioObjectPropertySelector selector,
```
padenot: Can you make it fit in 80 char columns ?

PR #103, opened 2016-04-22
Implement device notification callbacks for the WASAPI backend.

padenot:


padenot:
This is missing a number of things, I'll rework it.

PR #107, opened 2016-05-05
audiounit: Handle input render failure.

achronop:
In case of AudioUnitRender in input callback clean the corresponding input buffer and return the error. This is due to Bug 1269472

### DIFF
--- a/src/cubeb_audiounit.c
+++ b/src/cubeb_audiounit.c
@@ -151,6 +151,29 @@ audiotimestamp_to_latency(AudioTimeStamp const * tstamp, cubeb_stream * stream)
   return ((pres - now) * stream->output_desc.mSampleRate) / 1000000000LL;
 }
 
+static void
+audiounit_make_silent(AudioBuffer * ioData)
+{
+  memset(ioData->mData, 0, ioData->mDataByteSize);
+}
+
+static void
+PrintResult(OSStatus result, const char *operation)
+{
+  char errorString[20];
+  // see if it appears to be a 4-char-code
+  *(UInt32 *)(errorString + 1) = CFSwapInt32HostToBig(result);
+  if (isprint(errorString[1]) && isprint(errorString[2]) && isprint(errorString[3]) && isprint(errorString[4])) {
+    errorString[0] = errorString[5] = '\'';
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.c
```diff
@@ -151,6 +151,29 @@ audiotimestamp_to_latency(AudioTimeStamp const * tstamp, cubeb_stream * stream)
...
+}
+
+static void
+PrintResult(OSStatus result, const char *operation)
+{
+  char errorString[20];
+  // see if it appears to be a 4-char-code
+  *(UInt32 *)(errorString + 1) = CFSwapInt32HostToBig(result);
+  if (isprint(errorString[1]) && isprint(errorString[2]) && isprint(errorString[3]) && isprint(errorString[4])) {
+    errorString[0] = errorString[5] = '\'';
```
jesup: Why the errorstring+1 thing to insert an apostrophe?  Why not do something like:
if (all are printable) {
  marker = "\'";
} else {
  marker = "";
  sprintf(...);
}
LOG("Result: %s (%s%s)\n", operation, marker, errorString);
And that assumes the apostrophe is important (I suspect it's not, unless you *need* to separate "1234" from 1234); if not just don't bother.
All that said: this looks fine and safe.

achronop: This is adjusted from similar methods from apple sample code. IMO the reason that the apostrophe matters is that some OSStatus errors are four character codes and some other negative int32. By using the apostrophe you indicate that this is the readable error code. But indeed it's not important. The two case would look like
Result: Input AudioUnitRender failed with error ('nope')
Result: Input AudioUnitRender failed with error (-10879)

Using the marker is another way to do the same thing and looks more convenient than pointer arithmetic. Probably one more format specifier is needed.
LOG("Result: %s (%s%s**%s**)\n", operation, marker, errorString, **marker**);

kinetiknz:
I merged this as https://github.com/kinetiknz/cubeb/commit/368a1dd9e8ba0d2c5d10f8eeff671ff741c5483a without the PrintResult stuff, which is unnecessary and also doesn't fit the coding style.

PR #108, opened 2016-05-23
Add duplex JACK backend

zamaudio:

### DIFF
--- a/src/cubeb_jack.cpp
+++ b/src/cubeb_jack.cpp
@@ -1,6 +1,7 @@
 /*
  * Copyright © 2012 David Richards
  * Copyright © 2013 Sebastien Alaiwan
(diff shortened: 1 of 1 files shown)


padenot:
I've looked a this multiple times, and it works fine with Firefox, and passes all cubeb tests, I think we can merge it. Thoughts ?

achronop:
The get current device and enumerate device methods contain static hard coded values. Never the less this is something we can take care later.

kinetiknz:
No objections from me.  I haven't look at it at all, but if it works with in Gecko that's probably an improvement over the current situation.

padenot:
All right, merged, thanks again to @zamaudio ! Now, onto some integration in Firefox.

kinetiknz:
Thanks for working on this!

PR #111, opened 2016-06-03
test_sanity: Fix for zeroing buffers of correct type

zamaudio:


kinetiknz:
Thanks!

PR #114, opened 2016-06-10
Opensl es full duplex

achronop:
Enhancements for full duplex on Android. Resampler and test files changes pushed in previous PR. 
 
One thing to note is that full duplex operation is implemented in different thread. Initial implementation had fd on out callback but this created delay of the output comparing to input which degraded the final result.

Two new files added, one required due to Gecko implementation and a new data structure to transfer input/output buffer from/to consuming thread.

---
This change is [Reviewable](https://reviewable.io/reviews/kinetiknz/cubeb/114)

### DIFF
--- a/src/cubeb-sles.h
+++ b/src/cubeb-sles.h
@@ -1,0 +1,43 @@
+/*
+ * Copyright © 2016 Mozilla Foundation
+ *
+ * This program is made available under an ISC-style license.  See the
+ * accompanying file LICENSE for details.
+ */
+
+#ifndef _CUBEB_SLES_H_
+#define _CUBEB_SLES_H_
+#include <SLES/OpenSLES.h>
+
+static SLresult
+cubeb_get_sles_engine(SLObjectItf *           pEngine,
+                      SLuint32                numOptions,
+                      const SLEngineOption *  pEngineOptions,
+                      SLuint32                numInterfaces,
+                      const SLInterfaceID *   pInterfaceIds,
+                      const SLboolean *       pInterfaceRequired)
+{
+  return slCreateEngine(pEngine,
+                        numOptions,
+                        pEngineOptions,
+                        numInterfaces,
+                        pInterfaceIds,
+                        pInterfaceRequired);
+}
+
+static void
+cubeb_destroy_sles_engine(SLObjectItf * self)
+{
+  if (*self != NULL) {
+      (**self)->Destroy(*self);
+      *self = NULL;
+  }
+}
+
+static SLresult
+cubeb_realize_sles_engine(SLObjectItf self)
+{
+  return (*self)->Realize(self, SL_BOOLEAN_FALSE);
+}
+
+#endif
--- a/src/cubeb_array_queue.h
+++ b/src/cubeb_array_queue.h
@@ -1,0 +1,118 @@
+/*
+ * Copyright © 2016 Mozilla Foundation
+ *
+ * This program is made available under an ISC-style license.  See the
+ * accompanying file LICENSE for details.
+ */
+
+#ifndef CUBEB_ARRAY_QUEUE_H
+#define CUBEB_ARRAY_QUEUE_H
+
+#include <assert.h>
+#include <pthread.h>
+#include <unistd.h>
+
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+typedef struct
+{
+  void ** buf;
+  size_t num;
+  size_t writePos;
+  size_t readPos;
+  pthread_mutex_t mutex;
+  pthread_cond_t empty_convar;
+} array_queue;
+
+array_queue * array_queue_create(size_t num)
+{
+  assert(num != 0);
+  array_queue * new_queue = (array_queue*)calloc(1, sizeof(array_queue));
+  new_queue->buf = (void **)calloc(1, sizeof(void*) * num);
+  new_queue->readPos = 0;
+  new_queue->writePos = 0;
+  new_queue->num = num;
+
+  pthread_mutex_init(&new_queue->mutex, NULL);
+  pthread_cond_init(&new_queue->empty_convar, NULL);
+
+  return new_queue;
+}
+
+void array_queue_destroy(array_queue * aq)
+{
+  assert(aq);
+
+  free(aq->buf);
+  pthread_mutex_destroy(&aq->mutex);
+  pthread_cond_destroy(&aq->empty_convar);
+  free(aq);
+}
+
+int array_queue_push(array_queue * aq, void * item)
+{
+  assert(item);
+
+  pthread_mutex_lock(&aq->mutex);
+  int ret = -1;
+  if(aq->buf[aq->writePos % aq->num] == NULL)
+  {
+    aq->buf[aq->writePos % aq->num] = item;
+    aq->writePos = (aq->writePos + 1) % aq->num;
+    ret = 0;
+  }
+  // else queue is full
+  pthread_cond_signal(&aq->empty_convar);
+  pthread_mutex_unlock(&aq->mutex);
+  return ret;
+}
+
+void* array_queue_pop(array_queue * aq)
+{
+  pthread_mutex_lock(&aq->mutex);
+  void * value = aq->buf[aq->readPos % aq->num];
+  if(value)
+  {
+    aq->buf[aq->readPos % aq->num] = NULL;
+    aq->readPos = (aq->readPos + 1) % aq->num;
+  }
+  pthread_mutex_unlock(&aq->mutex);
+  return value;
+}
+
+void* array_queue_wait_pop(array_queue * aq)
+{
+  pthread_mutex_lock(&aq->mutex);
+  while (aq->readPos == aq->writePos) {
+    // Wait here if queue is empty
+    pthread_cond_wait(&aq->empty_convar, &aq->mutex);
+  }
+  void * value = aq->buf[aq->readPos % aq->num];
+  // Since it has been waiting it is expected to
+  // return a non NULL value.
+  assert(value);
+  aq->buf[aq->readPos % aq->num] = NULL;
+  aq->readPos = (aq->readPos + 1) % aq->num;
+  pthread_mutex_unlock(&aq->mutex);
+  return value;
+}
+
+size_t array_queue_get_size(array_queue * aq)
+{
+  pthread_mutex_lock(&aq->mutex);
+  ssize_t r = aq->writePos - aq->readPos;
+  if (r < 0) {
+    r = aq->num + r;
+    assert(r >= 0);
+  }
+  pthread_mutex_unlock(&aq->mutex);
+  return (size_t)r;
+}
+
+#if defined(__cplusplus)
+}
+#endif
+
+#endif //CUBE_ARRAY_QUEUE_H
--- a/src/cubeb_opensl.c
+++ b/src/cubeb_opensl.c
@@ -9,6 +9,7 @@
 #include <dlfcn.h>
 #include <stdlib.h>
 #include <pthread.h>
+#include <errno.h>
 #include <SLES/OpenSLES.h>
 #include <math.h>
 #include <time.h>
@@ -19,7 +20,31 @@
 #include <SLES/OpenSLES_Android.h>
 #include <android/log.h>
 #include <android/api-level.h>
+
+//#define LOGGING_ENABLED
+#ifdef LOGGING_ENABLED
 #define LOG(args...)  __android_log_print(ANDROID_LOG_INFO, "Cubeb_OpenSL" , ## args)
+#else
+#define LOG(...)
+#endif
+
+//#define TIMESTAMP_ENABLED
+#ifdef TIMESTAMP_ENABLED
+#define FILENAME (strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__)
+#define LOG_TS(args...)  __android_log_print(ANDROID_LOG_INFO, "Cubeb_OpenSL ES: Timestamp(usec)" , ## args)
+#define TIMESTAMP(msg) do {                           \
+  struct timeval timestamp;                           \
+  int ts_ret = gettimeofday(&timestamp, NULL);        \
+  if (ts_ret == 0) {                                  \
+    LOG_TS("%lld: %s (%s %s:%d)", timestamp.tv_sec * 1000000LL + timestamp.tv_usec, msg, __FUNCTION__, FILENAME, __LINE__);\
+  } else {                                            \
+    LOG_TS("Error: %s (%s %s:%d) - %s", msg, __FUNCTION__, FILENAME, __LINE__);\
+  }                                                   \
+} while(0)
+#else
+#define TIMESTAMP(...)
+#endif
+
 #define ANDROID_VERSION_GINGERBREAD_MR1 10
 #define ANDROID_VERSION_LOLLIPOP 21
 #define ANDROID_VERSION_MARSHMALLOW 23
@@ -28,6 +53,9 @@
 #include "cubeb-internal.h"
 #include "cubeb_resampler.h"
 #include "cubeb-sles.h"
+#include "cubeb_array_queue.h"
+
+#define DEFAULT_SAMPLE_RATE 48000
 
 static struct cubeb_ops const opensl_ops;
 
@@ -58,27 +86,123 @@ struct cubeb_stream {
   SLPlayItf play;
   SLBufferQueueItf bufq;
   SLVolumeItf volume;
-  uint8_t *queuebuf[NBUFS];
+  void ** queuebuf;
+  uint32_t queuebuf_capacity;
   int queuebuf_idx;
   long queuebuf_len;
   long bytespersec;
   long framesize;
+  /* Total number of played frames.
+   * Synchronized by stream::mutex lock. */
   long written;
+  /* Flag indicating drainig. Synchronized
+   * by stream::mutex lock. */
   int draining;
   cubeb_stream_type stream_type;
 
+  array_queue * output_queue;
+
+  /* Flags to determine in/out.*/
+  uint32_t input_enabled;
+  uint32_t output_enabled;
+
+  /* Recorder abstract object. */
+  SLObjectItf recorderObj;
+  /* Recorder Itf for input capture. */
+  SLRecordItf recorderItf;
+  /* Buffer queue for input capture. */
+  SLAndroidSimpleBufferQueueItf recorderBufferQueueItf;
+  /* Store input buffers. */
+  void ** input_buffer_array;
+  /* The capacity of the array.
+   * On capture only can be small (4).
+   * On full duplex is calculated to
+   * store 1 sec of data buffers. */
+  uint32_t input_array_capacity;
+  /* Current filled index of input buffer array.
+   * It is initiated to -1 indicating buffering
+   * have not started yet. */
+  int input_buffer_index;
+  /* Length of input buffer.*/
+  uint32_t input_buffer_length;
+  /* Input frame size */
+  uint32_t input_frame_size;
+  /* Device sampling rate. If user rate is not
+   * accepted an compatible rate is set. If it is
+   * accepted this is equal to params.rate. */
+  uint32_t input_device_rate;
+  /* Exchange input buffers between input
+   * and full duplex threads. */
+  array_queue * input_queue;
+  /* Silent input buffer used on full duplex. */
+  void * input_silent_buffer;
+  /* Number of input frames from the start of the stream*/
+  uint32_t input_total_frames;
+  /* Flag to stop the execution of user callback and
+   * close all working threads. Synchronized by
+   * stream::mutex lock. */
+  uint32_t shutdown;
+  /* Store user callback. */
   cubeb_data_callback data_callback;
+  /* Store state callback. */
   cubeb_state_callback state_callback;
+  /* User pointer for data & state callbacks*/
   void * user_ptr;
 
   cubeb_resampler * resampler;
   unsigned int inputrate;
-  unsigned int outputrate;
-  unsigned int latency;
+  unsigned int output_configured_rate;
+  unsigned int latency_frames;
   int64_t lastPosition;
   int64_t lastPositionTimeStamp;
   int64_t lastCompensativePosition;
 };
+
+/* Forward declaration. */
+static int opensl_stop_player(cubeb_stream * stm);
+static int opensl_stop_recorder(cubeb_stream * stm);
+
+static int
+opensl_get_draining(cubeb_stream * stm)
+{
+#ifdef DEBUG
+  int r = pthread_mutex_trylock(&stm->mutex);
+  assert(r == EDEADLK && "get_draining deadlock");
+#endif
+  return stm->draining;
+}
+
+static void
+opensl_set_draining(cubeb_stream * stm, int value)
+{
+#ifdef DEBUG
+  int r = pthread_mutex_trylock(&stm->mutex);
+  assert(r == EDEADLK && "set_draining deadlock");
+#endif
+  assert(value == 0 || value == 1);
+  stm->draining = value;
+}
+
+static uint32_t
+opensl_get_shutdown(cubeb_stream * stm)
+{
+#ifdef DEBUG
+  int r = pthread_mutex_trylock(&stm->mutex);
+  assert(r == EDEADLK && "get_shutdown deadlock");
+#endif
+  return stm->shutdown;
+}
+
+static void
+opensl_set_shutdown(cubeb_stream * stm, uint32_t value)
+{
+#ifdef DEBUG
+  int r = pthread_mutex_trylock(&stm->mutex);
+  assert(r == EDEADLK && "set_shutdown deadlock");
+#endif
+  assert(value == 0 || value == 1);
+  stm->shutdown = value;
+}
 
 static void
 play_callback(SLPlayItf caller, void * user_ptr, SLuint32 event)
@@ -89,15 +213,46 @@ play_callback(SLPlayItf caller, void * user_ptr, SLuint32 event)
   switch (event) {
   case SL_PLAYEVENT_HEADATMARKER:
     pthread_mutex_lock(&stm->mutex);
-    draining = stm->draining;
+    draining = opensl_get_draining(stm);
     pthread_mutex_unlock(&stm->mutex);
     if (draining) {
       stm->state_callback(stm, stm->user_ptr, CUBEB_STATE_DRAINED);
-      (*stm->play)->SetPlayState(stm->play, SL_PLAYSTATE_PAUSED);
+      if (stm->play) {
+        int r = opensl_stop_player(stm);
+        assert(r == CUBEB_OK);
+      }
+      if (stm->recorderItf) {
+        int r = opensl_stop_recorder(stm);
+        assert(r == CUBEB_OK);
+      }
     }
     break;
   default:
     break;
+  }
+}
+
+static void
+recorder_marker_callback (SLRecordItf caller, void * pContext, SLuint32 event)
+{
+  cubeb_stream * stm = pContext;
+  assert(stm);
+
+  if (event == SL_RECORDEVENT_HEADATMARKER) {
+    pthread_mutex_lock(&stm->mutex);
+    int draining = opensl_get_draining(stm);
+    pthread_mutex_unlock(&stm->mutex);
+    if (draining) {
+      stm->state_callback(stm, stm->user_ptr, CUBEB_STATE_DRAINED);
+      if (stm->recorderItf) {
+        int r = opensl_stop_recorder(stm);
+        assert(r == CUBEB_OK);
+      }
+      if (stm->play) {
+        int r = opensl_stop_player(stm);
+        assert(r == CUBEB_OK);
+      }
+    }
   }
 }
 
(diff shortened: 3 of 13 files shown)


achronop:
Travis failure seems to be an infra issue. Tests are successful on my account run https://travis-ci.org/achronop/cubeb/builds/136693799

src/cubeb-sles.h
```diff
@@ -0,0 +1,45 @@
+/*
```
kinetiknz: Please make a note to update media/libcubeb/update.sh so that this file is synced from upstream from the next cubeb update (bringing in these changes) onwards?

src/cubeb_array_queue.h
```diff
@@ -0,0 +1,115 @@
...
+    pthread_mutex_t empty_mutex;
+} array_queue;
+
+array_queue * array_queue_create(size_t num)
+{
+    assert(num > 0);
+
+    array_queue * new_queue = (array_queue*)calloc(1, sizeof(array_queue));
+    new_queue->buf = (void **)calloc(1, sizeof(void*) * num);
+    memset(new_queue->buf, 0, sizeof(void*)*num);
```
kinetiknz: calloc() returns a zero-initialized allocation, so the memset is redundant.

src/cubeb_array_queue.h
```diff
@@ -0,0 +1,115 @@
...
+#endif
+
+typedef struct
+{
+    void ** buf;
+    size_t num;
+    uint64_t writePos;
+    uint64_t readPos;
+    pthread_mutex_t mutex;
+    pthread_mutex_t empty_mutex;
```
kinetiknz: It looks like you want a pthread_cond_t instead of a second mutex here.

src/cubeb_array_queue.h
```diff
@@ -0,0 +1,115 @@
...
+    return new_queue;
+}
+
+void array_queue_destroy(array_queue * aq)
+{
+    if(aq)
+    {
+        free(aq->buf);
+        pthread_mutex_destroy(&aq->mutex);
+        pthread_mutex_destroy(&aq->empty_mutex);
```
kinetiknz: It's an error to destroy a locked mutex, which could be induced by a queue_create followed by a queue_destroy.

src/cubeb_array_queue.h
```diff
@@ -0,0 +1,115 @@
...
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+typedef struct
+{
+    void ** buf;
+    size_t num;
+    uint64_t writePos;
+    uint64_t readPos;
```
kinetiknz: I think these should be size_t, and the difference (where you're using "int r" can be ssize_t).

src/cubeb_array_queue.h
```diff
@@ -0,0 +1,115 @@
...
+    pthread_mutex_t mutex;
+    pthread_mutex_t empty_mutex;
+} array_queue;
+
+array_queue * array_queue_create(size_t num)
+{
+    assert(num > 0);
```
kinetiknz: size_t is unsigned, so this is always true.

src/cubeb-sles.h
achronop: This file actually exists in m-c just on different version. Created initially in m-c to re-use the the existing gecko classes that create and realize the engine. Here I simple redirect to the OpenSL ES api methods. The gecko change that introduced the use of the file did not import back the change on github repo. The two files are created to have different version of code.

kinetiknz: Ah, I see.  The old copyright confused me, it should be current.  I see the one in Gecko doesn't even have the appropriate license header... :/

src/cubeb-sles.h
```diff
@@ -0,0 +1,45 @@
...
+                      const SLInterfaceID *   pInterfaceIds,
+                      const SLboolean *       pInterfaceRequired) {
+    SLresult lRes;
+    lRes = slCreateEngine(pEngine,
+                          numOptions,
+                          pEngineOptions,
+                          numInterfaces,
+                          pInterfaceIds,
+                          pInterfaceRequired);
+    return lRes;
```
kinetiknz: Might as well make this a single return without the lRes temp.

src/cubeb-sles.h
```diff
@@ -0,0 +1,45 @@
...
+}
+
+static void cubeb_destroy_sles_engine(SLObjectItf * self) {
+    if (*self != NULL) {
```
kinetiknz: Indentation should match the rest of cubeb (2 spaces) in this file

src/cubeb-sles.h
```diff
@@ -0,0 +1,45 @@
...
+    }
+}
+
+/* Only synchronous operation is supported, as if the second
+   parameter was FALSE. */
+static SLresult cubeb_realize_sles_engine(SLObjectItf self) {
+    SLresult lRes = (*self)->Realize(self,SL_BOOLEAN_FALSE);
+    if (lRes != SL_RESULT_SUCCESS) {
+        return lRes;
+    }
```
kinetiknz: This doesn't return a value in the success case... again, it might as well simply return the result of calling Realize directly without the temp value.

Also space between |self,| and |SL_BOOLEAN_FALSE|

src/cubeb-sles.h
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright © 2013 Mozilla Foundation
```
kinetiknz: 2016 since this is new

src/cubeb_array_queue.h
```diff
@@ -0,0 +1,115 @@
...
+    pthread_mutex_init(&new_queue->mutex, NULL);
+    pthread_mutex_init(&new_queue->empty_mutex, NULL);
+    pthread_mutex_lock(&new_queue->empty_mutex);
+
+    return new_queue;
+}
+
+void array_queue_destroy(array_queue * aq)
+{
+    if(aq)
```
kinetiknz: Is there a reason this needs to handle NULL array_queues? Just assume or assert the caller passes a valid one in.

src/cubeb_array_queue.h
```diff
@@ -0,0 +1,115 @@
...
+        aq->writePos = (aq->writePos + 1) % aq->num;
+        ret = 0;
+    }
+    // else queue is full
+    pthread_mutex_unlock(&aq->empty_mutex);
+    pthread_mutex_unlock(&aq->mutex);
+    return ret;
+}
+
+void* array_queue_get(array_queue * aq)
```
kinetiknz: Call it _pop?

src/cubeb_array_queue.h
```diff
@@ -0,0 +1,115 @@
...
+    pthread_mutex_unlock(&aq->mutex);
+    return value;
+}
+
+void array_queue_wait_if_empty(array_queue * aq)
+{
+    while (pthread_mutex_trylock(&aq->empty_mutex) != 0) {
+        usleep(10);
+    }
+    pthread_mutex_unlock(&aq->empty_mutex);
```
kinetiknz: This can be replaced with waiting on the pthread_cond_t.

src/cubeb_array_queue.h
```diff
@@ -0,0 +1,115 @@
...
+}
+
+int array_queue_get_size(array_queue * aq)
```
kinetiknz: Should return an ssize_t

src/cubeb_array_queue.h
```diff
@@ -0,0 +1,115 @@
...
+
+    array_queue * new_queue = (array_queue*)calloc(1, sizeof(array_queue));
+    new_queue->buf = (void **)calloc(1, sizeof(void*) * num);
+    memset(new_queue->buf, 0, sizeof(void*)*num);
+    new_queue->readPos = 0;
+    new_queue->writePos = 0;
+    new_queue->num = num;
+
+    pthread_mutex_init(&new_queue->mutex, NULL);
+    pthread_mutex_init(&new_queue->empty_mutex, NULL);
```
padenot: Can we set our mutex to be `PTHREAD_MUTEX_ERRORCHECK` in debug ? This will make it easier to track down recursive locking and the like.

src/cubeb-sles.h
padenot: Can we just add `-Wall` in the makefile to catch those ?

kinetiknz: It's already there

src/cubeb_array_queue.h
kinetiknz: We can, but we'd also need to actually check the results of pthread_* calls everywhere to notice the errors.

src/cubeb_array_queue.h
```diff
@@ -0,0 +1,115 @@
...
+    }
+    pthread_mutex_unlock(&aq->mutex);
+    return r;
+}
+
+#if defined(__cplusplus)
+}
+#endif
+
+#endif //CUBE_ARRAY_QUEUE_H
```
padenot: We might want to just write a lock-free ring buffer at this point. Also waiting is scary.

src/cubeb_opensl.c
```diff
@@ -497,72 +822,212 @@ opensl_destroy(cubeb * ctx)
...
+                                                           lSoundRecorderReqs);
+  // Sample rate not supported. Try again with default sample rate!
+  if (res == SL_RESULT_CONTENT_UNSUPPORTED) {
+    if (stm->outputrate != 0 && stm->output_enabled) {
+      // Set the same with the player. Since there is no
+      // api for input device this is a safe choice.
+      stm->input_device_rate = stm->outputrate;
+    } else {
+      // A safe choice for Android.
+      stm->input_device_rate = DEFAULT_SAMPLE_RATE;
```
padenot: http://androidxref.com/6.0.1_r10/xref/frameworks/av/media/libmedia/AudioRecord.cpp#488. You can get this number using `cubeb_get_preferred_sample_rate`.

src/cubeb_opensl.c
```diff
@@ -604,20 +1070,30 @@ opensl_stream_init(cubeb * ctx, cubeb_stream ** stream, char const * stream_name
...
+                                                  req);
   }
 
   // Sample rate not supported? Try again with primary sample rate!
   if (res == SL_RESULT_CONTENT_UNSUPPORTED) {
-    if (opensl_get_preferred_sample_rate(ctx, &preferred_sampling_rate)) {
-      opensl_stream_destroy(stm);
-      return CUBEB_ERROR;
+    if (opensl_get_preferred_sample_rate(stm->context, &preferred_sampling_rate)) {
+      // If fail default is used
```
padenot: Did you hit this ?

src/cubeb_opensl.c
```diff
@@ -633,34 +1109,35 @@ opensl_stream_init(cubeb * ctx, cubeb_stream ** stream, char const * stream_name
...
-                                          CUBEB_RESAMPLER_QUALITY_DEFAULT);
-
-  if (!stm->resampler) {
-    opensl_stream_destroy(stm);
-    return CUBEB_ERROR;
+  // Calculate the capacity of input array
+  stm->queuebuf_capacity = NBUFS;
+  if (stm->output_enabled) {
+    // Full duplex, update capacity to hold 1 sec of data
+    stm->queuebuf_capacity = 1 * stm->outputrate / stm->queuebuf_len;
```
padenot: One second is very very high, isn't it ?

src/cubeb_opensl.c
```diff
@@ -727,46 +1217,300 @@ opensl_stream_init(cubeb * ctx, cubeb_stream ** stream, char const * stream_name
...
+  // and output.
+  cubeb_stream_params input_params;
+  if (input_stream_params) {
+    input_params = *input_stream_params;
+    input_params.rate = stm->input_device_rate;
+  }
+  cubeb_stream_params output_params;
+  if (output_stream_params) {
+    output_params = *output_stream_params;
+    output_params.rate = stm->outputrate;
```
padenot: Can you use the same style for naming here? `stm->outputrate` vs. `stm->input_device_rate`.

src/cubeb_array_queue.h
achronop: It's great idea to create one ring buffer or any other kind of data structure and use it in all back end.

Waiting is meant to be used only for the full duplex thread (which is different from IO threads) in order to wait idle (and not to consume power) when the array is empty. Wait is *not used* in IO threads!

src/cubeb_opensl.c
achronop: `cubeb_get_preferred_sample_rate` provides the preferred sampling rate for the output device. At this point we are configuring the input and I can see 2 possible options. The output is already configured which is the full duplex case and output is not configure which is the input only case. In first case the `cubeb_get_preferred_sample_rate` has already been called for output and the result is stored in `outputrate`. In second case (input only) I set a default value instead of calling this method because I do not have any indication about output. In extreme case output might not even exist and that should not be a reason for cubeb to fail. I know that in 99,999% of cases it is a safe assumption that output device exists but I prefer to have input completed decoupled from output. I was debating a lot (with myself) about this, comments are welcome.

PS. May the day come that ndk will contain proper methods to get preferred sampling rate for output and input.

src/cubeb_opensl.c
achronop: No never, on both device I've tested.

src/cubeb_opensl.c
achronop: That's high indeed but note it does not store the actual data frames but pointers to the buffers with the actual frames. For a typical scenario on 48000 Hz and 40 ms that's number is 50 which is more than the required, when everything operates smoothly, but it is small enough and it covers the edge cases when for example input thread is much faster than the output etc.

src/cubeb_opensl.c
achronop: Sure, but I would prefer to use my naming style. I've tried to go with the minimum changes for the existing code but since you comment on this I will be happy to follow.

src/cubeb_opensl.c
```diff
@@ -727,46 +1217,300 @@ opensl_stream_init(cubeb * ctx, cubeb_stream ** stream, char const * stream_name
...
+                   unsigned int latency,
+                   cubeb_data_callback data_callback, cubeb_state_callback state_callback,
+                   void * user_ptr)
+{
+  cubeb_stream * stm;
+
+  assert(ctx);
+  if (input_device || output_device) {
+    /* Device selection not yet implemented. */
+    return CUBEB_ERROR_DEVICE_UNAVAILABLE;
```
padenot: This will probably prevent audio working in gecko, that passes default input device or the chosen device all the time.

src/cubeb_opensl.c
```diff
@@ -727,46 +1217,300 @@ opensl_stream_init(cubeb * ctx, cubeb_stream ** stream, char const * stream_name
...
+  stm = calloc(1, sizeof(*stm));
+  assert(stm);
+
+  stm->context = ctx;
+  stm->data_callback = data_callback;
+  stm->state_callback = state_callback;
+  stm->user_ptr = user_ptr;
+  stm->latency = latency;
+  stm->input_enabled = (input_stream_params) ? 1 : 0;
+  stm->output_enabled = (output_stream_params) ? 1 : 0;
```
padenot: How about we store the stream params in the struct directly like on windows ?

src/cubeb_opensl.c
```diff
@@ -727,46 +1217,300 @@ opensl_stream_init(cubeb * ctx, cubeb_stream ** stream, char const * stream_name
...
+    return CUBEB_ERROR_DEVICE_UNAVAILABLE;
+  }
+
+  *stream = NULL;
+
+  if ((output_stream_params &&
+       (output_stream_params->channels < 1 || output_stream_params->channels > 32)) ||
+      latency < 1 || latency > 2000) {
+    return CUBEB_ERROR_INVALID_FORMAT;
+  }
```
padenot: Can we validate input stream params here ?

src/cubeb_opensl.c
```diff
@@ -727,46 +1217,300 @@ opensl_stream_init(cubeb * ctx, cubeb_stream ** stream, char const * stream_name
...
+    if (r != CUBEB_OK) {
+      opensl_stream_destroy(stm);
+      return r;
+    }
+  }
+
+  /* Configure resampler*/
+  uint32_t target_sample_rate;
+  if (input_stream_params) {
+    target_sample_rate = input_stream_params->rate;
```
padenot: if there is both input and output, we should assert that the rate are the same.

src/cubeb_opensl.c
```diff
@@ -158,6 +257,232 @@ bufferqueue_callback(SLBufferQueueItf caller, void * user_ptr)
...
+{
+  cubeb_stream * stm = p;
+  assert(stm);
+
+  while (1) {
+    pthread_mutex_lock(&stm->mutex);
+    int draining = stm->draining;
+    uint32_t shutdown = stm->shutdown;
+    pthread_mutex_unlock(&stm->mutex);
+    if (draining || shutdown){
```
padenot: Space between ) and {.

src/cubeb_opensl.c
```diff
@@ -158,6 +257,232 @@ bufferqueue_callback(SLBufferQueueItf caller, void * user_ptr)
...
+      LOG("Exit full duplex thread");
+      return NULL;
+    }
+
+    // Wait until some input exist.
+    array_queue_wait_if_empty(stm->input_queue);
```
padenot: It's better to fill with silence for the first couple silence, to not add delay on the output.

src/cubeb_opensl.c
```diff
@@ -158,6 +257,232 @@ bufferqueue_callback(SLBufferQueueItf caller, void * user_ptr)
...
+    draining = stm->draining;
+    shutdown = stm->shutdown;
+    pthread_mutex_unlock(&stm->mutex);
+    if (draining || shutdown){
+      LOG("Exit full duplex thread after wait");
+      return NULL;
+    }
+
+    // Get buffer to store the output
+    void * output_buffer = stm->queuebuf[stm->queuebuf_idx];
```
padenot: How is this one synchronized ?

src/cubeb_opensl.c
```diff
@@ -67,8 +78,56 @@ struct cubeb_stream {
...
+  SLAndroidSimpleBufferQueueItf recorderBufferQueueItf;
+  /* Store input buffers. */
+  void ** input_buffer_array;
+  /* The capacity of the array.
+   * On capture only can be small (4).
+   * On full duplex is calculated to
+   * store 1 sec of data buffers. */
+  uint32_t input_array_capacity;
+  /* Current filled index of input buffer array. */
+  int input_buffer_index;
```
padenot: Document why this in an `int` and not an `uint32_t`. Probably because we use -1 as a flag.

src/cubeb_opensl.c
```diff
@@ -67,8 +78,56 @@ struct cubeb_stream {
   int draining;
   cubeb_stream_type stream_type;
 
+  array_queue * output_queue;
+
+  /* Flags to determine in/out.*/
+  uint32_t input_enabled;
+  uint32_t output_enabled;
```
padenot: If we decide to store `cubeb_stream_param` structs, we can implement this by checking if the struct is valid.

src/cubeb_opensl.c
```diff
@@ -121,14 +215,19 @@ bufferqueue_callback(SLBufferQueueItf caller, void * user_ptr)
...
-    if (!draining) {
+    if (!draining && !shutdown) {
       written = cubeb_resampler_fill(stm->resampler,
                                      NULL, NULL,
                                      buf, stm->queuebuf_len / stm->framesize);
       if (written < 0 || written * stm->framesize > stm->queuebuf_len) {
-        (*stm->play)->SetPlayState(stm->play, SL_PLAYSTATE_PAUSED);
+        pthread_mutex_lock(&stm->mutex);
+        stm->shutdown = 1;
+        pthread_mutex_unlock(&stm->mutex);
```
padenot: Can you add a getter and a setter for the members that need to be protected by a mutex, and assert that the lock is taken when accessing shared data ?

src/cubeb_opensl.c
```diff
@@ -67,8 +78,56 @@ struct cubeb_stream {
...
+
+  /* On full duplex thread responsible to call
+   * user callback through resampler. */
+  pthread_t full_duplex_thread;
+
+  /* Flag to stop the execution of user callback and
+   * close all working threads.*/
+  uint32_t shutdown;
+
+  /* Store user callback. */
```
padenot: Can you document members that are accessed by multiple threads, and which lock is protecting them ? Better, you can use getters and setters that assert the lock is taken.

src/cubeb_opensl.c
padenot: Yeah but we need this number to get a fast track for input only. The correct rate to use to get a fast track for input only is the primary output rate (which is weird, but reading android's source code, is the valid way to do this).

src/cubeb_opensl.c
achronop: Rates are validated one level up, in `cubeb_stream_init` before get in the back end. I can assert here but since it is a global convention it may be better to assert in cubeb.c file.

src/cubeb_array_queue.h
padenot: We should use event-based scheduling and not waits, if we really want to limit CPU usage, using `pthread_cond_t` functions.

src/cubeb_opensl.c
achronop: Please note this is neither the input nor output thread. This is a 3rd thread created on full duplex case to execute the user call back. It is better to wait here for input before calling the user callback. The output thread is implemented in `player_fullduplex_callback` and silence is pushed on player when there is no output available.

src/cubeb_array_queue.h
achronop: This is also a comment from kinetik, I am currently working on that.

padenot:
Review status: 0 of 4 files reviewed at latest revision, 34 unresolved discussions.

---

*[src/cubeb_opensl.c, line 237 \[r1\]](https://reviewable.io:443/reviews/kinetiknz/cubeb/114#-KKhvEzQP82AriykTCsA:-KKhvEzQP82AriykTCsB:-218767044) ([raw file](https://github.com/kinetiknz/cubeb/blob/3dbead817778a4b911574580933ef4de3910e767/src/cubeb_opensl.c#L237)):*
> ```C
>     // Keep sending silent data even in draining mode to prevent the audio
>     // back-end from being stopped automatically by OpenSL/ES.
>     memset(buf + written * stm->framesize, 0, stm->queuebuf_len - written * stm->framesize);
> ```

Check for underflow in the third argument.

---

*[src/cubeb_opensl.c, line 301 \[r1\]](https://reviewable.io:443/reviews/kinetiknz/cubeb/114#-KKhp2brNRQzu-O4hjer:-KKhp2bsjIY7XYs_elx_:-738312512) ([raw file](https://github.com/kinetiknz/cubeb/blob/3dbead817778a4b911574580933ef4de3910e767/src/cubeb_opensl.c#L301)):*
> ```C
> 
>   if (stm->shutdown || stm->draining) {
>     // Accordint to the doc, on transition to the SL_RECORDSTATE_STOPPED state,
> ```

Can you link to the documentation here? Or maybe it;s on the weird PDF they have, in which case maybe just link to the section or page or something like that.

---

*[src/cubeb_opensl.c, line 369 \[r1\]](https://reviewable.io:443/reviews/kinetiknz/cubeb/114#-KKhpZ5AhZxRfujz8eiK:-KKhpZ5AhZxRfujz8eiL:700025808) ([raw file](https://github.com/kinetiknz/cubeb/blob/3dbead817778a4b911574580933ef4de3910e767/src/cubeb_opensl.c#L369)):*
> ```C
> 
>   assert(stm->input_queue);
>   r = array_queue_push(stm->input_queue, input_buffer);
> ```

So we have effectively two queues: the queue of the recorder, and the array_queue, before the buffer goes to the user, right ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kinetiknz/cubeb/114)*

achronop:
*[src/cubeb_opensl.c, line 369 \[r1\]](https://reviewable.io:443/reviews/kinetiknz/cubeb/114#-KKhpZ5AhZxRfujz8eiK:-KKi8o54HhXpBvzKTQu-:221854384) ([raw file](https://github.com/kinetiknz/cubeb/blob/3dbead817778a4b911574580933ef4de3910e767/src/cubeb_opensl.c#L369)):*
<details><summary><i>Previously, padenot (Paul Adenot) wrote…</i></summary>
> So we have effectively two queues: the queue of the recorder, and the array_queue, before the buffer goes to the user, right ?
</details>
The recorder queue is part of the opensl es implementation and is used when you need to control the input data using registered callbacks. It is more an opensl es element rather than a data structure to manipulate data. The array_queue is the implemented data structure to be used for sharing data between threads in the right order.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kinetiknz/cubeb/114)*

achronop:
*[src/cubeb_array_queue.h, line 31 \[r1\]](https://reviewable.io:443/reviews/kinetiknz/cubeb/114#-KKhofdTCWyR3tbbMnD1-r1-31:-KKrS7rUUUJOlw2CjjfX:-786687564) ([raw file](https://github.com/kinetiknz/cubeb/blob/3dbead817778a4b911574580933ef4de3910e767/src/cubeb_array_queue.h#L31)):*
<details><summary><i>Previously, kinetiknz (Matthew Gregan) wrote…</i></summary>
> size_t is unsigned, so this is always true.
</details>
It's no true when `num == 0`, I will change the assert o reflect that

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kinetiknz/cubeb/114)*

achronop:
*[src/cubeb_array_queue.h, line 49 \[r1\]](https://reviewable.io:443/reviews/kinetiknz/cubeb/114#-KKhofdTCWyR3tbbMnD1-r1-49:-KKrSltXA19kpNTj6sFi:394607497) ([raw file](https://github.com/kinetiknz/cubeb/blob/3dbead817778a4b911574580933ef4de3910e767/src/cubeb_array_queue.h#L49)):*
<details><summary><i>Previously, kinetiknz (Matthew Gregan) wrote…</i></summary>
> Is there a reason this needs to handle NULL array_queues? Just assume or assert the caller passes a valid one in.
</details>
During stream init, if the init process fail we destroy the half created stream and that is a case that might destroy a NULL queue. I moved the check from here to stream_destroy method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kinetiknz/cubeb/114)*

achronop:
*[src/cubeb_opensl.c, line 85 \[r1\]](https://reviewable.io:443/reviews/kinetiknz/cubeb/114#-KKhofdTCWyR3tbbMnD--r1-85:-KKrTrjiFyFbyW9998Ft:1325732867) ([raw file](https://github.com/kinetiknz/cubeb/blob/3dbead817778a4b911574580933ef4de3910e767/src/cubeb_opensl.c#L85)):*
<details><summary><i>Previously, padenot (Paul Adenot) wrote…</i></summary>
> If we decide to store `cubeb_stream_param` structs, we can implement this by checking if the struct is valid.
</details>
I will need to store `cubeb_stream_param` when we define and implement the stream restart behavior due to device changes. I would prefer to leave it as it is for now, since it is tested and working and replace it later when implement this behavior. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kinetiknz/cubeb/114)*

achronop:
*[src/cubeb_opensl.c, line 905 \[r1\]](https://reviewable.io:443/reviews/kinetiknz/cubeb/114#-KKhofdTCWyR3tbbMnD--r1-905:-KKrXiw5EGMzt_3KYiop:-570109431) ([raw file](https://github.com/kinetiknz/cubeb/blob/3dbead817778a4b911574580933ef4de3910e767/src/cubeb_opensl.c#L905)):*
<details><summary><i>Previously, padenot (Paul Adenot) wrote…</i></summary>
> Yeah but we need this number to get a fast track for input only. The correct rate to use to get a fast track for input only is the primary output rate (which is weird, but reading android's source code, is the valid way to do this).
</details>
I will change it to get the primary output rate even on input only scenario. If you have any informative link to provide as explanation for this please let me know. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kinetiknz/cubeb/114)*

achronop:
Review status: 0 of 4 files reviewed at latest revision, 34 unresolved discussions.

---

*[src/cubeb_opensl.c, line 1259 \[r1\]](https://reviewable.io:443/reviews/kinetiknz/cubeb/114#-KKhofdTCWyR3tbbMnD--r1-1259:-KKroqSAClloYLhf07Cl:-1582418045) ([raw file](https://github.com/kinetiknz/cubeb/blob/3dbead817778a4b911574580933ef4de3910e767/src/cubeb_opensl.c#L1259)):*
<details><summary><i>Previously, padenot (Paul Adenot) wrote…</i></summary>
> How about we store the stream params in the struct directly like on windows ?
</details>
See above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kinetiknz/cubeb/114)*

kinetiknz:
The AndroidStudio project needs to be cleaned up before it can be added, there are a lot of generated files and random cruft in there.  The files that are being added also need proper license headers, etc.

src/cubeb_opensl.c
```diff
@@ -58,28 +86,122 @@ struct cubeb_stream {
...
+}
+
+static void
+opensl_set_shutdown(cubeb_stream * stm, uint32_t value)
+{
+  int r = pthread_mutex_trylock(&stm->mutex);
+  assert(r == EBUSY);
+  assert(value == 0 || value == 1);
+  stm->shutdown = value;
+}
```
achronop: @padenot : I think you do not like the way that all these getters/setters are implemented. I created them in order to satisfy one of your comments but I think I have not understood completely what you had proposed. Can you please give me a detailed example to follow?

kinetiknz:
The more I think about it, the more I think we should split the AndroidStudio test project out into a separate repo.

kinetiknz:
#152 would then make integration between the libcubeb repo and the AndroidStudio test project easier, at least according to my quick look at http://tools.android.com/tech-docs/external-c-builds

achronop:
The question by moving the app outside of the repo is how to access the cubeb files. Right now the files inside the app are links to the corresponding `cubeb_*` files on the `src` directory so we are able to modify them from inside the AndroidStudio and commit the changes. 

Also there are other things we need to take care, for example AndroidStudio (right now) does not build any libcubeb.so but link the object directly into the Android project.

src/cubeb_opensl.c
padenot: Those getters are to ensure that the lock has been taken before changing or getting `stm->shutdown` or `stm->draining`, so that the access is synchronized between different threads. If we make a change to the code and forget to lock around one of those getters or setters, the assert will make the program crash.

It can be done like this, or it can be done by setting the mutex to be an error checking mutex, and trying to lock it normally, checking of `EDEADLK` (`man 3 pthread_mutexattr_init`, search for `PTHREAD_MUTEX_ERRORCHECK`).

kinetiknz:
I think we can deal with both of those with a git submodule in the test app repo and using CMake to drive the build.

achronop:
It makes sense to have the app in a separate repo but I would prefer to focus now on merging stuff. If we do not want to let the app reaches the cubeb upstream I can remove it for now. In general I feel that full duplex on android is running late and we should focus on landing.

That said I believe we are close to land. Optimization improves the situation dramatically and I do not see the initial problems. I will provide some numbers and plots. In the mean time I would like to ask you (@padenot and @kinetiknz) to provide all final comments in order to merge it. I will be very happy if we manage to merge it the following days.

src/cubeb_opensl.c
padenot: Agreed.

achronop: Actually the check already exists in
https://github.com/kinetiknz/cubeb/blob/master/src/cubeb.c#L79

kinetiknz:
> I think we should update speex separately

FWIW, this turned out to be easy (and there were changes to resample.c, so it was worth doing), so it's done in 6634f0eb3fc38019bf23ec4e46a3642c9fdabe62.

PR #115, opened 2016-06-15
Bump the stack size for WASAPI rendering threads.

padenot:
In Gecko, `webrtc.org` code we call from the callback is using big arrays
allocated on the stack, and we're running out of stack space. Bumping it by a
factor of two should be enough.


kinetiknz:
Is there a BMO bug for this?

padenot:
Not yet, it's just something we saw in crash-stats.

PR #120, opened 2016-07-01
Set in/out buffer size according to latency

achronop:
Set the size of the input and/or output buffer according to the given latency


padenot:
I tested this with all the headsets that prevented a similar patch to land, back in the days, and it works super nicely.

padenot:
I rebased and merged this manually.

PR #123, opened 2016-07-05
Latency numbers should be in frames

padenot:
We need to have a tight control over the latency numbers to trigger fast paths and such.

### DIFF
--- a/src/cubeb.c
+++ b/src/cubeb.c
@@ -107,7 +107,7 @@ validate_stream_params(cubeb_stream_params * input_stream_params,
 int
 validate_latency(int latency)
 {
-  if (latency < 1 || latency > 2000) {
+  if (latency < 1 || latency > 96000) {
     return CUBEB_ERROR_INVALID_PARAMETER;
   }
   return CUBEB_OK;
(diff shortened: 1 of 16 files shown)


src/cubeb.c
```diff
@@ -107,7 +107,7 @@ validate_stream_params(cubeb_stream_params * input_stream_params,
 int
 validate_latency(int latency)
 {
-  if (latency < 1 || latency > 2000) {
+  if (latency < 1 || latency > 96000) {
```
achronop: Not very important but the signature of this method could change and include the params in order to calculate 2 seconds of latency for every combination.

src/cubeb_pulse.c
```diff
@@ -1056,8 +1056,8 @@ pulse_sink_info_cb(pa_context * context, const pa_sink_info * info,
   devinfo->max_rate = PA_RATE_MAX;
   devinfo->default_rate = info->sample_spec.rate;
 
-  devinfo->latency_lo_ms = 25;
-  devinfo->latency_hi_ms = 400;
+  devinfo->latency_lo = 0;
+  devinfo->latency_hi = 0;
```
achronop: Here the values are hardcoded to 0. Don't we need to report min/max latency any more?

achronop:
Looks good. It might be a good idea to add a new cubeb method that converts frames<-->ms and the other way around.

PR #125, opened 2016-07-12
Don't prebuffer when using PulseAudio.

padenot:
This makes the PulseAudio backend behave like all other backends.

@achronop r?

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -120,6 +120,7 @@ struct cubeb_stream {
   pa_sample_spec input_sample_spec;
   int shutdown;
   float volume;
+  size_t pre_buffered_frames;
 };
 
 const float PULSE_NO_GAIN = -1.0;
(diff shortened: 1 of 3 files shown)


src/cubeb_pulse.c
```diff
@@ -772,6 +787,8 @@ pulse_stream_init(cubeb * context,
     r = stream_update_timing_info(stm);
   }
 
+  WRAP(pa_stream_set_write_callback)(stm->output_stream, stream_write_callback, stm);
```
achronop: You need to check again if output is requested `if (output_stream_params)`

achronop: So write callback will not be called again after that point (stream in ready state) until stream is uncorked.

achronop:
r+ with the output check.

kinetiknz:
Have you testing this with A/V sync? It looks like get_position needs to be adjusted however many bytes we wrote in stream_silence_callback.

padenot:
Yeah that's in the third patch.

padenot:
This is a bad approach and does not really work, I'll do something better.

PR #126, opened 2016-07-13
Audiounit linear buffer

achronop:
Replace input buffer array with linear buffer to solve mismatch device rate scenarios.


src/cubeb_audiounit.cpp
```diff
@@ -1123,8 +1127,14 @@ audiounit_stream_init(cubeb * context,
...
     // calculate the expected number of frames.
-    uint32_t output_buffer_frames = (latency / 1000.0) * output_hw_desc.mSampleRate;
-    LOG("Calculated output number of frames %u for latency %u\n", output_buffer_frames, latency);
+    // Use latency to calculate buffer size
+    uint32_t output_buffer_frames = 0;
+    if (output_hw_desc.mSampleRate == output_stream_params->rate) {
+      output_buffer_frames = latency_frames;
+    } else {
+      output_buffer_frames = (latency_frames * output_hw_desc.mSampleRate) / output_stream_params->rate;
+    }
```
padenot: OSX resample the output for us, so we can simply use the right number of frames here.

src/cubeb_audiounit.cpp
```diff
@@ -881,21 +891,37 @@ audiounit_create_unit(AudioUnit * unit,
...
+  // Pre-buffer silence if needed
+  if (capacity != 1) {
+    size_t silence_size = stream->input_buffer_frames *
+                          stream->input_desc.mChannelsPerFrame;
+    stream->input_linear_buffer->push_silence(silence_size);
+
+    assert(stream->input_linear_buffer->length() == silence_size);
+  }
+
+  stream->input_data = calloc(1, stream->input_buffer_frames *
```
padenot: I'd use `new` here.

src/cubeb_audiounit.cpp
```diff
@@ -225,13 +247,13 @@ audiounit_input_callback(void * user_ptr,
   /* Input only. Call the user callback through resampler.
      Resampler will deliver input buffer in the correct rate. */
   frames = input_frames;
-  input_buffer = ring_array_get_data_buffer(&stm->input_buffer_array);
-  assert(input_buffer && "fetch buffer is null in the input");
   outframes = cubeb_resampler_fill(stm->resampler,
-                                   input_buffer->mData,
+                                   stm->input_linear_buffer->data(),
                                    &frames,
```
padenot: Do we know that `input_linear_buffer` has enough frame here ?

src/cubeb_audiounit.cpp
achronop: Not really, the buffer size is tight with hardware sampling rate so in out device operates on 44100 and we request 48000 output the configured buffer size will be resampled up to a greater number. Unfortunately there is not doc on that it's a try-end-error observation.

src/cubeb_audiounit.cpp
achronop: If render fail we abort (line 237) before reach that point. If it does not fail we assume that we have all requested data since we use the official way to get it. It's the same assumption we have done so far. I can also assert it.

src/cubeb_audiounit.cpp
padenot: Ha well too bad.

src/cubeb_audiounit.cpp
padenot: No that's fine.

src/cubeb_audiounit.cpp
```diff
@@ -99,7 +99,9 @@ struct cubeb_stream {
   pthread_mutex_t mutex;
   /* Hold the input samples in every
    * input callback iteration */
-  ring_array input_buffer_array;
+  auto_array<short> * input_linear_buffer;
```
padenot: This should support floats.

src/cubeb_audiounit.cpp
```diff
@@ -57,7 +57,7 @@ typedef UInt32  AudioFormatFlags;
 #define CUBEB_AUDIOUNIT_SUBTYPE kAudioUnitSubType_HALOutput
 #endif
 
-//#define LOGGING_ENABLED
+#define LOGGING_ENABLED
```
padenot: Re-disable the log.

PR #127, opened 2016-07-17
ALSA-related fixups

i-rinat:
A typo in one of `snd_pcm_open()` call. According to alsa-lib source, `SND_PCM_NO_AUTO_RESAMPLE` is a bit flag, it's tested internally on `pcm->mode`. So it makes more sense to pass it as `mode` parameter, leaving `stream` only for stream type.

Suspend recovery. After suspend, ALSA could return `-ESTRPIPE` error code, which also should be handled. Fortunately, `snd_pcm_recover()` knows how to deal with it. But it doesn't always call `snd_pcm_prepare()` after that, which seems to be required on my hardware. Anyway, it shouldn't hurt to call it one more time.

Event demangling. ALSA API requires calling of `snd_pcm_poll_descriptors_revents()`. Raw `revent` values from `fds` have no meaning as-is.

### DIFF
--- a/src/cubeb_alsa.c
+++ b/src/cubeb_alsa.c
@@ -247,8 +247,9 @@ alsa_refill_stream(cubeb_stream * stm)
   pthread_mutex_lock(&stm->mutex);
 
   avail = snd_pcm_avail_update(stm->pcm);
-  if (avail == -EPIPE) {
+  if (avail < 0) {
     snd_pcm_recover(stm->pcm, avail, 1);
+    snd_pcm_prepare(stm->pcm);
     avail = snd_pcm_avail_update(stm->pcm);
   }
 
@@ -269,6 +270,7 @@ alsa_refill_stream(cubeb_stream * stm)
      a funky state, so recover and try again. */
   if (avail == 0) {
     snd_pcm_recover(stm->pcm, -EPIPE, 1);
+    snd_pcm_prepare(stm->pcm);
     avail = snd_pcm_avail_update(stm->pcm);
     if (avail <= 0) {
       pthread_mutex_unlock(&stm->mutex);
@@ -303,8 +305,9 @@ alsa_refill_stream(cubeb_stream * stm)
       }
     }
     wrote = snd_pcm_writei(stm->pcm, p, got);
-    if (wrote == -EPIPE) {
+    if (wrote < 0) {
       snd_pcm_recover(stm->pcm, wrote, 1);
+      snd_pcm_prepare(stm->pcm);
       wrote = snd_pcm_writei(stm->pcm, p, got);
     }
     assert(wrote >= 0 && wrote == got);
--- a/src/cubeb_alsa.c
+++ b/src/cubeb_alsa.c
@@ -971,7 +971,7 @@ alsa_get_preferred_sample_rate(cubeb * ctx, uint32_t * rate) {
 
   /* get a pcm, disabling resampling, so we get a rate the
    * hardware/dmix/pulse/etc. supports. */
-  r = snd_pcm_open(&pcm, CUBEB_ALSA_PCM_NAME, SND_PCM_STREAM_PLAYBACK | SND_PCM_NO_AUTO_RESAMPLE, 0);
+  r = snd_pcm_open(&pcm, CUBEB_ALSA_PCM_NAME, SND_PCM_STREAM_PLAYBACK, SND_PCM_NO_AUTO_RESAMPLE);
   if (r < 0) {
     return CUBEB_ERROR;
   }
--- a/src/cubeb_alsa.c
+++ b/src/cubeb_alsa.c
@@ -110,20 +110,6 @@ struct cubeb_stream {
 };
 
 static int
-any_revents(struct pollfd * fds, nfds_t nfds)
-{
-  nfds_t i;
-
-  for (i = 0; i < nfds; ++i) {
-    if (fds[i].revents) {
-      return 1;
-    }
-  }
-
-  return 0;
-}
-
-static int
 cmp_timeval(struct timeval * a, struct timeval * b)
 {
   if (a->tv_sec == b->tv_sec) {
@@ -389,7 +375,13 @@ alsa_run(cubeb * ctx)
 
     for (i = 0; i < CUBEB_STREAM_MAX; ++i) {
       stm = ctx->streams[i];
-      if (stm && stm->state == RUNNING && stm->fds && any_revents(stm->fds, stm->nfds)) {
+      if (stm && stm->state == RUNNING && stm->fds) {
+        unsigned short revents = 0;
+
+        snd_pcm_poll_descriptors_revents(stm->pcm, stm->fds, stm->nfds, &revents);
+        if (revents == 0)
+          continue;
+
         alsa_set_stream_state(stm, PROCESSING);
         pthread_mutex_unlock(&ctx->mutex);
         state = alsa_refill_stream(stm);


src/cubeb_alsa.c
```diff
@@ -247,8 +247,9 @@ alsa_refill_stream(cubeb_stream * stm)
   pthread_mutex_lock(&stm->mutex);
 
   avail = snd_pcm_avail_update(stm->pcm);
-  if (avail == -EPIPE) {
+  if (avail < 0) {
     snd_pcm_recover(stm->pcm, avail, 1);
+    snd_pcm_prepare(stm->pcm);
```
kinetiknz: Handling a wider range of errors seems fine here, but I'd like to understand why we need to call snd_pcm_prepare after snd_pcm_recover in each of these places.  As far as I understand, snd_pcm_recover calls snd_pcm_prepare internally for all cases that it needs to... what situation are you seeing where that's not the case?

src/cubeb_alsa.c
```diff
@@ -971,7 +971,7 @@ alsa_get_preferred_sample_rate(cubeb * ctx, uint32_t * rate) {
 
   /* get a pcm, disabling resampling, so we get a rate the
    * hardware/dmix/pulse/etc. supports. */
-  r = snd_pcm_open(&pcm, CUBEB_ALSA_PCM_NAME, SND_PCM_STREAM_PLAYBACK | SND_PCM_NO_AUTO_RESAMPLE, 0);
+  r = snd_pcm_open(&pcm, CUBEB_ALSA_PCM_NAME, SND_PCM_STREAM_PLAYBACK, SND_PCM_NO_AUTO_RESAMPLE);
```
kinetiknz: Nice catch!

src/cubeb_alsa.c
```diff
@@ -389,7 +375,13 @@ alsa_run(cubeb * ctx)
 
     for (i = 0; i < CUBEB_STREAM_MAX; ++i) {
       stm = ctx->streams[i];
-      if (stm && stm->state == RUNNING && stm->fds && any_revents(stm->fds, stm->nfds)) {
+      if (stm && stm->state == RUNNING && stm->fds) {
+        unsigned short revents = 0;
+
+        snd_pcm_poll_descriptors_revents(stm->pcm, stm->fds, stm->nfds, &revents);
```
kinetiknz: This is technically correct, but I'm worried that it'll regress de4940260e197592faaed7d3818e85c6aca56197, where we're forced to ignore the converted revents due to receiving wakeups from poll that results in a converted revent of 0 but also seem to require buffer refill to occur otherwise the devices/streams that behave this way simply don't play at all.

kinetiknz:
Thanks for the patches! I'm happy to merge 12f5e08 as-is, but I need to understand some of the changes in the other commits better before merging them; questions above.

src/cubeb_alsa.c
i-rinat: > I'd like to understand why we need to call snd_pcm_prepare after snd_pcm_recover

I'd like too. But despite documentation saying that `snd_pcm_recover()` should be sufficient, it's not. It seems to be some heuristic knowledge. Perhaps, a bug in a particular audio driver.

If I remove that `snd_pcm_prepare()` after `snd_pcm_recover()`, then after suspend-resume cycle I see `snd_pcm_writei()` returning `-ESTRPIPE` indefinitely on my machine. (With disabled assertion of `wrote >= 0 && wrote == got`, otherwise it just aborts). Playback stops.

Device is on-board audio:
```
00:1b.0 Audio device [0403]: Intel Corporation 7 Series/C210 Series Chipset Family High Definition Audio Controller [8086:1e20] (rev 04)
```

I also seen a message ([link](https://www.linux.org.ru/forum/desktop/12738746), in Russian) about similar issue in Fx47. When that person makes suspend-resume cycle, video playback stops. I can reproduce it too, when ALSA is used. With PulseAudio as a default sound system it works fine. Perhaps, we have similar hardware, I didn't ask.

src/cubeb_alsa.c
i-rinat: This patch is fixing an issue I found when tested suspend-resume recovery. Default frequency on my setup is 48000 Hz. When I tried to play sound file at 44100 Hz I got garbled sound that lasted for some seconds instead of full three minutes.

Here is what happens. `poll()` returns, and `revent` on the `fds` entry is set to `1`, `alsa_refill_stream()` gets called. It tries to determine if there is room available by calling `snd_pcm_avail_update()`. Sometime `avail` is a positive number. Then data callback is called. But sometimes it is just `0`. Then `snd_pcm_recover()` throws away whole buffer and tries to refill. After that another time `poll()` returns with `revent` set to `1`, and another time buffer is disposed. That causes eating data at enormous speeds, and totally breaks synchronization.

There is no sense `revent` value as is. It should be decoded by `snd_pcm_poll_descriptors_revents()` call first. And then it could be, there are no usable events for that stream.

i-rinat: The de4940260e197592faaed7d3818e85c6aca56197 was definitely related to using of `revent`, but I can't think of why it happened. Perhaps that was some peculiarity of an audio driver.

It there any way to reach the person who reported the bug, which was fixed by de4940260e197592faaed7d3818e85c6aca56197? Perhaps, we can ask them to test this patch?

i-rinat: Was that https://bugzilla.mozilla.org/show_bug.cgi?id=1279125 ?
The photo and e-mail look very similar to @DanielFragaBR

kinetiknz: Yeah, that's right.  He has been responsive in the bug, so we can ask him to test.

What we used to do was use any_revent to decide to call refill, then use snd_pcm_poll_descriptors_revents to decide if we'd actually refill the buffer (I don't remember why we have any_revent now), and de49402 removed the snd_pcm_poll_descriptors_revents part of that test specifically because we're getting poll wakeups that snd_pcm_poll_descriptors_revents reports 0 events for... but if we don't refill the stream on those wakeups, we never get playback.  So I'm reasonably sure your change effectively reverts this, but it's also possible that the root cause of that wakeup bug is elsewhere and the fix needs to be made elsewhere.

Here's PortAudio dealing with the same issue: http://music.columbia.edu/pipermail/portaudio/2011-December/013252.html

kinetiknz: So I think the solution is to keep the ignore-revents change I made, but also take the other change I asked Daniel to test in bug 1279125 where we treat avail == 0 as a non-error and avoid taking the snd_pcm_recover path.  That should solve the issue you're seeing without regressing the bugfix.

DanielFragaBR: Ok, I'm here too. What patch exactly do you want me to test? Just provide the patch (or the link) so I can test. Thank you!

kinetiknz:
I merged 12f5e08 and the non-snd_pcm_prepare parts of ba4017a.

Looking at http://www.alsa-project.org/alsa-doc/alsa-lib/pcm.html#pcm_errors, the correct way to handle -ESTRPIPE involves a snd_pcm_resume loop which snd_pcm_recover handles internally.

If we really need the additional snd_pcm_prepare call, we can do that, but it's easier to take the obviously correct changes first and layer on stuff we're less sure about later.

src/cubeb_alsa.c
kinetiknz: > (I don't remember why we have any_revent now)

Digging through history (b54c7c05): in the poll() loop, we didn't used to have access to the PCM to call snd_pcm_poll_descriptors_revents on because the poll() stuff was abstracted out as generic waitables, so we used any_revents to decide if we needed to call the callback for a waitable, which would then decode the revents using snd_pcm_poll_descriptors_revents before proceeding.

kinetiknz: @i-rinat I pushed my propose change:

> here we treat avail == 0 as a non-error and avoid taking the snd_pcm_recover path

as 18efde64d2126f351985bdcfe7b1293c2337df0f.  I'll push a try build for Daniel with the changes of yours I've merged so far plus this change, and another build with your change in 221fbe5 on top of that.

kinetiknz: Daniel confirmed in bug 1279125 that 221fbe5 does regress my fix, so we can't take your change unless we find an alternate fix for the underlying problem.

i-rinat:
Sorry for disappearing. :-(

I've tested again with libcubeb version from the master branch, and now 44100 audio is playing fine, while suspend-resume still stops audio completely. (As expected).

At this moment I see too many data callback calls, each for about 3 to 4 frames. For every poll wakeup alsa_set_stream_state() is called, which in turn calls poll_wake() and causes another wakeup. But I don't know how to fix this without causing that regression again. I need more time to think about it.

i-rinat:
I think, I found what caused resume recovery failure on my machine. I have alsa-lib 1.1.1 with default configuration (i.e. no configuration at all), so dmix is used. And here is what `snd_pcm_resume()` backend code looks like for dmix:
```c
int snd_pcm_direct_resume(snd_pcm_t *pcm)
{
        snd_pcm_direct_t *dmix = pcm->private_data;
        int err;
        
        snd_pcm_direct_semaphore_down(dmix, DIRECT_IPC_SEM_CLIENT);
        err = snd_pcm_resume(dmix->spcm);
        if (err == -ENOSYS) {
                /* FIXME: error handling? */
                snd_pcm_prepare(dmix->spcm);
                snd_pcm_start(dmix->spcm);
                err = 0;
        }
        snd_pcm_direct_semaphore_up(dmix, DIRECT_IPC_SEM_CLIENT);
        return err;
}
```

For some reason `snd_pcm_resume(dmix->spcm)` returns `-ENOSYS`. Perhaps, operation is not implemented in kernel driver at all. Nonetheless, that resets error code to `0`. And while `snd_pcm_prepare()` is called for `dmix->spcm`, it's not called for `dmix` itself. Returned `0` also prevents `snd_pcm_recover()` from calling `snd_pcm_prepare()`.

Latest git version of alsa-lib had changes to `/src/pcm/pcm_direct.c`. Now `snd_pcm_direct_resume()` always returns `-ENOSYS`, and therefore `snd_pcm_prepared()` is also called.

I've tried to run with alsa-lib built from the `master` branch, and resume now works with unchanged version of libcubeb.

kinetiknz:
Thanks for the patches, and thanks for digging in to the resume behaviour and confirming that it's fixed upstream in ALSA!

> At this moment I see too many data callback calls, each for about 3 to 4 frames. For every poll wakeup alsa_set_stream_state() is called, which in turn calls poll_wake() and causes another wakeup. But I don't know how to fix this without causing that regression again. I need more time to think about it.

Hmm, that's something I was worried about when I made the changes for bug 1279125 and it's really unfortunate... I think ultimately we need to find an alternative way to fix that bug, but I don't yet know what we need to do either.

PR #128, opened 2016-07-18
pulse: Stop auto call of write callback during setup

achronop:
Stop the call of write callback during playback connect and before the uncork of the stream.


achronop:
This is not ready yet, I will close it for now and will re-push later

PR #129, opened 2016-07-19
Clamp the latency to [256, platform default] on OSX.

padenot:
OSX allows to go lower, but a number of headsets and other hw don't work too
well with super low latency.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -992,7 +992,7 @@ audiounit_stream_init(cubeb * context,
     }
   } else {
     if (AudioUnitGetProperty(stm->input_unit, kAudioDevicePropertyBufferFrameSize,
-          kAudioUnitScope_Output, 0, &default_frame_count, &size) != 0) {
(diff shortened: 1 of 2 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -974,6 +978,22 @@ audiounit_stream_init(cubeb * context,
   /* Init data members where necessary */
   stm->hw_latency_frames = UINT64_MAX;
 
+  /* Silently clamp the latency down to the platform default, because we
+   * synthetize the clock from the callbacks, and we want the clock to update
+   * often. */
+
+  UInt32 default_frame_count;
+  size = sizeof(default_frame_count);
+  if (AudioUnitGetProperty(stm->output_unit, kAudioDevicePropertyBufferFrameSize,
```
achronop: In case of input only the output_unit is NULL and the stream init will fail here. You need to protect the get property call with something like `if (output_stream_params != NULL)`. In that case you need to get the default buffer size from input unit.

achronop:
r+

PR #134, opened 2016-07-22
Add travis support for OSX

padenot:
This fixes a thing in the OSX backend so it builds fine on Travis, changes a bit the .travis.yml, and disables the test on OSX.

Only five of them pass currently, and I don't know how to disable only a few test using autoconf and automake and all.

@kinetiknz, does that look reasonable?


kinetiknz:
LGTM

PR #136, opened 2016-07-25
Clamp latency inside pre-defined limits

achronop:
Create redefine limit and clamp latency inside. If more than one stream is initialized limits only lower values of latency are allowed.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -62,8 +62,8 @@ typedef UInt32  AudioFormatFlags;
 //#define LOGGING_ENABLED
 #ifdef LOGGING_ENABLED
 #define LOG(...) do {                           \
-    fprintf(stderr, "(line: %d) ", __LINE__);   \
     fprintf(stderr, __VA_ARGS__);               \
+    fprintf(stderr, "(line: %d)\n", __LINE__);  \
   } while(0)
 #else
 #define LOG(...)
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -62,8 +62,8 @@ typedef UInt32  AudioFormatFlags;
 //#define LOGGING_ENABLED
 #ifdef LOGGING_ENABLED
 #define LOG(...) do {                           \
+    fprintf(stderr, "(line: %d) ", __LINE__);   \
     fprintf(stderr, __VA_ARGS__);               \
```
padenot: This remove the \n, so it kind of break logging. Maybe you can put a third printf that only prints a \n.

src/cubeb_audiounit.cpp
```diff
@@ -1014,6 +1015,70 @@ audiounit_destroy_input_linear_buffer(cubeb_stream * stream)
   delete stream->input_linear_buffer;
 }
 
+static uint32_t
+audiounit_clamp_latency(cubeb_stream * stm,
+                              uint32_t latency_frames)
+{
+  // For the 1st stream set anything within safe min-max
+  if (stm->context->active_streams == 1) {
```
padenot: This has to be called after incrementing the stream count. Maybe we can just `assert(stm->context->active_streams != 0)` to catch this.

src/cubeb_audiounit.cpp
achronop: I did it that way in order to print the line and then the message in one line. It looks fine in my tests. I you do not like it I can bring it back.

padenot: But then we have all the logging on one line, right ?

achronop: I see your point with endl, many `LOG(...)` still use endl. They should be removed I guess.

PR #137, opened 2016-07-25
Use auto locks on OSX

padenot:


src/cubeb_audiounit.cpp
```diff
@@ -171,7 +171,7 @@ struct cubeb_stream {
   AudioUnit output_unit;
   /* Sample rate of input device*/
   Float64 input_hw_rate;
-  pthread_mutex_t mutex;
+  owned_critical_section * mutex;
```
achronop: Why you use a pointer here? You can keep the object in the struct and avoid new and delete.

src/cubeb_utils_unix.h
```diff
@@ -0,0 +1,56 @@
...
+
+  ~owned_critical_section()
+  {
+    int r = pthread_mutex_destroy(&mutex);
+    assert(r == 0);
+  }
+
+  void enter()
+  {
+    pthread_mutex_lock(&mutex);
```
achronop: Since we have ERRORCHECK on we can assert here for deadlocks. Copy from doc:
```
A thread attempting to relock this mutex without first unlocking it shall return with an error. A thread attempting to unlock a mutex which another thread has locked shall return with an error. A thread attempting to unlock an unlocked mutex shall return with an error. 
```

src/cubeb_utils_unix.h
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  void leave()
+  {
+    pthread_mutex_unlock(&mutex);
```
achronop: Same as above

achronop:
In general looks good. A couple of proposals. Thanks for taking the time to automate things.

r+

PR #141, opened 2016-08-01
Don't dlopen libmedia for OpenSL

snorp:
This is not allowed on Android N, so require the consumer to pass latency and
preferred sample rate themselves. On Android these can be obtained from
the AudioManager service.

### DIFF
--- a/src/cubeb_opensl.c
+++ b/src/cubeb_opensl.c
@@ -34,8 +34,6 @@ static struct cubeb_ops const opensl_ops;
 struct cubeb {
   struct cubeb_ops const * ops;
   void * lib;
-  void * libmedia;
-  int32_t (* get_output_latency)(uint32_t * latency, int stream_type);
   SLInterfaceID SL_IID_BUFFERQUEUE;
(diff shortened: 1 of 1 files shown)


src/cubeb_opensl.c
```diff
@@ -870,16 +691,13 @@ static struct cubeb_ops const opensl_ops = {
   .init = opensl_init,
   .get_backend_id = opensl_get_backend_id,
   .get_max_channel_count = opensl_get_max_channel_count,
-  .get_min_latency = opensl_get_min_latency,
```
achronop: If you want to remove a method from the cubeb_ops struct you must set it to NULL

src/cubeb_opensl.c
```diff
@@ -870,16 +691,13 @@ static struct cubeb_ops const opensl_ops = {
...
-  .get_preferred_sample_rate = opensl_get_preferred_sample_rate,
   .enumerate_devices = NULL,
```
achronop: The same here and bellow

achronop:
This has been fixed in #412.

PR #142, opened 2016-08-02
Pulse: Stop calling user callback before stream start

achronop:
PulseAudio uses pre buffering which means that fires once write callback during the setup before stream uncork. This is responsible for a huge number of warnings during linux debug testing described by  Bug 1286041. 

This PR solves the issue by hooking the write callback after stream uncork. For this reason the user callback must be called once manually at that point. In order to be called by the same thread as the rest of them it is done through a defer event.

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -73,12 +73,14 @@
   X(pa_stream_set_read_callback)                \
   X(pa_stream_connect_record)                   \
   X(pa_stream_readable_size)                    \
+  X(pa_stream_writable_size)                    \
   X(pa_stream_peek)                             \
   X(pa_stream_drop)                             \
   X(pa_stream_get_buffer_attr)                  \
   X(pa_stream_get_device_name)                  \
   X(pa_context_set_subscribe_callback)          \
   X(pa_context_subscribe)                       \
+  X(pa_mainloop_api_once)                       \
 
 #define MAKE_TYPEDEF(x) static typeof(x) * cubeb_##x;
 LIBPULSE_API_VISIT(MAKE_TYPEDEF);
@@ -120,6 +122,7 @@ struct cubeb_stream {
   pa_sample_spec input_sample_spec;
   int shutdown;
   float volume;
+  int state;
 };
 
 const float PULSE_NO_GAIN = -1.0;
@@ -173,13 +176,20 @@ stream_success_callback(pa_stream * s, int success, void * u)
 }
 
 static void
+stream_state_change_callback(cubeb_stream * stm, cubeb_state s)
+{
+  stm->state = s;
+  stm->state_callback(stm, stm->user_ptr, s);
+}
+
+static void
 stream_drain_callback(pa_mainloop_api * a, pa_time_event * e, struct timeval const * tv, void * u)
 {
   cubeb_stream * stm = u;
   /* there's no pa_rttime_free, so use this instead. */
   a->time_free(stm->drain_timer);
   stm->drain_timer = NULL;
-  stm->state_callback(stm, stm->user_ptr, CUBEB_STATE_DRAINED);
+  stream_state_change_callback(stm, CUBEB_STATE_DRAINED);
 }
 
 static void
@@ -187,7 +197,7 @@ stream_state_callback(pa_stream * s, void * u)
 {
   cubeb_stream * stm = u;
   if (!PA_STREAM_IS_GOOD(WRAP(pa_stream_get_state)(s))) {
-    stm->state_callback(stm, stm->user_ptr, CUBEB_STATE_ERROR);
+    stream_state_change_callback(stm, CUBEB_STATE_ERROR);
   }
   WRAP(pa_threaded_mainloop_signal)(stm->context->mainloop, 0);
 }
@@ -286,7 +296,8 @@ stream_write_callback(pa_stream * s, size_t nbytes, void * u)
 {
   LOG("Output callback to be written buffer size %zd\n", nbytes);
   cubeb_stream * stm = u;
-  if (stm->shutdown) {
+  if (stm->shutdown ||
+      stm->state != CUBEB_STATE_STARTED) {
     return;
   }
 
@@ -424,8 +435,8 @@ stream_cork(cubeb_stream * stm, enum cork_state state)
   WRAP(pa_threaded_mainloop_unlock)(stm->context->mainloop);
 
   if (state & NOTIFY) {
-    stm->state_callback(stm, stm->user_ptr,
-                        state & CORK ? CUBEB_STATE_STOPPED : CUBEB_STATE_STARTED);
+    stream_state_change_callback(stm, state & CORK ? CUBEB_STATE_STOPPED
+                                                   : CUBEB_STATE_STARTED);
   }
 }
 
@@ -719,6 +730,7 @@ pulse_stream_init(cubeb * context,
   stm->state_callback = state_callback;
   stm->user_ptr = user_ptr;
   stm->volume = PULSE_NO_GAIN;
+  stm->state = -1;
 
   WRAP(pa_threaded_mainloop_lock)(stm->context->mainloop);
   if (output_stream_params) {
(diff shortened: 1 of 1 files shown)


achronop:
@kinetiknz since Paul is on PTO can you please have a look?

achronop:
Try run of the patch:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=ca0305be33b0

kinetiknz:
5526dd2 is unrelated (owned_critical_section isn't even used in the pulse backend!) so please split it out.  Also, as I've asked before, please stop making updates to the copy in Gecko's tree directly, it's a recipe for losing changes in situations like this.

I'm also *not* happy that this wrapper has been made recursive in the pthreads implementation ( please don't propagate use of this wrapper until this is resolved) - the correct solution is to remove the use of recursive locks in the audiounit backend.  The Windows backend explicitly prohibits recursion; the behaviour should match across platforms as much as possible, and I do not want this code polluted with recursive locks (un)intentionally.  cc @padenot

kinetiknz:
d1d57d1 needs a reference to the BMO bug that you're fixing and some rationale for why this change is being made added to the commit message.

src/cubeb_pulse.c
```diff
@@ -732,7 +733,9 @@ pulse_stream_init(cubeb * context,
     stm->output_sample_spec = *(WRAP(pa_stream_get_sample_spec)(stm->output_stream));
 
     WRAP(pa_stream_set_state_callback)(stm->output_stream, stream_state_callback, stm);
-    WRAP(pa_stream_set_write_callback)(stm->output_stream, stream_write_callback, stm);
+    // De-register write cb here and register after PA is on ready state
+    // This will avoid calling the callback before stream start
+    WRAP(pa_stream_set_write_callback)(stm->output_stream, NULL, stm);
```
kinetiknz: There should be no callback registered at this point, so why do we need to set it to NULL explicitly?

src/cubeb_pulse.c
```diff
@@ -830,10 +833,37 @@ pulse_stream_destroy(cubeb_stream * stm)
   free(stm);
 }
 
+void
+pulse_defer_event_cb(pa_mainloop_api *a, pa_defer_event *e, void *userdata)
```
kinetiknz: Spaces around *, please follow the existing coding style

src/cubeb_pulse.c
```diff
@@ -830,10 +833,37 @@ pulse_stream_destroy(cubeb_stream * stm)
...
+{
+  cubeb_stream * stm = userdata;
+  size_t writeable_size = WRAP(pa_stream_writable_size)(stm->output_stream);
```
kinetiknz: size_t writable_size

src/cubeb_pulse.c
```diff
@@ -830,10 +833,37 @@ pulse_stream_destroy(cubeb_stream * stm)
...
   stream_cork(stm, UNCORK | NOTIFY);
+
+  /* We hook the write callback here in order to avoid the call of
+   * the cb during the setup (and before the start) of the write stream*/
+  if (stm->output_stream) {
+    /* Register write callback if needed after PA is in ready state. */
+    WRAP(pa_threaded_mainloop_lock)(stm->context->mainloop);
+    WRAP(pa_stream_set_write_callback)(stm->output_stream, stream_write_callback, stm);
+
+    /* On input only case need to manually call user cb once in order to make
```
kinetiknz: You mean output only case?  If we're here, we know stm->output_stream is non-NULL, and you're checking that stm->input_stream *is* NULL.

src/cubeb_pulse.c
```diff
@@ -830,10 +833,37 @@ pulse_stream_destroy(cubeb_stream * stm)
...
 pulse_stream_start(cubeb_stream * stm)
 {
   stream_cork(stm, UNCORK | NOTIFY);
+
+  /* We hook the write callback here in order to avoid the call of
+   * the cb during the setup (and before the start) of the write stream*/
+  if (stm->output_stream) {
+    /* Register write callback if needed after PA is in ready state. */
+    WRAP(pa_threaded_mainloop_lock)(stm->context->mainloop);
+    WRAP(pa_stream_set_write_callback)(stm->output_stream, stream_write_callback, stm);
```
kinetiknz: If we start/stop the stream multiple times, we re-register the callback here without ever having unregistered it.  Since there's only a single callback of each type per stream (AIUI), hopefully it's effectively a no-op.

src/cubeb_pulse.c
```diff
@@ -830,10 +833,37 @@ pulse_stream_destroy(cubeb_stream * stm)
...
+
 static int
 pulse_stream_start(cubeb_stream * stm)
 {
   stream_cork(stm, UNCORK | NOTIFY);
+
+  /* We hook the write callback here in order to avoid the call of
+   * the cb during the setup (and before the start) of the write stream*/
+  if (stm->output_stream) {
+    /* Register write callback if needed after PA is in ready state. */
```
kinetiknz: I'd remove both of these comments, they don't tell you anything that isn't obvious from the code itself.

src/cubeb_pulse.c
```diff
@@ -830,10 +833,37 @@ pulse_stream_destroy(cubeb_stream * stm)
...
+    WRAP(pa_threaded_mainloop_lock)(stm->context->mainloop);
+    WRAP(pa_stream_set_write_callback)(stm->output_stream, stream_write_callback, stm);
+
+    /* On input only case need to manually call user cb once in order to make
+     * things roll. This is done via a defer event in order to execute it from
+     * the same thread as the rest of them*/
+    if (!stm->input_stream) {
+      pa_mainloop_api * api = WRAP(pa_threaded_mainloop_get_api)(stm->context->mainloop);
+      api->defer_new(api, pulse_defer_event_cb, stm);
```
kinetiknz: Is there a reason we can't use pa_mainloop_api_once here?

src/cubeb_pulse.c
```diff
@@ -830,10 +833,37 @@ pulse_stream_destroy(cubeb_stream * stm)
...
+  cubeb_stream * stm = userdata;
+  size_t writeable_size = WRAP(pa_stream_writable_size)(stm->output_stream);
+  trigger_user_callback(stm->output_stream, NULL, writeable_size, stm);
+  a->defer_free(e);
+}
+
 static int
 pulse_stream_start(cubeb_stream * stm)
 {
   stream_cork(stm, UNCORK | NOTIFY);
```
kinetiknz: I'd prefer to uncork *after* we've run the stream_write_callback to reduce the chance of underruns.  Is there a reason we can't do that?

src/cubeb_pulse.c
```diff
@@ -830,10 +833,37 @@ pulse_stream_destroy(cubeb_stream * stm)
...
+  /* We hook the write callback here in order to avoid the call of
+   * the cb during the setup (and before the start) of the write stream*/
+  if (stm->output_stream) {
+    /* Register write callback if needed after PA is in ready state. */
+    WRAP(pa_threaded_mainloop_lock)(stm->context->mainloop);
+    WRAP(pa_stream_set_write_callback)(stm->output_stream, stream_write_callback, stm);
+
+    /* On input only case need to manually call user cb once in order to make
+     * things roll. This is done via a defer event in order to execute it from
+     * the same thread as the rest of them*/
```
kinetiknz: "rest of them" is vague.

src/cubeb_pulse.c
achronop: The only reason I see by testing is that we fire the 1st user callback before the cubeb change state callback and I am not sure what kind of implication this might cause. I would say to leave it as is now and keep an eye for underruns or other performance issues.

achronop:
@kinetiknz: I pushed a new version according to your comments. I've changed a little the way to avoid the pre buffering callback. Let me know what you think.

src/cubeb_pulse.c
```diff
@@ -120,6 +122,7 @@ struct cubeb_stream {
   pa_sample_spec input_sample_spec;
   int shutdown;
   float volume;
+  int state;
```
kinetiknz: Make this a cubeb_state since that's what it holds.

kinetiknz:
Thanks, looks good.  Merged in b42b7597 with state's type changed.

PR #143, opened 2016-08-03
Missing update from gecko

achronop:
This is to update the cubeb repo with a change made directly to m-c. `owned_critical_section` is not intend to be recursive. OSX implementation is not ready for non-recursive stream mutex. The change made directly in m-c in order to fix the broken nightly behavior. A bug will be opened to address all changes required in order to restore the non-recursing behavior.


achronop:
Issues #144 created to address the restore of non-recursive behavior of the wrapper.

kinetiknz:
Thanks for filing an issue!

Issue #144, opened 2016-08-03
Restore non-recursive behavior on pthread implemenation of owned_critical_section


achronop:
Mutex became recursive due to broken functionality on Nightly (Bug 1290625). The initial intention of the wrapper is to be non-recursive. This bug will tract the restore of non recursive functionality. 

There are additional disagreements about the recursive behavior of the wrapper in #142 in comment https://github.com/kinetiknz/cubeb/pull/142#issuecomment-237148357

In order to restore a non recursive behavior audiounit backend needs to remove the recursive dependency when other cubeb methods (`cubeb_stream_set_panning` in this case) are used from inside user callback.

#143 add the recursive behavior and listed here for reference.

kinetiknz:
Fixed by #145.

PR #145, opened 2016-08-04
Restore non-recursive behavior of mutex wrapper

achronop:
The `recursive` attribute has been replaced with `normal` in release or `errorcheck` in debug. The critical memory that caused the deadlock, data member `cubeb_stream::panning`, has been turned into atomic in order to ensure synchronization. Patch fixes issue #144 .


kinetiknz:
Thanks!

PR #146, opened 2016-08-11
audiounit: Use DefaultOutputUnit as the default output device.

kinetiknz:
6b2c610e changed the output unit from kAudioUnitSubType_DefaultOutput to
kAudioUnitSubType_HALOutput because capture is never available on the
DefaultOutputUnit.  For the case where we're doing output only to the default
device, this regressed the automatic device switching when the output device
was changed in the Sound system preferences.  Reverting to the
DefaultOutputUnit for this case restores the previous behaviour.  This
addresses BMO #1278612.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -54,12 +54,6 @@ typedef UInt32  AudioFormatFlags;
 #define AU_OUT_BUS    0
 #define AU_IN_BUS     1
 
-#if TARGET_OS_IPHONE
-#define CUBEB_AUDIOUNIT_SUBTYPE kAudioUnitSubType_RemoteIO
-#else
-#define CUBEB_AUDIOUNIT_SUBTYPE kAudioUnitSubType_HALOutput
-#endif
-
 //#define LOGGING_ENABLED
 #ifdef LOGGING_ENABLED
 #define LOG(...) do {                           \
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -928,37 +932,39 @@ audiounit_create_unit(AudioUnit * unit,
...
-    devid = reinterpret_cast<intptr_t>(device);
-  }
-  int err = AudioUnitSetProperty(*unit, kAudioOutputUnitProperty_CurrentDevice,
-                                 kAudioUnitScope_Global,
-                                 is_input ? AU_IN_BUS : AU_OUT_BUS,
-                                 &devid, sizeof(AudioDeviceID));
-  if (err != noErr) {
-    PRINT_ERROR_CODE("AudioUnitSetProperty/kAudioOutputUnitProperty_CurrentDevice", rv);
-    return CUBEB_ERROR;
+    if (device == NULL) {
```
jesup: This is now inside an "if (device != NULL || is_input)".... what if device == NULL and is_input isn't set?  The entire code block of the patch will be skipped and no SetProperty will be done...

kinetiknz: It is intended, it's invalid to set those properties on the DefaultOutputUnit.

I can make it clearer by introducing something like:

`bool use_defaultoutput = device == NULL && !is_input`

Then use that in the earlier test and !that in this test.

kinetiknz:
jesup r+ed this in bug 1278612 and the updated version via IRC.

Issue #147, opened 2016-08-23
pulse: race condition when draining after stop/destroy


achronop:
This is a problem reported during the landing of Bug 1286041. Check in is back out due to the following asan build error:
https://treeherder.mozilla.org/logviewer.html#?job_id=33402081&repo=mozilla-inbound#L1859

I tend to believe that the error is a race condition not immediately related to the patch. According to the asan logs the `stream_drain_callback()` method is called after the `cubeb_stream_destroy()`. I could not repro with firefox locally neither on try server (https://treeherder.mozilla.org/#/jobs?repo=try&revision=bd10d7af294d).

I believe the problem is that we wait an arbitrary amount of time in the case of drain, here:
https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_pulse.c#L271

If happen to stop and destroy the stream before pulse thread clean up the pending request the error should reproduced (I played with the timer and wait a lot but did not happen).

Since I am not able to repro first I would like to ask if you agree with that conclusion.
In the case you do, the most appropriate solution I see is to check in `stream_stop()` if the `stm->drain_timer` is not NULL and somehow wait to drain first and then stop the stream. 

If you have any other idea or approach please let me know.

kinetiknz:
> n the case you do, the most appropriate solution I see is to check in stream_stop() if the stm->drain_timer is not NULL and somehow wait to drain first and then stop the stream. 

That seems reasonable.

One other issue I noticed with the Gecko (AudioStream) side of the changes for this:  since you set the state to STARTED before calling cubeb_stream_start, it seems like it's possible for the following sequence:

Set mState = STARTED
Enter cubeb_stream_start
Enter pulse_defer_event_cb (on PA thread)
Run data_callback, consume all available data, return < aFrames and trigger draining
Finish draining and run state_callback with CUBEB_STATE_DRAINED
Leave cubeb_stream_start
Overwrite mState == DRAINED with mState == STARTED

kinetiknz:
Note that this sequence wasn't possible before because AudioStream::DataCallback returns early if mState == INITIALIZED.

achronop:
I could solve this by setting the mState value in state change callback method of AudioStream class. Defer event is pushed before the uncork notification so it will change first to STARTED and later to DRAINED even on a early drain scenario.

kinetiknz:
@achronop can we close this now that #148 is merged?

achronop:
Yes, this is fixed by #148

PR #148, opened 2016-08-24
pulse: Wait to drain if stop during draining

achronop:
In case of stop during drain wait for the pa_time_event to timeout before stop the stream. Solves the problem described in issue #147


kinetiknz:
Awesome, thanks!

PR #151, opened 2016-09-09
winmm: Fix bad fallback behaviour and shutdown hang with lost devices

kinetiknz:
Two changes to address BMO# 1274479.

- Change the WinMM backend's context initialization to require a device so that we don't permanently fall back to WinMM when no device is available when attempting to initialize the WASAPI backend.
- Try to detect that the waveout device has been lost in winmm_stream_destroy so we can avoid waiting indefinitely on buffers to become free from a non-existent device.

### DIFF
--- a/src/cubeb_winmm.c
+++ b/src/cubeb_winmm.c
@@ -538,23 +538,32 @@ winmm_stream_init(cubeb * context, cubeb_stream ** stream, char const * stream_n
 static void
 winmm_stream_destroy(cubeb_stream * stm)
 {
-  DWORD r;
   int i;
-  int enqueued;
 
   if (stm->waveout) {
+    MMTIME time;
+    MMRESULT r;
+    int device_valid;
+    int enqueued;
+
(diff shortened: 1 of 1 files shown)


src/cubeb_winmm.c
```diff
@@ -533,23 +538,31 @@ winmm_stream_init(cubeb * context, cubeb_stream ** stream, char const * stream_n
...
     EnterCriticalSection(&stm->lock);
     stm->shutdown = 1;
 
     waveOutReset(stm->waveout);
 
+    /* Don't need this value, we just want the result to detect invalid
+       handle/no device errors than waveOutReset doesn't seem to report. */
+    time.wType = TIME_SAMPLES;
+    r = waveOutGetPosition(stm->waveout, &time, sizeof(time));
+    device_valid == !(r == MMSYSERR_INVALHANDLE || r == MMSYSERR_NODRIVER);
```
padenot: s/==/=/

padenot:
r+ with the comment addressed.

PR #153, opened 2016-09-27
Fix drift in WASAPI when having different output and input sample device rate

padenot:
There was multiple causes:
- First, it's better to `roundf` than to `ceilf`, so that after many calls, the number does not drift one way or the other, and stay more or less stable (this is 44bf58d).
- Then, we would sometimes drop some output data, and this would result in regular glitches on the output side. This also mean that the synchronization between the input buffer and the output buffer would become off quite rapidly (we're talking about dropping 500 or so frames every second here). This is ca8f898. We can then remove the code that tries to limit the number of frames we get from the output `AudioClient`. This code was added in https://github.com/kinetiknz/cubeb/commit/17e3048d0afa1152776fb1867cdb61c49fae69e4, but the real fix was the `if (!input_buffer)` bit, the rest was harmful, but I must have not tested enough.
- Since we now accumulate the input data into the linearized buffer, it's better to check its length when checking for input under-run, this is 3203c8e.
- c6906ba fixes compilation for me on VS2015.

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -186,10 +186,11 @@ static void
 stream_drain_callback(pa_mainloop_api * a, pa_time_event * e, struct timeval const * tv, void * u)
 {
   cubeb_stream * stm = u;
+  stream_state_change_callback(stm, CUBEB_STATE_DRAINED);
   /* there's no pa_rttime_free, so use this instead. */
   a->time_free(stm->drain_timer);
   stm->drain_timer = NULL;
-  stream_state_change_callback(stm, CUBEB_STATE_DRAINED);
+  WRAP(pa_threaded_mainloop_signal)(stm->context->mainloop, 0);
 }
 
 static void
@@ -731,6 +732,7 @@ pulse_stream_init(cubeb * context,
   stm->user_ptr = user_ptr;
   stm->volume = PULSE_NO_GAIN;
   stm->state = -1;
+  assert(stm->shutdown == 0);
 
   WRAP(pa_threaded_mainloop_lock)(stm->context->mainloop);
   if (output_stream_params) {
@@ -853,6 +855,7 @@ pulse_defer_event_cb(pa_mainloop_api * a, void * userdata)
 static int
 pulse_stream_start(cubeb_stream * stm)
 {
+  stm->shutdown = 0;
   stream_cork(stm, UNCORK | NOTIFY);
 
   if (stm->output_stream && !stm->input_stream) {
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -689,7 +689,7 @@ refill_callback_duplex(cubeb_stream * stm)
   // When WASAPI has not filled the input buffer yet, send silence.
   double output_duration = double(output_frames) / stm->output_mix_params.rate;
   double input_duration = double(input_frames) / stm->input_mix_params.rate;
-  if (input_duration < output_duration) {
+  if (stm->linear_input_buffer.length() / stm->input_mix_params.channels < output_duration) {
```
kinetiknz: I don't think this comparison is valid.  `output_duration` is in seconds (as was `input_duration`), but the result of `stm->linear_input_buffer.length() / stm->input_mix_params.channels` will be frames.

src/cubeb_wasapi.cpp
```diff
@@ -1511,6 +1512,8 @@ int setup_wasapi_stream(cubeb_stream * stm)
     if (rv != CUBEB_OK) {
       return rv;
     }
+
+    stm->linear_input_buffer.push_front_silence(stm->latency * 4 * stm->input_stream_params.channels);
```
kinetiknz: Where does `4` come from? 

Also, please add a comment explaining what this is needed for.

src/cubeb_wasapi.cpp
```diff
@@ -585,13 +585,14 @@ bool get_input_buffer(cubeb_stream * stm)
         stm->linear_input_buffer.set_length(stm->linear_input_buffer.length() + packet_size * stm->input_stream_params.channels);
       } else if (should_downmix(stm->input_mix_params, stm->input_stream_params)) {
         bool ok = stm->linear_input_buffer.reserve(stm->linear_input_buffer.length() +
-                                                   packet_size * stm->input_stream_params.channels);
+                                                   packet_size);
```
kinetiknz: Why don't we need to convert `packet_size` from frames to samples here?  Isn't `linear_input_buffer` is allocated in samples?  This change is inconsistent with what the other branches of the if() here are doing for sizing the buffer.  Also, if we're always downmixing to mono, `packet_size * stm->input_stream_params.channels` vs `packet_size` should have no effect anyway.

src/cubeb_wasapi.cpp
```diff
@@ -585,13 +585,14 @@ bool get_input_buffer(cubeb_stream * stm)
...
         assert(ok);
         downmix(reinterpret_cast<float*>(input_packet), packet_size,
                 stm->linear_input_buffer.data() + stm->linear_input_buffer.length(),
                 stm->input_mix_params.channels,
                 stm->input_stream_params.channels);
-        stm->linear_input_buffer.set_length(stm->linear_input_buffer.length() + packet_size * stm->input_stream_params.channels);
+
+        stm->linear_input_buffer.set_length(stm->linear_input_buffer.length() + packet_size);
```
kinetiknz: Ditto above.

src/cubeb_resampler_internal.h
```diff
@@ -8,6 +8,10 @@
 #if !defined(CUBEB_RESAMPLER_INTERNAL)
 #define CUBEB_RESAMPLER_INTERNAL
 
+#ifndef NOMINMAX
```
kinetiknz: I'd prefer to declare this in the problematic cpp file(s) like we do with cubeb_wasapi.cpp rather than in a header, where the symbol can leak further than necessary.

padenot:
hrm, not sure why this got closed.

padenot:
apparently I'm hitting a github bug.

PR #155, opened 2016-09-28
Fix drift in WASAPI when having different output and input sample device rate

padenot:
(Re-opening a PR since I'm apparently hitting a race on github)

This addresses the comments, and is cleaner. I removed the explicit buffer append at start, in favour of letting the emergency padding logic automatically kick in. Usually, we never append input frames, it's mainly useful when we're intentionally blocking thread because we're hitting breakpoints, or at start when the input side has not produced enough buffers yet to feed the resampler.


kinetiknz:
For posterity, the old PR was: https://github.com/kinetiknz/cubeb/pull/153

PR #160, opened 2016-10-07
External logging

padenot:
This implements more or less what was in the issue, but I decided against putting things on the context, and rely on global instead, it's better for ergonomics.

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -575,7 +575,7 @@ int cubeb_register_device_collection_changed(cubeb * context,
                                           invalid pointers, or if level is not
                                           in cubeb_log_level. */
 int cubeb_set_log_callback(cubeb_log_level log_level,
-                           void * log_callback);
+                           cubeb_log_callback log_callback);
 
 #if defined(__cplusplus)
 }
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -574,8 +574,8 @@ int cubeb_register_device_collection_changed(cubeb * context,
     @retval CUBEB_ERROR_INVALID_PARAMETER if either context or log_callback are
                                           invalid pointers, or if level is not
                                           in cubeb_log_level. */
-CUBEB_EXPORT int cubeb_set_log_callback(cubeb_log_level log_level,
-                                        void * log_callback);
+int cubeb_set_log_callback(cubeb_log_level log_level,
+                           void * log_callback);
 
 #if defined(__cplusplus)
 }
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -168,6 +168,13 @@ typedef enum {
  *  across calls. */
 typedef void * cubeb_devid;
 
+/** Level (verbosity) of logging for a particular cubeb context. */
+typedef enum {
+  CUBEB_LOG_DISABLED = 0, /** < Logging disabled */
+  CUBEB_LOG_NORMAL = 1, /**< Logging lifetime operation (creation/destruction). */
+  CUBEB_LOG_VERBOSE = 2, /**< Verbose logging of callbacks, can have performance implications. */
+} cubeb_log_level;
+
 /** Stream format initialization parameters. */
 typedef struct {
   cubeb_sample_format format; /**< Requested sample format.  One of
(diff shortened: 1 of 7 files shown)


padenot:
Hrm, I did this on top of the cmake branch because it's so much nicer to work with.

src/cubeb.c
```diff
@@ -442,7 +445,29 @@ int cubeb_register_device_collection_changed(cubeb * context,
   return context->ops->register_device_collection_changed(context, devtype, callback, user_ptr);
 }
 
-void cubeb_crash()
+int cubeb_set_log_callback(cubeb_log_level log_level,
+                           void * log_callback)
+{
+  if (log_level < CUBEB_LOG_DISABLED ||log_level > CUBEB_LOG_VERBOSE) {
```
kinetiknz: Add a space in `||log_level`

include/cubeb/cubeb.h
```diff
@@ -552,6 +564,19 @@ int cubeb_register_device_collection_changed(cubeb * context,
...
+    @param log_callback A function called with a message when there is
+                        something to log. Passing stdout or stderr log to
+                        this fd instead. Passing NULL allows unregistering a
+                        function.
+    @retval CUBEB_OK in case of success.
+    @retval CUBEB_ERROR_INVALID_PARAMETER if either context or log_callback are
+                                          invalid pointers, or if level is not
+                                          in cubeb_log_level. */
+int cubeb_set_log_callback(cubeb_log_level log_level,
+                           void * log_callback);
```
kinetiknz: Can this be `cubeb_log_callback` rather than `void *`?

src/cubeb.c
```diff
@@ -16,6 +16,9 @@
 
 #define NELEMS(x) ((int) (sizeof(x) / sizeof(x[0])))
 
+cubeb_log_level g_log_level;
+void * g_log_callback;
```
kinetiknz: Can this be `cubeb_log_callback` rather than `void *`?

padenot: Yes. Users that pass a FILE* in will have to cast, though.

src/cubeb_log.h
```diff
@@ -0,0 +1,49 @@
...
+
+#define LOG(...) do {                                \
+  LOG_INTERNAL(CUBEB_LOG_NORMAL, __VA_ARGS__);       \
+} while(0)
+
+#define LOG_INTERNAL(level, ...) do {                                \
+  if (g_log_callback && level >= g_log_level) {                      \
+    if ((FILE *)g_log_callback == stderr ||                          \
+        (FILE *)g_log_callback == stdout) {                          \
+      fprintf((FILE*)g_log_callback, "%s:%d: ", __FILE__, __LINE__); \
```
achronop: Concerning the __FILE__ macro, in OpenSL ES I used the a define to truncate the fullpath and keep the filename. That helped me to avoid the long line in the log. I mention in case you find it useful. The define can be found here:
https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_opensl.c#L33

src/cubeb_log.h
```diff
@@ -0,0 +1,49 @@
...
+} while(0)
+
+#define LOG(...) do {                                \
+  LOG_INTERNAL(CUBEB_LOG_NORMAL, __VA_ARGS__);       \
+} while(0)
+
+#define LOG_INTERNAL(level, ...) do {                                \
+  if (g_log_callback && level >= g_log_level) {                      \
+    if ((FILE *)g_log_callback == stderr ||                          \
+        (FILE *)g_log_callback == stdout) {                          \
```
achronop: Since log callback is a method accepting `const char *` I need to cast in order to set `stdout` or `stderr` as a log callback.  That's not a big problem but it is not very obvious that the method can be used in that way.

src/cubeb_log.h
```diff
@@ -0,0 +1,49 @@
...
+#define LOGV(...) do {                               \
+  LOG_INTERNAL(CUBEB_LOG_VERBOSE, __VA_ARGS__);      \
+} while(0)
+
+#define LOG(...) do {                                \
+  LOG_INTERNAL(CUBEB_LOG_NORMAL, __VA_ARGS__);       \
+} while(0)
+
+#define LOG_INTERNAL(level, ...) do {                                \
+  if (g_log_callback && level >= g_log_level) {                      \
```
achronop: I think we print LOGV in every case. If `g_log_level == NORMAL(1)` and `level == VERBOSE (2)` the `level >= g_log_level` is true so we print verbose log in normal scenario. That is not the intention according to the description and also has performance implications.

src/cubeb_log.h
```diff
@@ -0,0 +1,49 @@
...
+        snprintf(buf, MAX_MSG_SIZE, "%s:%d: ", __FILE__, __LINE__);  \
+      int rv = snprintf(buf + strlen(buf),                           \
+                        MAX_MSG_SIZE - fileandline,                  \
+                        __VA_ARGS__);                                \
+      if (rv >= MAX_MSG_SIZE) {                                      \
+        fprintf(stderr,                                              \
+                "Message truncated (size: %d, max was %d)\n",        \
+                rv, MAX_MSG_SIZE);                                   \
+      }                                                              \
+      ((cubeb_log_callback)g_log_callback)(buf);                     \
```
achronop: If we pass a `FILE*` other than `stdout` or `stderr`, I guess we will crash here.

PR #162, opened 2016-10-11
Bug 1308423 - Don't copy owned_critical_section in cubeb.

CervantesYu:


src/cubeb_audiounit.cpp
```diff
@@ -429,7 +429,7 @@ audiounit_init(cubeb ** context, char const * context_name)
 
   ctx->ops = &audiounit_ops;
 
-  ctx->mutex = owned_critical_section();
+  // Don't initialize ctx->mutex because it's already done in cubeb ctor.
```
achronop: We need the double initialization of critical section here otherwise the audiounit backend will crash. This is because we destroy the already initialized by ctor critical section at line 428 with `PodZero(ctx, 1);`. Despite re-initialization is not ideal we need to do it here to avoid crashing.

src/cubeb_audiounit.cpp
```diff
@@ -1130,7 +1130,8 @@ audiounit_stream_init(cubeb * context,
   stm->state_callback = state_callback;
   stm->user_ptr = user_ptr;
   stm->device_changed_callback = NULL;
-  stm->mutex = owned_critical_section();
+
+  // Don't initialize stm->mutex because it's already done in cubeb_stream ctor.
```
achronop: The same here

src/cubeb_audiounit.cpp
CervantesYu: OK. Then we need to call the dtor before PodZero() ctx, or we'll leak the mutex.

achronop: Well, that would be the ideal even if I haven't noticed any leak about it.

CervantesYu: I think it'd be clearer if we calloc() the memory for cubeb and then placement new cubeb. This saves us the extra destruction and prevents similar bugs when new data members are added in the future.

CervantesYu: Do you mean pushing the change to try? If yes, I can do that.

achronop: Try push is good but I was referring to execute locally the cubeb tests. You have to clone the cubeb github repo, build cubeb and execute the tests. There are instructions in the README file in the repo.

kinetiknz:
Thanks, this looks good.  cubeb_audiounit.cpp was converted from C to C++ a bit hastily and the result is kind of a mess.  Reverting to calloc + manual ctor is cleaner than what we had before.  We should use `free` rather than `delete` to release the memory.  I've fixed that as part of the merge.

kinetiknz:
Fixed in 22557d466eceb6ff6ba70ae30d2dcd87648cde0b.

kinetiknz:
FWIW, I also added restrictions to the Pod* helpers in 58f471095c19b86cf37c48ba19d79cb2e7c02b7b to prevent this type of misuse in the future.

PR #163, opened 2016-10-11
Automatically switch to new devices on OSX

padenot:
It quite similar to what we do on Windows:
- Extract everything that touches the devices from `audiounit_stream_{init, destroy}`, and call them in the device change callback
- On OSX the device change callback runs on the same thread as the audio callback, so we don't have to be as careful as on windows
- `AudioOutputUnitStop` returns when the callback has returned (if it was executing). (Source: <http://lists.apple.com/archives/coreaudio-api/2005/Dec/msg00055.html>)
- We keep using the system default for playback-only stream, maybe we could change this ?

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -511,6 +511,22 @@ audiounit_property_listener_callback(AudioObjectID id, UInt32 address_count,
   cubeb_stream * stm = (cubeb_stream*) user;
 
   LOG("Audio device changed, %d events.", address_count);
+#ifdef LOGGING_ENABLED
+  for (UInt32 i = 0; i < address_count; i++) {
+    printf("%u", i);
+    switch(addresses[i].mSelector) {
+      case kAudioHardwarePropertyDefaultOutputDevice:
+        printf(" mSelector == kAudioHardwarePropertyDefaultOutputDevice\n");
+        break;
+      case kAudioHardwarePropertyDefaultInputDevice:
+        printf(" mSelector == kAudioHardwarePropertyDefaultInputDevice\n");
+        break;
+      case kAudioDevicePropertyDataSource:
+        printf(" mSelector == kAudioHardwarePropertyDataSource\n");
+        break;
+    }
+  }
+#endif // LOGGING_ENABLED
 
   for (UInt32 i = 0; i < address_count; i++) {
     switch(addresses[i].mSelector) {
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -388,6 +388,8 @@ audiounit_output_callback(void * user_ptr,
     if (stm->input_linear_buffer->length() == 0) {
       /* Do nothing, there should be enough pre-buffered data to consume. */
       LOG("Input hole. Requested more input than ouput.");
+      stm->input_linear_buffer->push_silence(stm->input_buffer_frames *
+                                            stm->input_desc.mChannelsPerFrame);
     }
     // The input buffer
     input_buffer = stm->input_linear_buffer->data();
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -510,6 +510,8 @@ audiounit_property_listener_callback(AudioObjectID id, UInt32 address_count,
 {
   cubeb_stream * stm = (cubeb_stream*) user;
 
+  LOG("Audio device changed, %d events.", address_count);
+
   for (UInt32 i = 0; i < address_count; i++) {
     switch(addresses[i].mSelector) {
     case kAudioHardwarePropertyDefaultOutputDevice:
@@ -1407,6 +1409,12 @@ audiounit_stream_init(cubeb * context,
     }
   }
 
+  r = audiounit_install_device_changed_callback(stm);
+  if (r != CUBEB_OK) {
+    LOG("Could not install the device change callback.");
+    return r;
+  }
+
   *stream = stm;
   LOG("Cubeb stream (%p) init successful.", stm);
   return CUBEB_OK;
@@ -1434,7 +1442,10 @@ audiounit_stream_destroy(cubeb_stream * stm)
   cubeb_resampler_destroy(stm->resampler);
 
 #if !TARGET_OS_IPHONE
-  audiounit_uninstall_device_changed_callback(stm);
+  int r = audiounit_uninstall_device_changed_callback(stm);
+  if (r != CUBEB_OK) {
+    LOG("Could not uninstall the device changed callback");
+  }
 #endif
 
   {
@@ -1708,7 +1719,7 @@ int audiounit_stream_device_destroy(cubeb_stream * stream,
 }
 
 int audiounit_stream_register_device_changed_callback(cubeb_stream * stream,
-                                                      cubeb_device_changed_callback  device_changed_callback)
+                                                      cubeb_device_changed_callback device_changed_callback)
 {
   /* Note: second register without unregister first causes 'nope' error.
    * Current implementation requires unregister before register a new cb. */
@@ -1717,16 +1728,8 @@ int audiounit_stream_register_device_changed_callback(cubeb_stream * stream,
   auto_lock lock(stream->mutex);
 
   stream->device_changed_callback = device_changed_callback;
-  int r = CUBEB_OK;
-#if !TARGET_OS_IPHONE
-  if (device_changed_callback) {
-    r = audiounit_install_device_changed_callback(stream);
-  } else {
-    r = audiounit_uninstall_device_changed_callback(stream);
-  }
-#endif
 
-  return r;
+  return CUBEB_OK;
 }
 
 static OSStatus
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1469,11 +1469,9 @@ audiounit_stream_init(cubeb * context,
 }
 
 static void
-audiounit_stream_destroy(cubeb_stream * stm)
+close_audiounit_stream(cubeb_stream * stm)
 {
-  stm->shutdown = 1;
-
-  if (stm->input_unit != NULL) {
+  if (has_input(stm)) {
     AudioOutputUnitStop(stm->input_unit);
     AudioUnitUninitialize(stm->input_unit);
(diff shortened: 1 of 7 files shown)


padenot:
@achronop, can you have a look ?

src/cubeb_audiounit.cpp
```diff
@@ -388,6 +388,8 @@ audiounit_output_callback(void * user_ptr,
     if (stm->input_linear_buffer->length() == 0) {
       /* Do nothing, there should be enough pre-buffered data to consume. */
       LOG("Input hole. Requested more input than ouput.");
+      stm->input_linear_buffer->push_silence(stm->input_buffer_frames *
+                                            stm->input_desc.mChannelsPerFrame);
```
achronop: Why don't you create more initial silence. This will affect all the holes and will reveal a problem that we have put too much effort to solve.

src/cubeb_audiounit.cpp
```diff
@@ -526,6 +570,22 @@ audiounit_property_listener_callback(AudioObjectID id, UInt32 address_count,
...
 
+  {
+    auto_lock lock(stm->mutex);
+    close_audiounit_stream(stm);
+    rv = setup_audiounit_stream(stm);
+    if (rv != CUBEB_OK) {
+      LOG("Could not reopen a stream after switching.");
+    }
+
+    stm->frames_read = 0;
```
achronop: Can you please document that this value is reset in case of switch.

src/cubeb_audiounit.cpp
```diff
@@ -526,6 +570,22 @@ audiounit_property_listener_callback(AudioObjectID id, UInt32 address_count,
...
+
+    if (!stm->shutdown) {
+      auto_unlock lock(stm->mutex);
+      audiounit_stream_start(stm);
+    }
+  }
```
achronop: It would be safer to move the switching code inside the case block of the specific events.

src/cubeb_audiounit.cpp
```diff
@@ -1406,35 +1424,122 @@ audiounit_stream_init(cubeb * context,
...
+   * often. */
+  stm->latency_frames = audiounit_clamp_latency(stm, latency_frames);
+  assert(latency_frames > 0);
+
+  {
+    // It's not critical to lock here, because no other thread has been started
+    // yet, but it allows to assert that the lock has been taken in
+    // `setup_audiounit_stream`.
+    auto_lock lock(stm->mutex);
+    r = setup_audiounit_stream(stm);
```
achronop: You do not do anything with return value here. There is full error handling and logging inside `setup_audiounit_stream` so you do not have to do much but you need to check and return in case of the error otherwise you might end up calling the `audiounit_install_device_changed_callback` (bellow) with an invalid stream.

src/cubeb_audiounit.cpp
```diff
@@ -1406,35 +1424,122 @@ audiounit_stream_init(cubeb * context,
...
 
 static void
-audiounit_stream_destroy(cubeb_stream * stm)
+close_audiounit_stream(cubeb_stream * stm)
 {
-  stm->shutdown = 1;
-
-  if (stm->input_unit != NULL) {
+  // We can't take the lock here, it would deadlock inside the platform code.
+  // AudioOutputUnitStop blocks until the audio callback has returned, though.
```
achronop: I guess this is the case when `close_audiounit_stream` is called within property callback. At least we can assert that the lock is owned by the current thread.

src/cubeb_audiounit.cpp
```diff
@@ -1449,11 +1554,10 @@ audiounit_stream_destroy(cubeb_stream * stm)
...
 {
-  {
-    auto_lock lock(stm->mutex);
-    stm->shutdown = 0;
-    stm->draining = 0;
-  }
+  auto_lock lock(stm->mutex);
+
+  stm->shutdown = 0;
+  stm->draining = 0;
```
achronop: Is the change of the scope here intentional. `AudioOutputUnitStart` method takes some time to complete. In case of duplex stream the output callback may block for a long time until the input get started.

src/cubeb_audiounit.cpp
```diff
@@ -526,6 +570,22 @@ audiounit_property_listener_callback(AudioObjectID id, UInt32 address_count,
     }
   }
 
+  {
+    auto_lock lock(stm->mutex);
+    close_audiounit_stream(stm);
```
achronop: I think we must stop the stream before close. Otherwise we might delete a working stream and have undefined behavior.

src/cubeb_audiounit.cpp
```diff
@@ -526,6 +570,22 @@ audiounit_property_listener_callback(AudioObjectID id, UInt32 address_count,
...
+    rv = setup_audiounit_stream(stm);
+    if (rv != CUBEB_OK) {
+      LOG("Could not reopen a stream after switching.");
+    }
+
+    stm->frames_read = 0;
+
+    if (!stm->shutdown) {
+      auto_unlock lock(stm->mutex);
+      audiounit_stream_start(stm);
```
achronop: I believe it's not good that we call the start directly here because the state callback is also called.

src/cubeb_audiounit.cpp
```diff
@@ -526,6 +570,22 @@ audiounit_property_listener_callback(AudioObjectID id, UInt32 address_count,
...
+    auto_lock lock(stm->mutex);
+    close_audiounit_stream(stm);
+    rv = setup_audiounit_stream(stm);
+    if (rv != CUBEB_OK) {
+      LOG("Could not reopen a stream after switching.");
+    }
+
+    stm->frames_read = 0;
+
+    if (!stm->shutdown) {
```
achronop: We may need to check the `stm->shutdown` before switch started. If you check it after close it may have been set due to closing callbacks.

src/cubeb_audiounit.cpp
achronop: With my usb headset I had to move the `close_audiounit_stream` outside the lock in order to work.

src/cubeb_audiounit.cpp
padenot: Fixed.

src/cubeb_audiounit.cpp
padenot: Fixed, I don't reset it anymore.

src/cubeb_audiounit.cpp
padenot: Fixed.

src/cubeb_audiounit.cpp
padenot: Fixed by using atomic, and refactoring the start and stop methods, as well as `close_audiounit_stream`.

src/cubeb_audiounit.cpp
padenot: Fixed, this does not lock anymore.

src/cubeb_audiounit.cpp
padenot: Fixed, this is now finer grained.

src/cubeb_audiounit.cpp
padenot: Fixed.

src/cubeb_audiounit.cpp
padenot: Fixed as well.

src/cubeb_audiounit.cpp
```diff
@@ -289,7 +328,18 @@ audiounit_input_callback(void * user_ptr,
 
   // Full Duplex. We'll call data_callback in the AudioUnit output callback.
   if (stm->output_unit != NULL) {
-    // User callback will be called by output callback
+    // This happens when we're finally getting a new input callback after having
+    // switched device, we can clear the input buffer now, only keeping the data
+    // we just got.
```
padenot: It is used, but at this point, `switching` is probably not `true` anymore, since the device switch callback has returned. This is now explained in a comment in the `cubeb_stream`.

src/cubeb_audiounit.cpp
```diff
@@ -326,15 +376,15 @@ audiounit_output_callback(void * user_ptr,
...
+  stm->output_callback_in_a_row++;
+
   LOGV("output(%p): buffers %d, size %d, channels %d, frames %d.", stm,
        outBufferList->mNumberBuffers, outBufferList->mBuffers[0].mDataByteSize,
        outBufferList->mBuffers[0].mNumberChannels, output_frames);
 
   long outframes = 0, input_frames = 0;
   void * output_buffer = NULL, * input_buffer = NULL;
 
-  auto_lock lock(stm->mutex);
```
padenot: `input_desc` is modified when the device are not running, they have been stopped.

src/cubeb_audiounit.cpp
```diff
@@ -361,12 +411,17 @@ audiounit_output_callback(void * user_ptr,
...
       stm->input_linear_buffer->push_silence(stm->input_buffer_frames *
-                                            stm->input_desc.mChannelsPerFrame);
+                                             stm->input_desc.mChannelsPerFrame);
     }
     /* Input samples stored previously in input callback. */
     if (stm->input_linear_buffer->length() == 0) {
-      /* Do nothing, there should be enough pre-buffered data to consume. */
       LOG("Input hole. Requested more input than ouput.");
+      // If we're switching devices, we put in a bit of silence in the output to
+      // keep everything flowing.
```
padenot: This is now explained in a comment at the top, in `cubeb_stream`.

src/cubeb_audiounit.cpp
```diff
@@ -289,7 +328,18 @@ audiounit_input_callback(void * user_ptr,
...
   if (stm->output_unit != NULL) {
-    // User callback will be called by output callback
+    // This happens when we're finally getting a new input callback after having
+    // switched device, we can clear the input buffer now, only keeping the data
+    // we just got.
+    if (stm->output_callback_in_a_row > 2) {
+      stm->input_linear_buffer->pop(
+          nullptr,
+          stm->input_linear_buffer->length() -
+          input_frames * stm->input_stream_params.channels);
```
achronop: FWIW I am not able to make crash or fail because of it, so it remains a theory.

achronop:
Closed be commit ddbc367cdf58bdf954feff92035fa4b8aadd84e2

PR #164, opened 2016-10-12
Simplify logging macro and clean up a bit

kinetiknz:
The new LOG*() macros are too complicated for my liking, which I glossed over in my initial review:

- the stdout/stderr casting is extremely hacky and can be easily achieved in the caller-supplied log callback
- introducing variables in the LOG*() macros causes naming clashes in real code (c.f. ec775436921bbe2eb200aac89fafc2284c81ca7f for the first example)
- the macros should be as simple as possible since they're effectively inlined everywhere

There were also a bunch of places where LOG*() was being supplied with a newline, which is now fixed.


padenot:
Better indeed, thanks.

PR #166, opened 2016-10-13
Fallback to a the default device if the current device has been invalidated, when opening a stream

padenot:
The long-term plan is to implement the device change callback and to be able to disable only a side of the stream, but this prevent locking everything up without informing the user.

If we can't open the device because it's been invalidated, we simply retry with the default device.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -1381,34 +1381,49 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
   HRESULT hr;
 
   stm->stream_reset_lock.assert_current_thread_owns();
-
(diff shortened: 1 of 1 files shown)


padenot:
@kinetiknz can you have a look ?

kinetiknz:
Why isn't this handled by the caller? If it's done silently inside cubeb, the caller has no knowledge of this happening and can't prevent it or adjust the UI to suit.

padenot:
The real proposal is at #167, this is just a quick something we can do so we can recover.

kinetiknz:
Okay, we can take this (r+ the code changes) but I don't agree with the policy this implements.  Merging since we plan to address this systematically via #167.

PR #171, opened 2016-10-19
Support multiple channels on Windows

ChunMinChang:
changes:
- Add a `enum` for channel layout
- Add a property in `cubeb_stream_params` to specify the channel layout
- Set `dwChannelMask` by the given channel layout in cubeb_wasapi
- Test channel layout in test_audio

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -435,7 +435,7 @@ CUBEB_EXPORT int cubeb_get_max_channel_count(cubeb * context, uint32_t * max_cha
 
 /** Get the minimal latency value, in frames, that is guaranteed to work
     when creating a stream for the specified sample rate. This is platform,
-    hardware and backend dependant.
+    hardware and backend dependent.
     @param context A pointer to the cubeb context.
     @param params On some backends, the minimum achievable latency depends on
                   the characteristics of the stream.
@@ -449,7 +449,7 @@ CUBEB_EXPORT int cubeb_get_min_latency(cubeb * context,
                                        uint32_t * latency_frames);
 
 /** Get the preferred sample rate for this backend: this is hardware and
-    platform dependant, and can avoid resampling, and/or trigger fastpaths.
+    platform dependent, and can avoid resampling, and/or trigger fastpaths.
     @param context A pointer to the cubeb context.
     @param rate The samplerate (in Hz) the current configuration prefers.
     @retval CUBEB_OK
@@ -458,7 +458,7 @@ CUBEB_EXPORT int cubeb_get_min_latency(cubeb * context,
 CUBEB_EXPORT int cubeb_get_preferred_sample_rate(cubeb * context, uint32_t * rate);
 
 /** Get the preferred layout for this backend: this is hardware and
-    platform dependant.
+    platform dependent.
     @param context A pointer to the cubeb context.
     @param layout The layout of the current speaker configuration.
     @retval CUBEB_OK
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -176,12 +176,71 @@ typedef enum {
   CUBEB_LOG_VERBOSE = 2, /**< Verbose logging of callbacks, can have performance implications. */
 } cubeb_log_level;
 
+/** SMPTE channel layout (also known as wave order)
+ * DUAL-MONO      L   R
+ * DUAL-MONO-LFE  L   R   LFE
+ * MONO           M
+ * MONO-LFE       M   LFE
+ * STEREO         L   R
+ * STEREO-LFE     L   R   LFE
+ * 3F             L   R   C
+ * 3F-LFE         L   R   C    LFE
+ * 2F1            L   R   S
+ * 2F1-LFE        L   R   LFE  S
+ * 3F1            L   R   C    S
+ * 3F1-LFE        L   R   C    LFE S
+ * 2F2            L   R   LS   RS
+ * 2F2-LFE        L   R   LFE  LS   RS
+ * 3F2            L   R   C    LS   RS
+ * 3F2-LFE        L   R   C    LFE  LS   RS
+ * 3F3R-LFE       L   R   C    LFE  RC   LS   RS
+ * 3F4-LFE        L   R   C    LFE  RLS  RRS  LS   RS
+ *
+ * The abbreviation of channel name is defined in following table:
+ * Abbr  Channel name
+ * ---------------------------
+ * M     Mono
+ * L     Left
+ * R     Right
+ * C     Center
+ * LS    Left Surround
+ * RS    Right Surround
+ * RLS   Rear Left Surround
+ * RC    Rear Center
+ * RRS   Rear Right Surround
+ * LFE   Low Frequency Effects
+ */
+
+typedef enum {
+  CUBEB_LAYOUT_UNDEFINED, // Indicate the speaker's layout is undefined.
+  CUBEB_LAYOUT_DUAL_MONO,
+  CUBEB_LAYOUT_DUAL_MONO_LFE,
+  CUBEB_LAYOUT_MONO,
+  CUBEB_LAYOUT_MONO_LFE,
+  CUBEB_LAYOUT_STEREO,
+  CUBEB_LAYOUT_STEREO_LFE,
+  CUBEB_LAYOUT_3F,
+  CUBEB_LAYOUT_3F_LFE,
+  CUBEB_LAYOUT_2F1,
+  CUBEB_LAYOUT_2F1_LFE,
+  CUBEB_LAYOUT_3F1,
+  CUBEB_LAYOUT_3F1_LFE,
+  CUBEB_LAYOUT_2F2,
+  CUBEB_LAYOUT_2F2_LFE,
+  CUBEB_LAYOUT_3F2,
+  CUBEB_LAYOUT_3F2_LFE,
+  CUBEB_LAYOUT_3F3R_LFE,
+  CUBEB_LAYOUT_3F4_LFE,
+  CUBEB_LAYOUT_MAX
+} cubeb_channel_layout;
+
 /** Stream format initialization parameters. */
 typedef struct {
-  cubeb_sample_format format; /**< Requested sample format.  One of
-                                   #cubeb_sample_format. */
-  unsigned int rate;          /**< Requested sample rate.  Valid range is [1000, 192000]. */
-  unsigned int channels;      /**< Requested channel count.  Valid range is [1, 8]. */
+  cubeb_sample_format format;   /**< Requested sample format.  One of
+                                     #cubeb_sample_format. */
+  unsigned int rate;            /**< Requested sample rate.  Valid range is [1000, 192000]. */
+  unsigned int channels;        /**< Requested channel count.  Valid range is [1, 8]. */
+  cubeb_channel_layout layout;  /**< Requested channel layout. This must be consistent with the provided channels. */
 #if defined(__ANDROID__)
   cubeb_stream_type stream_type; /**< Used to map Android audio stream types */
 #endif
@@ -397,6 +456,15 @@ CUBEB_EXPORT int cubeb_get_min_latency(cubeb * context,
     @retval CUBEB_ERROR_INVALID_PARAMETER
     @retval CUBEB_ERROR_NOT_SUPPORTED */
 CUBEB_EXPORT int cubeb_get_preferred_sample_rate(cubeb * context, uint32_t * rate);
+
+/** Get the preferred layout for this backend: this is hardware and
+    platform dependant.
+    @param context A pointer to the cubeb context.
+    @param layout The layout of the current speaker configuration.
+    @retval CUBEB_OK
+    @retval CUBEB_ERROR_INVALID_PARAMETER
+    @retval CUBEB_ERROR_NOT_SUPPORTED */
+CUBEB_EXPORT int cubeb_get_preferred_channel_layout(cubeb * context, cubeb_channel_layout * layout);
 
 /** Destroy an application context. This must be called after all stream have
  *  been destroyed.
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -399,12 +399,94 @@ bool should_upmix(cubeb_stream_params & stream, cubeb_stream_params & mixer)
 
 bool should_downmix(cubeb_stream_params & stream, cubeb_stream_params & mixer)
 {
-  return mixer.channels < stream.channels;
+  return mixer.channels < stream.channels ||
+         (stream.layout == CUBEB_LAYOUT_3F2 &&
+          (mixer.layout == CUBEB_LAYOUT_2F2_LFE ||
+           mixer.layout == CUBEB_LAYOUT_3F1_LFE));
 }
 
 double stream_to_mix_samplerate_ratio(cubeb_stream_params & stream, cubeb_stream_params & mixer)
 {
   return double(stream.rate) / mixer.rate;
+}
+
+/* Convert the channel layout into the corresponding KSAUDIO_CHANNEL_CONFIG.
+   See more: https://msdn.microsoft.com/en-us/library/windows/hardware/ff537083(v=vs.85).aspx */
+#define MASK_DUAL_MONO      (SPEAKER_FRONT_LEFT | SPEAKER_FRONT_RIGHT)
+#define MASK_DUAL_MONO_LFE  (MASK_DUAL_MONO | SPEAKER_LOW_FREQUENCY)
+#define MASK_MONO           (KSAUDIO_SPEAKER_MONO)
+#define MASK_MONO_LFE       (MASK_MONO | SPEAKER_LOW_FREQUENCY)
+#define MASK_STEREO         (KSAUDIO_SPEAKER_STEREO)
+#define MASK_STEREO_LFE     (MASK_STEREO | SPEAKER_LOW_FREQUENCY)
+#define MASK_3F             (MASK_STEREO | SPEAKER_FRONT_CENTER)
+#define MASK_3F_LFE         (MASK_3F | SPEAKER_LOW_FREQUENCY)
+#define MASK_2F1            (MASK_STEREO | SPEAKER_BACK_CENTER)
+#define MASK_2F1_LFE        (MASK_2F1 | SPEAKER_LOW_FREQUENCY)
+#define MASK_3F1            (KSAUDIO_SPEAKER_SURROUND)
+#define MASK_3F1_LFE        (MASK_3F1 | SPEAKER_LOW_FREQUENCY)
+#define MASK_2F2            (MASK_STEREO | SPEAKER_SIDE_LEFT | SPEAKER_SIDE_RIGHT)
+#define MASK_2F2_LFE        (MASK_2F2 | SPEAKER_LOW_FREQUENCY)
+#define MASK_3F2            (MASK_3F | SPEAKER_SIDE_LEFT | SPEAKER_SIDE_RIGHT)
+#define MASK_3F2_LFE        (KSAUDIO_SPEAKER_5POINT1_SURROUND)
+#define MASK_3F3R_LFE       (MASK_3F2_LFE | SPEAKER_BACK_CENTER)
+#define MASK_3F4_LFE        (KSAUDIO_SPEAKER_7POINT1_SURROUND)
+
+static DWORD
+channel_layout_to_mask(cubeb_channel_layout layout)
+{
+  XASSERT(layout > CUBEB_LAYOUT_UNDEFINED && layout < CUBEB_LAYOUT_MAX &&
+          "This mask conversion is not allowed.");
+
+  // This variable may be used for multiple times, so we should avoid to
+  // allocate it in stack, or it will be created and removed repeatedly.
+  // Use static to allocate this local variable in data space instead of stack.
+  static DWORD map[CUBEB_LAYOUT_MAX] = {
+    0,                    // CUBEB_LAYOUT_UNDEFINED (this won't be used.)
+    MASK_DUAL_MONO,       // CUBEB_LAYOUT_DUAL_MONO
+    MASK_DUAL_MONO_LFE,   // CUBEB_LAYOUT_DUAL_MONO_LFE
+    MASK_MONO,            // CUBEB_LAYOUT_MONO
+    MASK_MONO_LFE,        // CUBEB_LAYOUT_MONO_LFE
+    MASK_STEREO,          // CUBEB_LAYOUT_STEREO
+    MASK_STEREO_LFE,      // CUBEB_LAYOUT_STEREO_LFE
+    MASK_3F,              // CUBEB_LAYOUT_3F
+    MASK_3F_LFE,          // CUBEB_LAYOUT_3F_LFE
+    MASK_2F1,             // CUBEB_LAYOUT_2F1
+    MASK_2F1_LFE,         // CUBEB_LAYOUT_2F1_LFE
+    MASK_3F1,             // CUBEB_LAYOUT_3F1
+    MASK_3F1_LFE,         // CUBEB_LAYOUT_3F1_LFE
+    MASK_2F2,             // CUBEB_LAYOUT_2F2
+    MASK_2F2_LFE,         // CUBEB_LAYOUT_2F2_LFE
+    MASK_3F2,             // CUBEB_LAYOUT_3F2
+    MASK_3F2_LFE,         // CUBEB_LAYOUT_3F2_LFE
+    MASK_3F3R_LFE,        // CUBEB_LAYOUT_3F3R_LFE
+    MASK_3F4_LFE,         // CUBEB_LAYOUT_3F4_LFE
+  };
+  return map[layout];
+}
+
+cubeb_channel_layout
+mask_to_channel_layout(DWORD mask)
+{
+  switch (mask) {
+    // MASK_DUAL_MONO(_LFE) is same as STEREO(_LFE), so we skip it.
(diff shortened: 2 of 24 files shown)


include/cubeb/cubeb.h
```diff
@@ -176,12 +176,43 @@ typedef enum {
   CUBEB_LOG_VERBOSE = 2, /**< Verbose logging of callbacks, can have performance implications. */
 } cubeb_log_level;
 
+/* Channel layout for a `cubeb_stream`. */
+/* See more: https://en.wikipedia.org/wiki/Surround_sound */
+typedef enum {
+  CUBEB_CHANNEL_LAYOUT_NONE = 0,
```
ChunMinChang: ok, I could update it up to 8 channels and find the references for the channel orders.

I plan to replace ```LAYOUT_NONE``` with ```LAYOUT_DEFAULT```, which implies that the layout will use the default settings of the speaker.

include/cubeb/cubeb.h
```diff
@@ -176,12 +176,43 @@ typedef enum {
...
+  CUBEB_CHANNEL_LAYOUT_3_0 = 4,       // Front L, Front R, Front C
+  CUBEB_CHANNEL_LAYOUT_3_1 = 5,       // Front L, Front R, Front C, LFE
+  CUBEB_CHANNEL_LAYOUT_4_0 = 6,       // Front L, Front R, Front C, Back C
+  CUBEB_CHANNEL_LAYOUT_QUAD = 7,      // Front L, Front R, Back L, Back R
+  CUBEB_CHANNEL_LAYOUT_5_0 = 8,       // Front L, Front R, Front C, Back L, Back R
+  CUBEB_CHANNEL_LAYOUT_5_1 = 9        // Front L, Front R, Front C, LFE, Back L, Back R
+} cubeb_channel_layout;
+
+/* Get channel counts from layout */
+static const int cubeb_layout_to_channels[10] =
```
ChunMinChang: It's used to check whether the channels counts of the layout is same as the pre-defined number of channels. It's used in ```cubeb_guess_channel_layout```, which would be called if the channels counts is different. However, I plan to remove them in next version.

ChunMinChang:
Hi Matthew, 
it's really thanks for your suggestions. 
I was wondering if you could explain more about the "capture streams" mentioned in the comments. Is it something I should check  when the the device is added or removed?

kinetiknz:
> I was wondering if you could explain more about the "capture streams" mentioned in the comments. 

cubeb_stream_init can create a stream for playback (output), capture (input from microphone), or full-duplex (both playback and mic input) depending on the set of input_stream_params and output_stream_params are passed to cubeb_stream_init.

ChunMinChang:
Oh, I see. Thanks for the explanation. So, I guess I also need to set the ```layout``` values to the parameters for the input stream.

padenot:
> @padenot does WebAudio (or WebRTC) have (now or foreseeable) requirements that aren't satisfied by this API?

I think that's fine. For now, Web Audio, the following ordering is specced:
```
    Mono
      0: M: mono

    Stereo
      0: L: left
      1: R: right
    

  Quad
      0: L:  left
      1: R:  right
      2: SL: surround left
      3: SR: surround right

    5.1
      0: L:   left
      1: R:   right
      2: C:   center
      3: LFE: subwoofer
      4: SL:  surround left
      5: SR:  surround right
```

This is effectively WaveEx or SMPTE, so anything works. Maybe we'll add more ordering later, but at this point I'll push to have something that respects a particular spec instead of rolling our own downmix and upmix matrix.

For capture, we want to support at least mono and stereo, I don't know of any standard ordering for input.

ChunMinChang:
Thanks for the hints.  
> @kinetiknz It'd be good to be explicit about which "standard" channel ordering this cubeb API exposes. It looks like WaveEx or SMPTE. 

I was wondering if there are more information about the WaveEx and SMPTE because I am not familiar with these stuff.

After searching, I know SMPTE is [Society of Motion Picture & Television Engineers](https://en.wikipedia.org/wiki/Society_of_Motion_Picture_and_Television_Engineers) and it defines a set of channel order. The channel order of SMPTE could be found on [multimedia wiki](https://wiki.multimedia.cx/index.php?title=Channel_order) ,[avid](http://avid.force.com/pkb/articles/en_US/faq/SMPTE-file-FAQ) and [protools website](http://www.protoolsisawesome.com/?p=575). However, the former conflicts with the latter two. The best way to find out the answer is to check its [standard](https://www.smpte.org/digital-library), but I have no IEEE accounts to access it. Is there other ways?

On the other hand, the WaveEX is totally a mystery to me. Is WaveEX a [TAPIEx software](http://www.tapiex.com/WaveEx.Net.htm)? or it stands for the WAV file? If it's latter, the channel orders is defined [here](https://msdn.microsoft.com/en-us/library/windows/hardware/dn653308(v=vs.85).aspx#Default_Channel_Ordering).

AFAIK, there is no exact definition about the left/right surround for audio 5.1 in [ITU-R BS.775](https://www.itu.int/dms_pubrec/itu-r/rec/bs/R-REC-BS.775-3-201208-I!!PDF-E.pdf) and [ITU-R BS.2159](http://www.itu.int/dms_pub/itu-r/opb/rep/R-REP-BS.2159-7-2015-PDF-E.pdf). Left/Right surround could be __back__ left/right or __side__ left/right. I believe that's why Windows provide two [channel configs](https://msdn.microsoft.com/en-us/library/windows/hardware/ff537083.aspx) for audio 5.1.

The chromium project is the best reference for me now. The channel layout and order are defined exactly in [channel_layout.h](https://cs.chromium.org/chromium/src/media/base/channel_layout.h?l=14) and [channel_layout.cc](https://cs.chromium.org/chromium/src/media/base/channel_layout.cc?l=55). Thus, the current channel layout and order are based on its definition.

> @padenot For capture, we want to support at least mono and stereo, I don't know of any standard ordering for input.

To support multichannel recording, it might need to replace ```stm->output_client``` with ```audio_client```, which is the parameter of ```setup_wasapi_stream_one_side```, in ```handle_channel_layout``` to use the [IAudioClient::IsFormatSupported](https://msdn.microsoft.com/zh-tw/library/windows/desktop/dd370876(v=vs.85).aspx). Nevertheless, I need to buy more equipment to set up an environment that can support multitrack recording. Does it need to be implemented in this bug? or I could open another bug to track it?

kinetiknz:
> After searching, I know SMPTE is Society of Motion Picture & Television Engineers and it defines a set of channel order. The channel order of SMPTE could be found on multimedia wiki ,avid and protools website. However, the former conflicts with the latter two. The best way to find out the answer is to check its standard, but I have no IEEE accounts to access it. Is there other ways?

Not sure.  That's why I referred to the ITU specs, which seem to be freely available.  It'd be good to at least confirm the ITU and SMPTE stuff is aligned, though.

> On the other hand, the WaveEX is totally a mystery to me. Is WaveEX a TAPIEx software? or it stands for the WAV file? If it's latter, the channel orders is defined here.

WaveEX is a common(-ish) but non-standard nickname referring to WAVE_FORMAT_EXTENSIBLE, which is what you have linked to, but seems pretty vaguely defined beyond what's in that link in terms of the other aspects mentioned (up/down mixing, etc.)

> The chromium project is the best reference for me now. The channel layout and order are defined exactly in channel_layout.h and channel_layout.cc. The layout used in chromium is same as FFmpeg configuration in our gecko. Thus, the current channel layout of this patch is based on its definition.

Okay.  It'd be good to refer to that somewhere in the changes so it's easier to verify or extend in the future, as I mentioned earlier.

> To support multichannel recording, it might need to replace stm->output_client with audio_client, which is the parameter of setup_wasapi_stream_one_side, in handle_channel_layout to use the IAudioClient::IsFormatSupported. Nevertheless, I need to buy more equipment to set up an environment that can support multitrack recording. Does it need to be implemented in this bug? or I could open another bug to track it?

As I said before, I don't believe there's any need to support anything beyond mono and stereo for capture.  All that needs to be done is specify and document the behaviour for capture streams that the cubeb API will support.

ChunMinChang:
Thanks for the explanation.

The [table 5 in [0]][AESTD1001] indicates the channel assignment of SMPTE from 1 to 6 is L, R, C, LFE, LS, LR, which is similar to the channel maps from the [table 6 in [1]][BS2159]: FL, FR, FC, LFE, BL, BR. Thus, I currently use it as the channel order that are commented above the channel layout in the code. Also, all the references are documented in the comments, so it could be referred in the future.

For capture stream, it stays the same as before. If there are more than two channels, the ```XASSERT(false)``` will be triggered. The behavior of multi-channels are documented in the declaration of the channel layout enum in the cubeb.h.

[[0]: Multichannel surroundsound systems andoperations][AESTD1001]
[[1]: ITU-R BS.2159][BS2159]

[AESTD1001]: http://www.aes.org/technical/documents/AESTD1001.pdf  "AESTD1001"
[BS2159]: http://www.itu.int/dms_pub/itu-r/opb/rep/R-REP-BS.2159-7-2015-PDF-E.pdf "ITU-R BS.2159"

jyavenard:
https://dxr.mozilla.org/mozilla-central/source/dom/media/MediaInfo.cpp#18

is the SMPTE order all media decoders are using.

Note that we have defined our own layout structure, which is pretty much a vector containing the channels in the order you want.

ChunMinChang:
Thanks for the information, jya!

kinetiknz:
Thanks, this is looking good so far.

I don't see any changes to the upmix/downmix code, what's the plan for dealing with that?  The code that's in cubeb right now isn't correct for most channel configurations and exists mostly to support switching devices (e.g. headphones unplugged -> switch to speakers) without completely failing due to mismatched device formats.

The current default behaviour is that when the stream is initialized for the default device (e.g. with a null devid), the stream will automatically follow the OS's default device if it changes.  Because the format required by the default device can change when the device changes, we need to be able to handle resampling and upmixing/downmixing.

Right now, we're mostly avoiding the problem in Gecko by downmixing everything to stereo in AudioConverter, but once we allow different channel configurations within cubeb, we need the upmix/downmix code in cubeb to do the right thing.  We have duplicate code for upmixing/downmixing, and resampling within cubeb and AudioConverter.  The code in cubeb is, as mentioned above, required for the current device switching behaviour we have.

This interacts with the device change handling discussed in issue 167.  I think long term we're better off to avoid duplicate code and have this behaviour all handled in a single place.

kinetiknz:
Actually, it's probably better to discuss the big picture plan separately from this pull request, but I do think we need an overall plan for each of the parts before attacking each individual piece.  I've opened issue #178 for the big picture plan.  Let's keep this PR scope narrow and just cover the API changes and the WASAPI backend.

kinetiknz:
Sorry, trying to catch up on this.  Are you ready for review/merge of the current work in this branch?  Please let me know if you're waiting on me and what you need from me.  Thanks!

ChunMinChang:
@kinetiknz Thanks for the concern! It will be great if you can give me some feedback about current implementation. The following is the changes:
- Add layout setting to cubeb_stream_params
  - The downmix/upmix will be applied if the input layout is different from the output layout
-  Apply layout setting in tests
-  Separate downmix amd upmix into an independent module and add the test for mixing

I am surveying a way to test the multi-channels support. Is it possible to fake a virtual audio device and use this device to intercept and verify the output data?

src/cubeb_mixing.cpp
```diff
@@ -0,0 +1,323 @@
...
+  }
+
+  unsigned int in_channels = CUBEB_CHANNEL_LAYOUT_MAPS[in_layout].channels;
+  unsigned int out_channels = CUBEB_CHANNEL_LAYOUT_MAPS[out_layout].channels;
+
+  // Conversion from 3F2 to 2F2-LFE or 3F1-LFE is allowed, so we use '<=' instead of '<'.
+  assert(out_channels <= in_channels);
+
+  long out_index = 0;
+  unsigned int layout_index = out_layout - CUBEB_LAYOUT_MONO; // The matrix is started from mono.
```
kinetiknz: Rather than calculate the index and then use it later, it'd be clearer to take a reference to the appropriate matrix here and then use it directly, e.g. `auto & downmix_matrix = DOWNMIX_MATRIX_3F2_LFE[out_layout - CUBEB_LAYOUT_MONO];`

src/cubeb_mixing.cpp
```diff
@@ -0,0 +1,323 @@
...
+        out[out_index + j] += DOWNMIX_MATRIX_3F2_LFE[layout_index][j][k] * data;
+      }
+    }
+    out_index += out_channels;
+  }
+
+  return true;
+}
+
+/* Map the audio channel data by layouts*/
```
kinetiknz: Space between layouts and */

src/cubeb_wasapi.cpp
```diff
@@ -392,19 +393,88 @@ bool has_output(cubeb_stream * stm)
...
 {
-  return mixer.channels < stream.channels;
+  XASSERT(layout > CUBEB_LAYOUT_UNSUPPORTED && layout < CUBEB_LAYOUT_MAX &&
+          "This mask conversion is not allowed.");
+
+  // This variable may be used for multiple times, so we should avoid to
+  // allocate it in stack, or it will be created and removed repeatedly.
+  // Use static to allocate this local variable in data space instead of stack.
+  static DWORD map[CUBEB_LAYOUT_MAX] = {
+    0,                    // CUBEB_LAYOUT_UNSUPPORTED(this won't be used.)
```
kinetiknz: Space between UNSUPPORTED and (

src/cubeb_wasapi.cpp
```diff
@@ -392,19 +393,88 @@ bool has_output(cubeb_stream * stm)
...
+  };
+  return map[layout];
 }
 
-double stream_to_mix_samplerate_ratio(cubeb_stream_params & stream, cubeb_stream_params & mixer)
+cubeb_channel_layout
+mask_to_channel_layout(DWORD mask)
 {
-  return double(stream.rate) / mixer.rate;
+  switch(mask) {
```
kinetiknz: Space between switch and (

src/cubeb_wasapi.cpp
```diff
@@ -438,72 +508,15 @@ frames_to_hns(cubeb_stream * stm, uint32_t frames)
...
-}
-
 /* This returns the size of a frame in the stream, before the eventual upmix
    occurs. */
 static size_t
 frames_to_bytes_before_mix(cubeb_stream * stm, size_t frames)
 {
-  size_t stream_frame_size = stm->output_stream_params.channels * sizeof(float);
+  // This is called only when we has a output client.
+  XASSERT(stm->output_client); // or use XASSERT(has_output(stm))?
```
kinetiknz: has_output is probably slightly better/clearer

src/cubeb_wasapi.cpp
```diff
@@ -438,72 +508,15 @@ frames_to_hns(cubeb_stream * stm, uint32_t frames)
...
+  // The output_mix_params.channels is the actual number of output channels.
+  size_t stream_frame_size = stm->output_mix_params.channels * sizeof(float);
```
kinetiknz: Why'd this change from output_stream_params to output_mix_params?

If this change is correct, then the name of the function needs to change, I think.

src/cubeb_wasapi.cpp
```diff
@@ -518,8 +531,8 @@ refill(cubeb_stream * stm, float * input_buffer, long input_frames_count,
      avoid a copy. */
   float * dest = nullptr;
   if (has_output(stm)) {
-    if (should_upmix(stm->output_stream_params, stm->output_mix_params) ||
-        should_downmix(stm->output_stream_params, stm->output_mix_params)) {
+    if (cubeb_should_upmix(&(stm->output_stream_params), &(stm->output_mix_params)) ||
```
kinetiknz: Don't need extra parens here.

src/cubeb_wasapi.cpp
```diff
@@ -518,8 +531,8 @@ refill(cubeb_stream * stm, float * input_buffer, long input_frames_count,
...
+        cubeb_should_downmix(&(stm->output_stream_params), &(stm->output_mix_params))) {
```
kinetiknz: Ditto.

src/cubeb_wasapi.cpp
```diff
@@ -550,12 +563,14 @@ refill(cubeb_stream * stm, float * input_buffer, long input_frames_count,
   XASSERT(out_frames == output_frames_needed || stm->draining || !has_output(stm));
 
   if (has_output(stm)) {
-    if (should_upmix(stm->output_stream_params, stm->output_mix_params)) {
-      upmix(dest, out_frames, output_buffer,
+    if (cubeb_should_upmix(&(stm->output_stream_params), &(stm->output_mix_params))) {
```
kinetiknz: Ditto.

src/cubeb_wasapi.cpp
```diff
@@ -550,12 +563,14 @@ refill(cubeb_stream * stm, float * input_buffer, long input_frames_count,
...
+      cubeb_upmix_float(dest, out_frames, output_buffer,
             stm->output_stream_params.channels, stm->output_mix_params.channels);
-    } else if (should_downmix(stm->output_stream_params, stm->output_mix_params)) {
-      downmix(dest, out_frames, output_buffer,
-              stm->output_stream_params.channels, stm->output_mix_params.channels);
+    } else if (cubeb_should_downmix(&(stm->output_stream_params), &(stm->output_mix_params))) {
+      cubeb_downmix_float(dest, out_frames, output_buffer,
+              stm->output_stream_params.channels, stm->output_mix_params.channels,
+              stm->output_stream_params.layout,
+              stm->output_mix_params.layout);
```
kinetiknz: Indent these to align with `dest`

src/cubeb_wasapi.cpp
```diff
@@ -614,23 +629,25 @@ bool get_input_buffer(cubeb_stream * stm)
       LOG("insert silence: ps=%u", packet_size);
       stm->linear_input_buffer.push_silence(packet_size * stm->input_stream_params.channels);
     } else {
-      if (should_upmix(stm->input_mix_params, stm->input_stream_params)) {
+      if (cubeb_should_upmix(&(stm->input_mix_params), &(stm->input_stream_params))) {
```
kinetiknz: Don't need extra parens

src/cubeb_wasapi.cpp
```diff
@@ -614,23 +629,25 @@ bool get_input_buffer(cubeb_stream * stm)
...
         bool ok = stm->linear_input_buffer.reserve(stm->linear_input_buffer.length() +
                                                    packet_size * stm->input_stream_params.channels);
         XASSERT(ok);
-        upmix(reinterpret_cast<float*>(input_packet), packet_size,
+        cubeb_upmix_float(reinterpret_cast<float*>(input_packet), packet_size,
               stm->linear_input_buffer.data() + stm->linear_input_buffer.length(),
```
kinetiknz: Fix indentation after function name change

src/cubeb_wasapi.cpp
```diff
@@ -614,23 +629,25 @@ bool get_input_buffer(cubeb_stream * stm)
...
               stm->input_mix_params.channels,
               stm->input_stream_params.channels);
         stm->linear_input_buffer.set_length(stm->linear_input_buffer.length() + packet_size * stm->input_stream_params.channels);
-      } else if (should_downmix(stm->input_mix_params, stm->input_stream_params)) {
+      } else if (cubeb_should_downmix(&(stm->input_mix_params), &(stm->input_stream_params))) {
```
kinetiknz: Don't need extra parens

src/cubeb_wasapi.cpp
```diff
@@ -614,23 +629,25 @@ bool get_input_buffer(cubeb_stream * stm)
...
         bool ok = stm->linear_input_buffer.reserve(stm->linear_input_buffer.length() +
                                                    packet_size * stm->input_stream_params.channels);
         XASSERT(ok);
-        downmix(reinterpret_cast<float*>(input_packet), packet_size,
+        cubeb_downmix_float(reinterpret_cast<float*>(input_packet), packet_size,
                 stm->linear_input_buffer.data() + stm->linear_input_buffer.length(),
                 stm->input_mix_params.channels,
```
kinetiknz: Fix indent

src/cubeb_wasapi.cpp
```diff
@@ -1497,16 +1538,39 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
   }
   com_heap_ptr<WAVEFORMATEX> mix_format(tmp);
 
-  handle_channel_layout(stm, mix_format, stream_params);
+  /* Set channel layout only when there're more than two channels. Otherwise,
+   * use the default setting retrieved from the stream format of the audio
+   * engine's internal processing by GetMixFormat.*/
```
kinetiknz: Space between . and */

src/cubeb_wasapi.cpp
```diff
@@ -1497,16 +1538,39 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
...
   /* Shared mode WASAPI always supports float32 sample format, so this
    * is safe. */
   mix_params->format = CUBEB_SAMPLE_FLOAT32NE;
   mix_params->rate = mix_format->nSamplesPerSec;
   mix_params->channels = mix_format->nChannels;
-  LOG("Setup requested=[f=%d r=%u c=%u] mix=[f=%d r=%u c=%u]",
+  WAVEFORMATEXTENSIBLE * format_pcm = reinterpret_cast<WAVEFORMATEXTENSIBLE *>(mix_format.get());
+  mix_params->layout = mask_to_channel_layout(format_pcm->dwChannelMask);
+  if (mix_params->layout == CUBEB_LAYOUT_UNSUPPORTED) {
+    LOG("Speakers use the unsupported layout!\n");
```
kinetiknz: "Output using undefined layout" or something?

src/cubeb_wasapi.cpp
```diff
@@ -1497,16 +1538,39 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
...
+  } else if (mix_format->nChannels != CUBEB_CHANNEL_LAYOUT_MAPS[mix_params->layout].channels) {
+    // The CUBEB_CHANNEL_LAYOUT_MAPS[mix_params->layout].channels may be
+    // different from the mix_params->channels. 6 channel ouput with stereo
+    // layout is acceptable in Windows. If this happens, it should not downmix
+    // audio according to layout.
+    LOG("Channel counts is different from the layout standard!\n");
```
kinetiknz: "Channel count" (no "s")

src/cubeb_wasapi.cpp
```diff
@@ -1710,10 +1774,14 @@ wasapi_stream_init(cubeb * context, cubeb_stream ** stream,
   if (input_stream_params) {
     stm->input_stream_params = *input_stream_params;
     stm->input_device = utf8_to_wstr(reinterpret_cast<char const *>(input_device));
+    // Make sure the layout matches to the channels.
+    XASSERT(stm->input_stream_params.channels == CUBEB_CHANNEL_LAYOUT_MAPS[stm->input_stream_params.layout].channels);
   }
   if (output_stream_params) {
     stm->output_stream_params = *output_stream_params;
     stm->output_device = utf8_to_wstr(reinterpret_cast<char const *>(output_device));
+    // Make sure the layout matches to the channels.
```
kinetiknz: "matches the channel count"

src/cubeb_wasapi.cpp
```diff
@@ -1710,10 +1774,14 @@ wasapi_stream_init(cubeb * context, cubeb_stream ** stream,
   if (input_stream_params) {
     stm->input_stream_params = *input_stream_params;
     stm->input_device = utf8_to_wstr(reinterpret_cast<char const *>(input_device));
+    // Make sure the layout matches to the channels.
```
kinetiknz: "matches the channel count"

src/cubeb_wasapi.cpp
ChunMinChang: Oh, I misunderstood the code so I modified it. I will change it back.

include/cubeb/cubeb.h
```diff
@@ -176,12 +176,84 @@ typedef enum {
...
+ * 3F1-LFE        L   R   C    LFE S
+ * 2F2            L   R   LS   RS
+ * 2F2-LFE        L   R   LFE  LS   RS
+ * 3F2            L   R   C    LS   RS
+ * 3F2-LFE        L   R   C    LFE  LS   RS
+ * 3F3R-LFE       L   R   C    LFE  BC   LS   RS
+ * 3F4-LFE        L   R   C    LFE  Rls  Rrs  LS   RS
+ */
+typedef enum {
+  CUBEB_LAYOUT_UNSUPPORTED, // Indicate the speaker's layout is not supported.
```
kinetiknz: In that case, UNSUPPORTED is fine.  I was thinking of also using the value to request the case where the caller wants to ask for no up/downmixing or remapping (which is perhaps too much of an edge case to worry about).

ChunMinChang: If we allow something like LAYOUT_UNDEFINDED(LAYOUT_UNSUPPORTED) to ask for no up/downmixing, would it be ok to write audio data with 6 channels to the output buffer with stereo layout? Or we just apply the fullback plan(passing data by channel index) when caller use LAYOUT_UNDEFINDED? 

Does this need to implement in this patch? Maybe we could open another bug to discuss about it.

ChunMinChang:
@kinetiknz : Thanks for your help! I already revise the patches, are you available to review the patches again?

kinetiknz:
Thanks, I'll review again now.  For future reference (although this doesn't apply to every project on GitHub), it's preferable to include the changes made during review as a separate commit initially (which we can squash down later) because it's much easier to see what changed and GitHub's PR mechanism doesn't offer a way to generate an interdiff between PR revisions.

PR #172, opened 2016-10-21
Add a test that makes sure the clock value is not reset by starting/stopping the stream

padenot:
This would have caught the OpenSLES issues when merging into gecko I believe

### DIFF
--- a/test/test_sanity.cpp
+++ b/test/test_sanity.cpp
@@ -471,6 +471,21 @@ test_stream_position(void)
     delay(500);
   }
 
+  /* test that the position is valid even when starting and
+   * stopping the stream.  */
+  for (i = 0; i < 5; ++i) {
+    r = cubeb_stream_stop(stream);
+    assert(r == 0);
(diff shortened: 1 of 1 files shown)


test/test_sanity.cpp
```diff
@@ -471,6 +471,21 @@ test_stream_position(void)
...
   }
 
+  /* test that the position is valid even when starting and
+   * stopping the stream.  */
+  for (i = 0; i < 5; ++i) {
+    r = cubeb_stream_stop(stream);
+    assert(r == 0);
+    r = cubeb_stream_get_position(stream, &position);
+    assert(r == 0);
+    assert(last_position <= position);
```
achronop: Since we wait for 500ms I would prefer to turn this check to `last_position < position`. I do not know why to expect that the position is not moved at all and if it hasn't moved it's an issue worth investigation.

padenot: I agree, fixed.

PR #175, opened 2016-10-25
Fix unused parameter/variable warnings in cubeb_audiounit.cpp

padenot:
@achronop, PTAL


achronop:
Looks good

kinetiknz:
For C++ it's better to remove the parameter completely rather than clutter the function decl up with comments.

PR #176, opened 2016-10-25
Improve logging

padenot:
We need to be able to ask people to enable logging and have a lot of info from machines we don't control, so I added a bunch of logging statements. Sometimes, the logging statement is only there to see where the program errored.

This includes logging about device enumeration, that I copied and adapted from `test_device.c`.

@kinetiknz, @achronop, can you have a look ?

### DIFF
--- a/src/cubeb.c
+++ b/src/cubeb.c
@@ -384,10 +384,97 @@ int cubeb_stream_register_device_changed_callback(cubeb_stream * stream,
   return stream->context->ops->stream_register_device_changed_callback(stream, device_changed_callback);
 }
 
+void log_device(cubeb_device_info * device_info)
+{
+  char devfmts[128] = "";
+  const char * devtype, * devstate, * devdeffmt;
+
+  switch (device_info->type) {
+    case CUBEB_DEVICE_TYPE_INPUT:
+      devtype = "input";
+      break;
+    case CUBEB_DEVICE_TYPE_OUTPUT:
+      devtype = "output";
+      break;
+    case CUBEB_DEVICE_TYPE_UNKNOWN:
+    default:
+      devtype = "unknown?";
+      break;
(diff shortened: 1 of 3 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -749,11 +756,16 @@ refill_callback_output(cubeb_stream * stm)
     return true;
   }
 
+
```
achronop: Nit: Extra new line

padenot: Fixed.

src/cubeb.c
```diff
@@ -384,18 +384,113 @@ int cubeb_stream_register_device_changed_callback(cubeb_stream * stream,
   return stream->context->ops->stream_register_device_changed_callback(stream, device_changed_callback);
 }
 
+void log_device(cubeb_device_info * device_info)
```
kinetiknz: Should be static

src/cubeb_audiounit.cpp
```diff
@@ -1180,6 +1180,7 @@ setup_audiounit_stream(cubeb_stream * stm)
   UInt32 size;
 
   if (has_input(stm)) {
+    LOG("Opening input side, rate: %d", stm->input_stream_params.rate);
```
kinetiknz: Should be %u, rate is unsigned

src/cubeb_audiounit.cpp
```diff
@@ -1190,6 +1191,7 @@ setup_audiounit_stream(cubeb_stream * stm)
   }
 
   if (has_output(stm)) {
+    LOG("Opening output side, rate: %d", stm->input_stream_params.rate);
```
kinetiknz: Ditto

PR #180, opened 2016-11-01
audiounit: Revert bad allocation change made in c4dc6e02

kinetiknz:
The deallocation is handled via free() in cubeb_device_info_destroy in cubeb.c
and the allocator and deallocator *must* match.


achronop:
Looks good thanks. We missed that when converting from c to cpp. 

Since you are doing it, we have a similar case here: cubeb_audiounit.cpp#L2197 and cubeb_audiounit.cpp#L2175

kinetiknz:
I'll take a look tomorrow.  I see cubeb_wasapi.cpp had the same bug introduced in 6f131137, so I'll fix that too.

kinetiknz:
Latest version fixes the mismatch on line 2197.

PR #181, opened 2016-11-02
wasapi: Revert bad allocation changes made in 6f13113.

kinetiknz:
cubeb_device_info member deallocation is handled via free() in
cubeb_device_info_destroy in cubeb.c but was being allocated with new[]
in wstr_to_utf8.

std::unique_ptr<wchar_t *> uses a different deallocator to
std::unique_ptr<wchar_t []>, resulting in an allocator mismatch.

r? @padenot

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -106,7 +106,7 @@ int wasapi_stream_start(cubeb_stream * stm);
 void close_wasapi_stream(cubeb_stream * stm);
 int setup_wasapi_stream(cubeb_stream * stm);
 static char * wstr_to_utf8(const wchar_t * str);
-static const wchar_t * utf8_to_wstr(char* str);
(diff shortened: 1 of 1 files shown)


kinetiknz:
Build failure is caused by an old G++ (spec change from C++11 to C++14), we get:

`cubeb_wasapi.cpp:2009:10: error: cannot convert ‘ret’ from type ‘std::unique_ptr<wchar_t []>’ to type ‘std::unique_ptr<wchar_t []>&&’`

Which can be solved with an explicit `std::move` on the return value.

padenot:
Thanks. Maybe it would be good to standardize on `new` instead of malloc at some point.

kinetiknz:
> Maybe it would be good to standardize on new instead of malloc at some point.

Sure.  For that to be safe, all backends need to be converted at once or the deallocators in cubeb.c need to be made backend-specific.

PR #182, opened 2016-11-02
 Make API visible cubeb_device_info members and cubeb_devid const.

kinetiknz:
Make API visible cubeb_device_info members and cubeb_devid const.

The API user is never permitted to modify these (nor deallocate them
directly, which non-const can imply), so they should marked const in the
public definition.

I suspect this happened originally because free()ing the
cubeb_device_info members requires casting const away in
cubeb_device_info_destroy, but other than being slightly ugly there's no
problem doing that and it's better to have the public API be consistent.

This also allows fixing the const-ness of utf8_to_wstr and wstr_to_utf8
const in cubeb_wasapi.cpp.

r? @padenot

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -167,7 +167,7 @@ typedef enum {
 
 /** An opaque handle used to refer a particular input or output device
  *  across calls. */
-typedef void * cubeb_devid;
+typedef void const * cubeb_devid;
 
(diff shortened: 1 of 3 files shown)


kinetiknz:
Build error should be fixed by rebasing on the fixed version of PR #181.

kinetiknz:
Ignore the continuous-integration/appveyor/pr breakage, it's from before I rebased the PR.

PR #184, opened 2016-11-03
Avoid using cubeb_devid internally.

kinetiknz:
cubeb_devid is intended to be an opaque type for API users.  Internally,
a cubeb_devid should be converted to the original type at the earliest
possible point.

It's difficult to get the types right on all platforms, since some of the internal types for cubeb_devid have implicit conversion to and from void*.  A useful compile test is to change the definition of cubeb_devid from void* to uintptr_t and rebuild, fixing whatever breaks.

It'd be nice if cubeb_devid was always defined as uintptr_t, but that loses the nice API property that allows passing a NULL pointer for the default device.

Note, this is based on the constify_api branch, so must be merged in the correct order.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -168,9 +168,9 @@ struct cubeb_stream {
   /* Stream creation parameters */
   cubeb_stream_params input_stream_params;
   cubeb_stream_params output_stream_params;
-  cubeb_devid input_device;
   bool is_default_input;
-  cubeb_devid output_device;
+  AudioDeviceID input_device;
+  AudioDeviceID output_device;
   /* User pointer of data_callback */
   void * user_ptr;
   /* Format descriptions */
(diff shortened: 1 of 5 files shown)


padenot:
Depending on the backend, passing `NULL` into the function more-or-less makes sense now. Maybe we should have a special `cubeb_devid` value that means "default" instead of passing `NULL` or `0` at the call site.

Depending on where <https://w3c.github.io/mediacapture-output/> and <https://pbos.github.io/mst-content-hint/> go, we might want to have special values for "communication device" or "media device" or something like that.

kinetiknz:
This PR only aims to remove the internal uses of cubeb_devid to restore cubeb_devid's function as an external-only opaque type.

If we want to change the API, we can do that in a later PR once it's clear what changes are necessary.

I've rebased on top of master with achronop's latest changes, so this is ready to merge once it's r+.

PR #186, opened 2016-11-08
Implement a lock-free single consumer single producer queue

padenot:
This is going to be useful for a number of things:
- I'm trying to fix drifting on WASAPI by using multiple threads for input and output first, before implementing a real adaptative resampling pass (more new on that soon)
- We're locking around the array buffer access on OSX, it's best to use a lock-free wait-free queue
- Possibly for OpenSL ES and PulseAudio as well, we need a buffer between input and output

This queue can be configured at compile time to use atomics or not, so it's quite versatile and we could use it for non-concurrent scenario as well.

In terms of testing, it's clear that this does not _prove_ that the list is correct, but has found a lot of issues on my system: making one of the index non-atomic fails every time on my machine, for example. The design is very classic. Juce's [`AbstractFifo`](https://github.com/julianstorer/JUCE/blob/master/modules/juce_core/containers/juce_AbstractFifo.h) and boost's [`spsc_queue`](https://github.com/boostorg/lockfree/blob/develop/include/boost/lockfree/spsc_queue.hpp) implement something similar.

### DIFF
--- a/src/cubeb_ringbuffer.h
+++ b/src/cubeb_ringbuffer.h
@@ -18,14 +18,16 @@ enum ThreadSafety
   /* No attempt to synchronize the queue is made. The queue is only safe when
    * used on a single thread. */
   Unsafe,
-  /** Atomics are used to synchronize read and write. The queue is safe to used
-   * from two thread: one producer, one consumer. */
+  /** Atomics are used to synchronize read and write. The queue is safe when
+   * used from two thread: one producer, one consumer. */
   Safe
 };
 
 /** Policy to enable thread safety on the queue. */
 template<ThreadSafety>
 struct ThreadSafePolicy;
+
+typedef int RingBufferIndex;
 
 /** Policy for thread-safe internal index for the queue.
  *
@@ -36,7 +38,7 @@ struct ThreadSafePolicy;
 template<>
 struct ThreadSafePolicy<Safe>
 {
-  typedef std::atomic<int> IndexType;
+  typedef std::atomic<RingBufferIndex> IndexType;
 };
 
 /**
@@ -46,7 +48,7 @@ struct ThreadSafePolicy<Safe>
 template<>
 struct ThreadSafePolicy<Unsafe>
 {
-  typedef int IndexType;
+  typedef RingBufferIndex IndexType;
 };
 
 /**
@@ -58,30 +60,31 @@ struct ThreadSafePolicy<Unsafe>
  * to disable the use of atomics at compile time and only use this data
  * structure on one thread.
  *
- * The role for the producer and the consumer must be constant, i.e., the same
- * thread should always be the producer and another thread should always be the
- * consumer.
+ * The role for the producer and the consumer must be constant, i.e., the
+ * producer should always be on one thread and the consumer should always be on
+ * another thread.
  *
  * The public interface of this class uses frames.
  *
  * Some words about the inner workings of this class:
  * - Capacity is fixed. Only one allocation is performed, in the constructor.
- *   When reading and writing, the return value of the method allow checking is
+ *   When reading and writing, the return value of the method allow checking if
  *   the ring buffer is empty or full.
- * - We always keep the read index at least one element ahead of the write
+ * - We always keep the read index at least one frame ahead of the write
  *   index, so we can distinguish between an empty and a full ring buffer: an
  *   empty ring buffer is when the write index is at the same position as the
- *   write index. A full buffer is when the write index is exactly one position
+ *   read index. A full buffer is when the write index is exactly one position
  *   before the read index.
  * - We synchronize updates to the read index after having read the data, and
  *   the write index after having written the data. This means that the each
  *   thread can only touch a portion of the buffer that is not touched by the
  *   other thread.
- * - Caller is expected to provide an output buffer. When writing to the queue,
- *   elements are copied. When reading from the queue, the user is expected to
- *   provide a buffer. Because this is a ring buffer, data might not be
- *   contiguous in memory, providing an external buffer to copy into is an easy
- *   way to have linear data for further processing.
+ * - Callers are expected to provide buffers. When writing to the queue,
+ *   frames are copied into the internal storage from the buffer passed in.
+ *   When reading from the queue, the user is expected to provide a buffer.
+ *   Because this is a ring buffer, data might not be contiguous in memory,
+ *   providing an external buffer to copy into is an easy way to have linear
+ *   data for further processing.
  */
 template <typename T,
           ThreadSafety Safety = ThreadSafety::Safe>
@@ -101,13 +104,14 @@ public:
     : read_index_(0)
     , write_index_(0)
     , channel_count_(channel_count)
-    /* One element more to distinguish from emtpy and full buffer. */
+    /* One frame more to distinguish from emtpy and full buffer. */
     , capacity_(frames_to_samples(capacity_in_frames + 1))
   {
     static_assert(std::is_trivial<T>::value,
                   "ring_buffer_base requires trivial type");
 
-    assert(storage_capacity() < std::numeric_limits<T>::max() &&
+    assert(storage_capacity() <
+           std::numeric_limits<RingBufferIndex>::max() &&
            "buffer to large for the type of index used.");
     assert(channel_count_ > 0);
     assert(capacity_in_frames > 0);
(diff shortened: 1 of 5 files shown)


kinetiknz:
Sorry for the delay.  I'll take a look at this soon (maybe tomorrow, otherwise early next week).

test/test_ring_buffer.cpp
```diff
@@ -0,0 +1,191 @@
...
+  }
+}
+
+template<typename T>
+void test_ring_multi(lock_free_queue<T>& buf, int channels, int capacity_frames)
+{
+  sequence_verifier<T> checker(channels);
+  std::unique_ptr<T[]> out_buffer(new T[capacity_frames * channels]);
+
+  const int block_size = 128;
```
achronop: Nit: Hard coded values, through out the test, can be represent by a macro or `const int`

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,391 @@
...
+   *
+   * @param channel_count the number of channels of the stream for this ring buffer.
+   * @param capacity_in_frames The maximum number of frames this ring buffer will hold.
+   */
+  ring_buffer_base(int channel_count, int capacity_in_frames)
+    : read_index_(0)
+    , write_index_(0)
+    , channel_count_(channel_count)
+    /* One element more to distinguish from emtpy and full buffer. */
+    , capacity_(frames_to_samples(capacity_in_frames + 1))
```
padenot: I decided not to because it's not necessary, and using `int` allows checking underflow more easily.

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,391 @@
...
+  {
+    static_assert(std::is_trivial<T>::value,
+                  "ring_buffer_base requires trivial type");
+
+    assert(storage_capacity() < std::numeric_limits<T>::max() &&
+           "buffer to large for the type of index used.");
```
padenot: This was wrong. Fixed.

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,391 @@
...
+    assert(channel_count_ > 0);
+    assert(capacity_in_frames > 0);
+
+    data_.reset(new T[frames_to_samples(storage_capacity())]);
+    PodZero(data_.get(), storage_capacity());
```
padenot: Yes.

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,391 @@
...
+   * `elements`, if non-null.
+   *
+   * Only safely called on the consumer side.
+   *
+   * @param elements A pointer to a buffer with space for at least `count`
+   * frames of audio. If `elements` is `nullptr`, elements will be discarded.
+   * @param count The maximum number of elements to dequeue.
+   * @return The number of frames of audio written to `elements`.
+   */
+  int dequeue(T * elements, int count)
```
padenot: Why would you use this for ? We can still add it later if needed.

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,391 @@
...
+  {
+    /** Don't make this two operations, `index` might be atomic, we want other
+     * threads to see either the old or the new value, but not an intermediary
+     * computation step: it should only be assigned once. */
+    index = (index + increment) % capacity_;
+  }
+  /** Index at which the oldest frame is at. In samples. */
+  typename ThreadSafePolicy<Safety>::IndexType read_index_;
+  /** Index at which to write new frames. In samples. `write_index` is always at
+   * most one element behind `read_index_`. */
```
padenot: Hopefully the new way if clearer.

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,391 @@
...
+   * Get the number of available frames of audio for consuming.
+   *
+   * Only safely called on the producer thread.
+   *
+   * @return The number of empty slots in the buffer, available for writing.
+   */
+  int available_write() const
+  {
+    return samples_to_frames(available_write_internal(read_index_,
+                                                      write_index_));
```
padenot: Done.

padenot:
@kinetiknz, the last commit implements what we talked about over email, namely having a non-audio-aware base class. It also adds a method that take a single element, and adds some tests.

padenot:
I busted the build but I have it fixed on a branch that update the compilers to use GCC 5, I'll push my fixes here tomorrow, but it's certainly reviewable.

kinetiknz:
> I busted the build but I have it fixed on a branch that update the compilers to use GCC 5, I'll push my fixes here tomorrow, but it's certainly reviewable.

If it's just `std::is_trivially_copyable<T>::value`, use std::is_trivial<T>::value instead.  It's not quite the same, but that's what PodZero et. al. are using.  I'd prefer not to upgrade Travis too much to try to keep it as close to the minimum build requirements for Gecko as possible, otherwise we just end up having to fix breakage later in the dev cycle.

padenot:
> If it's just std::is_trivially_copyable<T>::value, use std::is_trivial<T>::value instead. It's not quite the same, but that's what PodZero et al. are using. I'd prefer not to upgrade Travis too much to try to keep it as close to the minimum build requirements for Gecko as possible, otherwise we just end up having to fix breakage later in the dev cycle.

I made this because I'm using the lock-free queue to implement a asynchronous logging mechanism (to come in another PR once this is merged) so we can log a lot from the callback to diagnose things without perturbing the real-time thread. I had a lot of issues and mis-diagnosis of issues recently because printfs and other log was syncing with the main thread (because printf and friends take a global lock). This caused problems because the new logging mechanism is using a `memcpy`-safe but non trivial class for the log message.

Instead of updating the compiler, I've taken an alternative route, writing new memory-manipulation functions that behave similarly to `PodCopy` and `PodZero`, but handle non-trivial objects properly, statically dispatching betwee `PodCopy` and `PodZero` (for trivial objects and numeric types) and a normal loop (for object that need to have the ctor and dtor called). This works on gcc 4.8 and clang and visual studio, and it's better anyways.

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,541 @@
...
+   * @param index a reference to the index to increment.
+   * @param increment the number by which `index` is incremented.
+   */
+  template <typename IndexType>
+  void increment_index(IndexType& index, uint32_t increment) const
+  {
+    /** Don't make this two operations, `index` might be atomic, we want other
+     * threads to see either the old or the new value, but not an intermediary
+     * computation step: it should only be assigned once. */
+    index = (index + increment) % capacity_;
```
kinetiknz: Unless the value of `capacity_` is limited to ~half of the max value, `index + increment` could overflow, which is undefined behaviour with the current signed index type.

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,541 @@
...
+   * @param capacity The maximum number of element this ring buffer will hold.
+   */
+  ring_buffer_base(int capacity)
+    : read_index_(0)
+    , write_index_(0)
+    /* One more element to distinguish from emtpy and full buffer. */
+    , capacity_(capacity + 1)
+  {
+    assert(storage_capacity() <
+           std::numeric_limits<RingBufferIndex>::max() &&
```
kinetiknz: If you're going to use signed index types, the capacity needs to be limited to ~half of the max value, otherwise there are various places in the code where signed overflow can occur, which is undefined behaviour.

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,541 @@
...
+  {
+    return (write_index + 1) % capacity_ == read_index;
+  }
+  /**
+   * Return the size of the storage. It is one more than the number of elements
+   * that can be stored in the buffer.
+   *
+   * @return the number of elements that can be stored in the buffer.
+   */
+  int storage_capacity() const
```
kinetiknz: Either this isn't needed, or places using `capacity_` directly need to be fixed.

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,541 @@
...
+   * @return The number of empty slots in the buffer, available for writing.
+   */
+  int available_write() const
+  {
+    return available_write_internal(read_index_, write_index_);
+  }
+  /**
+   * Get the total capacity, for this ring buffer.
+   *
+   * Can be called safely on any thread.
```
kinetiknz: I think there's technically a data race here under very weakly ordered memory models (maybe only Alpha which basically means we don't care, I think this bug would rely on dependent load reordering).  If one thread creates the queue and then publishes the pointer to access it to the other thread and that thread calls `capacity()`, there's no release/acquire that ensures the value of `capacity_` is visible in the calling thread.  I think setting the index atomics as the last action in the ctor would be sufficient to prevent that, since (in the default seq_cst mode we're using) they act as a full barrier.

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,541 @@
...
+    return rv;
+  }
+  /**
+   * Increments an index, wrapping it around the storage.
+   *
+   * @param index a reference to the index to increment.
+   * @param increment the number by which `index` is incremented.
+   */
+  template <typename IndexType>
+  void increment_index(IndexType& index, uint32_t increment) const
```
kinetiknz: This should return the new index for the caller to use, and should take a simple index type by value (and not the std::atomic wrapper).  The type of `increment` should be the same basic type as the index type, and asserted to be >= 0 if IndexType is signed.

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,541 @@
...
+ * to disable the use of atomics at compile time and only use this data
+ * structure on one thread.
+ *
+ * The role for the producer and the consumer must be constant, i.e., the
+ * producer should always be on one thread and the consumer should always be on
+ * another thread.
+ *
+ * Some words about the inner workings of this class:
+ * - Capacity is fixed. Only one allocation is performed, in the constructor.
+ *   When reading and writing, the return value of the method allow checking if
```
kinetiknz: allows

src/cubeb_ringbuffer.h
```diff
@@ -0,0 +1,541 @@
...
+   */
+  ring_buffer_base(int capacity)
+    : read_index_(0)
+    , write_index_(0)
+    /* One more element to distinguish from emtpy and full buffer. */
+    , capacity_(capacity + 1)
+  {
+    assert(storage_capacity() <
+           std::numeric_limits<RingBufferIndex>::max() &&
+           "buffer to large for the type of index used.");
```
kinetiknz: Typo: to -> too

kinetiknz:
Thanks, looks great! I merged this in 4aa67a9a9a720741004130540e270ee944d5adc7 with a couple of minor tweaks:
- removed the remaining gcc-5 .travis.yml changes
- removed CUBEB_RELEASE_CONST and made the thread ids mutable instead

PR #187, opened 2016-11-09
Bail out safely from the rendering loop if we could not join the rendering thread in time

padenot:
This is over-complicated because we don't have `atomic_shared_ptr` in standard C++ for now, and because I don't want this to be non-atomic, and I don't want to add a global lock or another mutex.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -22,6 +22,7 @@
 #include <algorithm>
 #include <memory>
 #include <limits>
+#include <atomic>
 
 #include "cubeb/cubeb.h"
 #include "cubeb-internal.h"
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -1155,9 +1164,13 @@ void stop_and_join_render_thread(cubeb_stream * stm)
      * process. */
     LOG("Destroy WaitForSingleObject on thread timed out,"
         " leaking the thread: %d", GetLastError());
+    *(stm->emergency_bailout) = true;
+    rv = false;
   }
   if (r == WAIT_FAILED) {
+    *(stm->emergency_bailout) = true;
```
kinetiknz: Swap this line and the LOG, just for consistency with the code above it.

src/cubeb_wasapi.cpp
```diff
@@ -1903,7 +1931,12 @@ int wasapi_stream_stop(cubeb_stream * stm)
     stm->state_callback(stm, stm->user_ptr, CUBEB_STATE_STOPPED);
   }
 
-  stop_and_join_render_thread(stm);
+  if (stop_and_join_render_thread(stm)) {
+    if (stm->emergency_bailout.load()) {
+      delete stm->emergency_bailout.load();
```
kinetiknz: It's safe to delete a nullptr, so the if (load) test can be dropped.

PR #189, opened 2016-11-10
Convert tests to gtests

kinetiknz:
This replaces the ad-hoc test harness we're currently using with googletest pulled in via a git submodule, and converts the existing tests to gtests.

The conversion was fairly crude and mechanical aiming for minimal changes to aid review.  We can make a second pass later to make more use of the facilities provided by gtest.

Aside from being generally nicer to write tests with a proper harness (and we need more and better tests), additional motivation for this comes from BMO bug 1314514 trying to remove the XPCOM dependencies from the cubeb tests, which is blocked on converting the existing tests to gtests.

Note, to build this branch, you must perform an initial submodule setup:
1.  `git submodule init`
2.  `git submodule update`


padenot:
Looks good. Do you know what's up with the failure on AppVeyor ? If that's an AppVeyor issue, we can simply merge. I can't retrigger (or I haven't found how).

I'll handle rebasing the lockfree queue on top of this PR.

kinetiknz:
appveyor/pr got caught out when I rebased and tried to grab a cset that no longer exists.  It seems to happen a lot with that particular builder; I'll dig around and see if there's a solution.

PR #190, opened 2016-11-10
pulse: Bail early from pulse_defer_event_cb when shutting down.

kinetiknz:
When starting a stream, trigger_user_callback may be called from
stream_write_callback and immediately enter a drain situation, creating
a drain timer and setting shutdown to true.  If pulse_defer_event_cb
then runs without checking for shutdown, it can overwrite the current
timer with a new timer, resulting in a leaked timer and a null pointer
assertion.

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -181,9 +181,9 @@ static void
 stream_drain_callback(pa_mainloop_api * a, pa_time_event * e, struct timeval const * tv, void * u)
 {
   (void)a;
-  (void)e;
   (void)tv;
   cubeb_stream * stm = u;
+  assert(stm->drain_timer == e);
   stream_state_change_callback(stm, CUBEB_STATE_DRAINED);
   /* there's no pa_rttime_free, so use this instead. */
   a->time_free(stm->drain_timer);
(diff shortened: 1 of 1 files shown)


kinetiknz:
I've seen this cause the mda mochitests to fail with a frequency of 1:10 on `Linux x64 opt tc-M(mda)`, usually while running the test dom/media/mediasource/test/test_WaitingToEndedTransition_mp4.html.

achronop:
The fix is safe, the only thing makes me wonder is how the `stream_write_callback` is called before the `pulse_defer_event_cb`. Actually the reason that `pulse_defer_event` exists is to trigger `stream_write_callback` and make it running. 

To be more specific write callback is expected before  `stream_start` due to pre-buffering. We ignore that callback (not pushing any data) so it stops (that's a PA thing). On `stream_start` we send the defer event to push some data and make the write callback roll again. If write callback is running we do not need the defer event at all. I wonder if the write callback that you see is due to a late pre-buffering (so state == CUBEB_STATE_STARTED).

Maybe it's a good idea to set the write callback (`pa_stream_set_write_callback`) in `stream_start` method in order to eliminate the possibility to have it before defer callback. Otherwise, in theory, we can completely ignore the defer callback no matter if it is a shutdown or not.

Shutdown is also set from the `pulse_stream_stop` method, so that fix is valid when we start and stop immediately.

I hope I get the scenario correctly. Please let me know your opinion.

kinetiknz:
> The fix is safe, the only thing makes me wonder is how the stream_write_callback is called before the pulse_defer_event_cb. Actually the reason that pulse_defer_event exists is to trigger stream_write_callback and make it running.

I should've been more clear: this happens for streams that have been started and stopped repeatedly.

I think (from the logging I did on try) what's happening is the stream is corked just before PA would've called stream_write_callback,  so uncorking (which happens before pulse_defer_event_cb is queued)  results in a  stream_write_callback before pulse_defer_event_cb.

kinetiknz:
50 green `Linux x64 opt tc-M(mda)` runs: https://treeherder.mozilla.org/#/jobs?repo=try&revision=7d1068029d439137cec49d612681c1c9af4671dc

achronop:
I see the problem in the logs. Thanks for the extra info.

PR #192, opened 2016-11-11
wasapi: First pass at using modern C++.

kinetiknz:
This converts the WASAPI backend to using smart pointers where possible.  I wrote a small com_ptr since the standard Microsoft version requires building with ATL, which we (and Gecko) want to avoid.

There's plenty more we can do (e.g. avoid double-initializating cubeb_stream by passing the params into a constructor), but I've tried to make each of these changes minimal to make review easier.

r? @padenot


kinetiknz:
I've made a similar set of changes for AudioUnit in https://github.com/kinetiknz/cubeb/tree/audiounit_cleanup, but that's not quite ready to review and merge yet.

PR #195, opened 2016-11-15
audiounit: First pass at using modern C++.

kinetiknz:
Same concept as PR #192, but for the AudioUnit backend.

r? @padenot or @achronop

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ before_install:
       if [[ $TRAVIS_OS_NAME = "osx" ]]; then
         . scan-build-install.sh;
       else
-        export SCAN_BUILD_PATH=scan-build;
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ if(POLICY CMP0063)
   cmake_policy(SET CMP0063 NEW)
 endif()
 set(CMAKE_C_STANDARD 99)
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
```
kinetiknz: I'll have to revert this change and deal with the fallout.  Gecko is only built with C++11; not sure where I got the idea that it had already moved to C++14 from.

kinetiknz:
> This also moves the initialization of the members from the ctor to the class definition itself, maybe it would be good to put it in the commit message.

Thanks, that should've been in a separate commit.

PR #201, opened 2016-11-23
pulse: Complete pulse_stream_set_panning implementation.

kinetiknz:
The existing implementation called pa_cvolume_set_balance on an
uninitialized pa_cvolume (which triggered a Valgrind warning) and then
forgot to apply the result to the sink input.

This is https://bugzilla.mozilla.org/show_bug.cgi?id=1319623

r? @achronop

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -27,6 +27,7 @@
   X(pa_context_get_server_info)                 \
   X(pa_context_get_sink_info_by_name)           \
   X(pa_context_get_sink_info_list)              \
+  X(pa_context_get_sink_input_info)             \
   X(pa_context_get_source_info_list)            \
   X(pa_context_get_state)                       \
   X(pa_context_new)                             \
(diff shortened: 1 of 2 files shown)


src/cubeb_pulse.c
```diff
@@ -999,23 +1000,64 @@ pulse_stream_set_volume(cubeb_stream * stm, float volume)
...
+
+  struct sink_input_info_result r = { &cvol, stm->context->mainloop };
+  op = WRAP(pa_context_get_sink_input_info)(stm->context->context,
+                                            index,
+                                            sink_input_info_cb,
+                                            &r);
+  if (op) {
+    operation_wait(stm->context, stm->output_stream, op);
+    WRAP(pa_operation_unref)(op);
+  }
```
achronop: One discussion point is, if we can just use the `pa_cvolume_init` instead of query the current value. The `pa_cvolume_set_balance` method will modify it anyway or fail. In failure we can use `pa_cvolume_valid` and not set it if the latter return false. Those methods do not go through the server so might be faster. In any case panning is not a high traffic method and it's good as is.

src/cubeb_pulse.c
```diff
@@ -999,23 +1000,64 @@ pulse_stream_set_volume(cubeb_stream * stm, float volume)
...
+
+  WRAP(pa_cvolume_set_balance)(&cvol, map, panning);
+
+  op = WRAP(pa_context_set_sink_input_volume)(stm->context->context,
+                                              index, &cvol, volume_success,
+                                              stm);
+  if (op) {
+    operation_wait(stm->context, stm->output_stream, op);
+    WRAP(pa_operation_unref)(op);
+  }
```
achronop: +1 we were totally missing that and I really wonder how right now works (?)

kinetiknz: It didn't work at all... since it's not used in Gecko and the test in test_audio needs manual verification, I guess nobody noticed.

src/cubeb_pulse.c
kinetiknz: I tried that first, but it didn't work.

pa_cvolume_init doesn't give you a pa_cvolume that is valid to use (pa_cvolume_valid returns false), it's basically a memset.  You need a combination of a volume and a sample spec to create a valid one, and since we don't know the sink input's current volume we need to query it, which is how we end up with the current code.

PR #202, opened 2016-11-28
Support playback+capture for alsa

debianuser1:
Pull request for https://gist.github.com/anonymous/77ecd0d54b9484be89f312484e8d4ace patches implementing playback+capture support for alsa. Comments welcome!

### DIFF
--- a/src/cubeb_alsa.c
+++ b/src/cubeb_alsa.c
@@ -107,6 +107,9 @@ struct cubeb_stream {
      being logically active and playing. */
   struct timeval last_activity;
   float volume;
+
+  char * buffer;
+  snd_pcm_uframes_t bufframes;
 };
 
 static int
@@ -253,8 +256,6 @@ alsa_refill_stream(cubeb_stream * stm)
 {
   unsigned short revents;
   snd_pcm_sframes_t avail;
-  long got;
-  void * p;
   int draining;
 
   draining = 0;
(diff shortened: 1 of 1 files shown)


kinetiknz:
Thanks for this! I'll try to get it reviewed soon, but I have a backlog of stuff so it may be a little while.

One thing that needs to be looked at is the Clang analyzer failure showing up on CI at https://travis-ci.org/kinetiknz/cubeb/jobs/179335705.  It looks like you need to handle errors from `alsa_get_max_channel_count` and `alsa_get_preferred_sample_rate` in `alsa_enumerate_devices`.

debianuser1:
Thank you for your response! I added error checks to alsa_enumerate_devices() and rebased commits. Also updated patches gist: https://gist.github.com/debianuser1/25964bea1a58a16aa30f0d72d3b8cae7 Tell if you want me to do anything else.

src/cubeb_alsa.c
```diff
@@ -107,6 +107,12 @@ struct cubeb_stream {
      being logically active and playing. */
   struct timeval last_activity;
   float volume;
+
+  char *buffer;
```
kinetiknz: Space between * and buffer

src/cubeb_alsa.c
```diff
@@ -235,6 +241,14 @@ set_timeout(struct timeval * timeout, unsigned int ms)
 }
 
 static void
+stream_buffer_decrement(cubeb_stream * stm, long count)
+{
+  char *bufremains = stm->buffer + snd_pcm_frames_to_bytes(stm->pcm, count);
```
kinetiknz: Space between * and bufremains

src/cubeb_alsa.c
```diff
@@ -249,92 +263,173 @@ alsa_set_stream_state(cubeb_stream * stm, enum stream_state state)
...
   int draining;
 
   draining = 0;
 
   pthread_mutex_lock(&stm->mutex);
 
+  /* Call _poll_descriptors_revents() even if we don't use it
+     to let underlying plugins clear null events.  Otherwise poll()
+     may wake up again and again, producing unnecessary CPU usage. */
+  snd_pcm_poll_descriptors_revents(stm->pcm, stm->fds, stm->nfds, &revents);
```
kinetiknz: Interesting! Thanks.

src/cubeb_alsa.c
```diff
@@ -774,37 +869,28 @@ alsa_destroy(cubeb * ctx)
...
+                        cubeb_state_callback state_callback,
+                        void * user_ptr)
 {
   (void)stream_name;
   cubeb_stream * stm;
   int r;
   snd_pcm_format_t format;
   snd_pcm_uframes_t period_size;
   int latency_us = 0;
+  const char *pcm_name = deviceid ? (const char *)deviceid : CUBEB_ALSA_PCM_NAME;
```
kinetiknz: Space between ) and deviceid, and const after char x2

src/cubeb_alsa.c
```diff
@@ -1125,13 +1283,62 @@ alsa_stream_set_volume(cubeb_stream * stm, float volume)
...
+  if (r != CUBEB_OK) {
+    return CUBEB_ERROR;
+  }
+
+  r = alsa_get_max_channel_count(context, &max_channels);
+  if (r != CUBEB_OK) {
+    return CUBEB_ERROR;
+  }
+
+  *collection = (cubeb_device_collection *)calloc(1, sizeof(cubeb_device_collection) + 1*sizeof(cubeb_device_info *));
```
kinetiknz: Space between ) and calloc

src/cubeb_alsa.c
```diff
@@ -1125,13 +1283,62 @@ alsa_stream_set_volume(cubeb_stream * stm, float volume)
...
+  assert(*collection);
+
+  const char * a_name = "default";
```
kinetiknz: const after char

src/cubeb_alsa.c
```diff
@@ -1125,13 +1283,62 @@ alsa_stream_set_volume(cubeb_stream * stm, float volume)
...
+  (*collection)->device[0] = (cubeb_device_info *)calloc(1, sizeof(cubeb_device_info));
```
kinetiknz: Space between ) and calloc

debianuser1:
Ah, thank you for your comment! I was not paying attention to spaces. I hope they're all fixed now.

I also updated patches gist: https://gist.github.com/debianuser1/25964bea1a58a16aa30f0d72d3b8cae7

As for cubeb_alsa.c:1095:29: warning: Access to field 'pcm' results in a dereference of a null pointer (loaded from variable 'stm')... That 'stm' variable could only be null if the previous alsa_stream_init() call fails, and that's already checked right there a few lines above. Or should there be some better check?

debianuser1:
Looks like assert() makes scan-build happy. But is that the right way to fix it?

kinetiknz:
Thanks for the updates!

> Looks like assert() makes scan-build happy. But is that the right way to fix it?

Yeah, I think that's fine.  The `assert()` is providing the analyzer with information we knew that it couldn't deduce without help.

kinetiknz:
I pushed the squashed version as e5f6b69f6cb46e283d9d39dcce5c5457dae706a0.

Thanks for the patch!

PR #204, opened 2016-12-21
Warning fix

ChunMinChang:
Fix warning on Travis CI

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -795,7 +795,8 @@ refill_callback_input(cubeb_stream * stm)
                      nullptr,
                      0);
 
-  consumed_all_buffer = (unsigned long)read == stm->linear_input_buffer.length();
+  XASSERT(read >= 0);
+  consumed_all_buffer = (unsigned long) read == stm->linear_input_buffer.length();
 
   stm->linear_input_buffer.clear();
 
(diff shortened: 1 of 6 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -831,15 +831,15 @@ refill_callback_output(cubeb_stream * stm)
        output_frames, got);
 
   XASSERT(got >= 0);
-  XASSERT(got == output_frames || stm->draining);
+  XASSERT((unsigned long)got == output_frames || stm->draining);
```
kinetiknz: Space between ) and got here and everywhere else there's a cast.

src/cubeb_wasapi.cpp
```diff
@@ -795,7 +795,7 @@ refill_callback_input(cubeb_stream * stm)
                      nullptr,
                      0);
 
-  consumed_all_buffer = read == stm->linear_input_buffer.length();
+  consumed_all_buffer = (unsigned long)read == stm->linear_input_buffer.length();
```
kinetiknz: I think we need an `XASSERT(read >= 0)` before this to ensure we're not casting negative values into an unsigned type.

ChunMinChang: ok.

src/cubeb_wasapi.cpp
ChunMinChang: ok.

kinetiknz:
Thanks! I merged this via a squash in https://github.com/kinetiknz/cubeb/commit/b0f1a6331f327b4fc127c8d206bca944f1141b80.

ChunMinChang:
Thanks for the help and correction about my wrong fix for resampler.c. You save the code! 
I'll try upstreaming the fixes on its own repo!

kinetiknz:
No problem, thanks for the patches! Let me know if I can help with the speexdsp upstreaming at all.

PR #205, opened 2016-12-21
Add a class to log asynchrously.

padenot:
We can only log from a single thread at the moment, we can add an MPSC ring buffer later I suppose. Also will need squashing before merging.

### DIFF
--- a/src/cubeb_log.cpp
+++ b/src/cubeb_log.cpp
@@ -1,0 +1,105 @@
+/*
+ * Copyright © 2016 Mozilla Foundation
+ *
+ * This program is made available under an ISC-style license.  See the
+ * accompanying file LICENSE for details.
+ */
+#define NOMINMAX
+
+#include "cubeb_ringbuffer.h"
+#include "cubeb_log.h"
+#include <cstdarg>
+
+cubeb_log_level g_log_level;
+cubeb_log_callback g_log_callback;
+
+/** The maximum size of a log message, after having been formatted. */
+const size_t CUBEB_LOG_MESSAGE_MAX_SIZE = 256;
+/** The maximum number of log messages that can be queued before dropping
+ * messages. */
+const size_t CUBEB_LOG_MESSAGE_QUEUE_DEPTH = 40;
+/** Number of milliseconds to wait before dequeuing log messages. */
+const std::chrono::milliseconds CUBEB_LOG_BATCH_PRINT_INTERVAL_MS =
+                                              std::chrono::milliseconds(10);
+
+/**
+  * This wraps an inline buffer, that represents a log message, that must be
+  * null-terminated.
+  * This class should not use system calls or other potentially blocking code.
+  */
+class cubeb_log_message
+{
+public:
+  cubeb_log_message()
+  {
+    PodZero(storage, CUBEB_LOG_MESSAGE_MAX_SIZE);
+  }
+  cubeb_log_message(char str[CUBEB_LOG_MESSAGE_MAX_SIZE])
+  {
+    size_t length = strlen(str);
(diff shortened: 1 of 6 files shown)


kinetiknz:
This will be useful and generally looks fine, but I think we need the MSPC queue or some solution to allow logging from any thread, otherwise this is too much of a foot-gun right now.

padenot:
I'll pick that up tomorrow, we need it to debug a real problem during calls. I'll probably remove the ALOGV statements from this patch set, and manually apply gecko-only patches in mozilla-central.

If this class is used from multiple threads, it will blow up immediately anyways. I haven't had time to do a MPSC queue, but that's still planed.

src/cubeb_log.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+
+void cubeb_async_log(const char * fmt, ...)
+{
+  // This is going to copy a 256 bytes array around, which is fine.
+  // We don't want to allocate memory here, because this is made to
+  // be called from a real-time callback.
+  va_list args;
+  va_start(args, fmt);
+  char msg[CUBEB_LOG_MESSAGE_MAX_SIZE];
```
padenot: This is just a stack pointer increment, no big deal.

src/cubeb_log.cpp
```diff
@@ -0,0 +1,105 @@
...
+  vsnprintf(msg, CUBEB_LOG_MESSAGE_MAX_SIZE, fmt, args);
+  cubeb_async_logger::get().push(msg);
```
padenot: You're right, I'll fix this.

padenot:
@kinetiknz, in light of https://github.com/kinetiknz/cubeb/pull/205#issuecomment-285120410, can we reconsider merging this? As said previously, I need it to debug a nasty clock drift issue when doing WebRTC calls.

Or maybe it's better to apply this as a patch in `mozilla-central`, considering I'm also going to have invasive logging inside the resampler anyways. What do you think ?

src/cubeb_log.cpp
```diff
@@ -0,0 +1,108 @@
...
+        while (msg_queue.dequeue(&msg, 1)) {
+          LOGV("%s", msg.get());
+        }
+        std::this_thread::sleep_for(CUBEB_LOG_BATCH_PRINT_INTERVAL_MS);
+      }
+    }).detach();
+  }
+private:
+  cubeb_async_logger()
+    :msg_queue(CUBEB_LOG_MESSAGE_QUEUE_DEPTH)
```
kinetiknz: Space after :

src/cubeb_log.cpp
```diff
@@ -0,0 +1,108 @@
...
+  }
+  /** This is quite a big data structure, but is only instantiated if the
+   * asynchronous logger is used.*/
+  lock_free_queue<cubeb_log_message> msg_queue;
+};
+
+
+void cubeb_async_log(const char * fmt, ...)
+{
+  if (g_log_callback) {
```
kinetiknz: !g_log_callback?

src/cubeb_log.h
```diff
@@ -8,6 +8,8 @@
 #ifndef CUBEB_LOG
 #define CUBEB_LOG
 
+#include "cubeb/cubeb.h"
```
kinetiknz: Why's this needed?

padenot: `cubeb_log_level`, `cubeb_log_callback`, and the values of `cubeb_log_level`.

padenot:
This takes care of your comments.

kinetiknz:
Merged in d974ad6b5fb71bae0c0089cbeb54645dfe4ecbc9 with minor fixes.

PR #207, opened 2016-12-23
Multiple channel support on Linux

ChunMinChang:
- Convert ```cubeb_channel_layout``` to ```pa_channel_map```
- Set ```pa_channel_map``` in ```pa_stream_new```
- Add ```pulse_get_preferred_channel_layout```
- Convert ```pa_channel_map``` to ```cubeb_channel_layout```

### DIFF
--- a/src/cubeb_mixer.h
+++ b/src/cubeb_mixer.h
@@ -9,6 +9,7 @@
 #define CUBEB_MIXER
 
 #include "cubeb/cubeb.h" // for cubeb_channel_layout ,CUBEB_CHANNEL_LAYOUT_MAPS and cubeb_stream_params.
+#include <stdbool.h>
 
 #if defined(__cplusplus)
 extern "C" {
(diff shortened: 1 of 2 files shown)


ChunMinChang:
The pulseAudio already has its own [mixing mechanism[0]](https://github.com/ChunMinChang/pulseaudio/blob/1d5dfccbb2b72abca6989a1c85b92e3e6e23bded/src/pulsecore/resampler.c#L841-L900), so we could leave the up/downmix for now. However, the mixing mechanism is different from the spec for 5.1-to-stereo, so I'll try to upstream some patches to pulseAudio.

[0] https://github.com/ChunMinChang/pulseaudio/blob/1d5dfccbb2b72abca6989a1c85b92e3e6e23bded/src/pulsecore/resampler.c#L841-L900

kinetiknz:
Thanks!

PR #208, opened 2016-12-23
audiounit: Avoid AudioCore dealock in stream init (Bug 1317234)

achronop:
AudioCore is prompt to deadlock sporadically. In order to avoid that the following enhancements took place: 

- I added synchronization on stream init/start/stop/destroy.
- Avoid to set property from notification callback.
- Re init stream through a asynchronous dispatch queue.
- Wait after setting buffer size for the change to take effect.

Also some refactoring took place.

Try run on the top of latest commit:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=205781e0ae4e04011fb27d0cdbf8f900b5e00b62

Please find more info about the deadlock on the corresponding bug: 
https://bugzilla.mozilla.org/show_bug.cgi?id=1317234

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -44,6 +44,8 @@ typedef UInt32  AudioFormatFlags;
     LOG("System call failed: %s (rv: %d)", str, (int) r);       \
   } while(0)
 
+#define DISPATC_QUEUE_LABEL "com.mozilla.cubeb"
+
 /* Testing empirically, some headsets report a minimal latency that is very
  * low, but this does not work in practice. Lie and say the minimum is 256
  * frames. */
@@ -66,6 +68,7 @@ struct cubeb {
   /* Differentiate input from output devices. */
   cubeb_device_type collection_changed_devtype = CUBEB_DEVICE_TYPE_UNKNOWN;
   std::vector<AudioObjectID> devtype_device_array;
+  dispatch_queue_t serial_queue = dispatch_queue_create(DISPATC_QUEUE_LABEL, nullptr);
 };
 
 struct auto_array_wrapper {
@@ -623,12 +626,15 @@ audiounit_property_listener_callback(AudioObjectID /* id */, UInt32 address_coun
     }
   }
 
-  // Switch to the new device
-  if (audiounit_reinit_stream(stm, was_running) != CUBEB_OK) {
-    stm->state_callback(stm, stm->user_ptr, CUBEB_STATE_STOPPED);
-    LOG("(%p) Could not reopen the stream after switching.", stm);
-  }
-  stm->switching_device = false;
+  // Use a new thread, through the queue, to avoid deadlock when calling
+  // Get/SetProperties method from inside notify callback
+  dispatch_async(stm->context->serial_queue, ^() {
+    if (audiounit_reinit_stream(stm, was_running) != CUBEB_OK) {
+      stm->state_callback(stm, stm->user_ptr, CUBEB_STATE_STOPPED);
+      LOG("(%p) Could not reopen the stream after switching.", stm);
+    }
+    stm->switching_device = false;
+  });
 
   return noErr;
 }
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -180,6 +180,7 @@ struct cubeb_stream {
   std::atomic<int> output_callback_in_a_row{ 0 };
   /* This is true if a device change callback is currently running.  */
   std::atomic<bool> switching_device{ false };
+  std::atomic<bool> buffer_size_change_state{ false };
 };
 
 bool has_input(cubeb_stream * stm)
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -1514,6 +1514,7 @@ audiounit_stream_init(cubeb * context,
     stm->output_device = reinterpret_cast<uintptr_t>(output_device);
   }
 
+  auto_lock context_locl(context->mutex);
```
padenot: `s/context_locl/context_lock/` ?

src/cubeb_audiounit.cpp
```diff
@@ -1285,6 +1286,133 @@ audiounit_configure_input(cubeb_stream * stm)
   return CUBEB_OK;
 }
 
+static bool
+can_change_buffer_size_to(cubeb_stream * stm, uint32_t * new_buffer_size)
```
padenot: Maybe drop the `to` here ? It sounds weird.

src/cubeb_audiounit.cpp
```diff
@@ -1285,6 +1286,133 @@ audiounit_configure_input(cubeb_stream * stm)
...
+{
+  uint32_t output_buffer_frames = 0;
+  UInt32 size = sizeof(output_buffer_frames);
+  // Use the queue in order to sleep in queue's thread.
```
padenot: s/in queue's/in the queue's/

src/cubeb_audiounit.cpp
```diff
@@ -1285,6 +1286,133 @@ audiounit_configure_input(cubeb_stream * stm)
...
+    if (r != noErr) {
+      PRINT_ERROR_CODE("AudioUnitAddPropertyListener/output/kAudioDevicePropertyBufferFrameSize", r);
+    }
+
+    return CUBEB_ERROR;
+  }
+
+  int count = 0;
+  while (!stm->buffer_size_change_state && count++ < 30) {
+    // Wait for the change to take effect without pausing the "main" thread
```
padenot: What is the "main" thread here?

src/cubeb_audiounit.cpp
```diff
@@ -1285,6 +1286,133 @@ audiounit_configure_input(cubeb_stream * stm)
...
+    dispatch_sync(stm->context->serial_queue, ^(){usleep(100000);});
```
padenot: `usleep(3)` is deprecated, use `nanosleep(2)` instead, and its behaviour is clearly defined when signals are involved.

src/cubeb_audiounit.cpp
padenot: Where is the 30 coming from ? Does it make sense to have a limit ?

src/cubeb_audiounit.cpp
```diff
@@ -1285,6 +1286,133 @@ audiounit_configure_input(cubeb_stream * stm)
...
+  if (stm->context->active_streams == 1) {
+    return true;
+  }
+
+  *new_buffer_size = output_buffer_frames;
+  return false;
+}
+
+static void
+buffer_size_changed_callback(void * inClientData,
```
padenot: This could use a high level comment describing the general strategy: 
- register a listener for the buffer size property
- change the property
- wait a bit
- the property has been changed
- remove the listener for the buffer size change

src/cubeb_audiounit.cpp
padenot: Here, you block the calling thread and the `serial_queue` thread, right ? Maybe it would be best to simply use a condition variable ?

src/cubeb_audiounit.cpp
```diff
@@ -1189,6 +1189,160 @@ audiounit_clamp_latency(cubeb_stream * stm, uint32_t latency_frames)
...
+            au_type, new_buffer_size, inScope);
+      }
+      stm->buffer_size_change_state = true;
+      break;
+    }
+  }
+}
+
+static int
+audiounit_set_buffer_size(cubeb_stream * stm, uint32_t new_size_frames, bool is_input = false)
```
padenot: Don't use boolean argument, that's impossible to read on the callsite. Have an INPUT / OUTPUT enum instead.

src/cubeb_audiounit.cpp
```diff
@@ -1225,15 +1379,9 @@ audiounit_configure_input(cubeb_stream * stm)
...
-  LOG("(%p) Input buffer frame count %u.", stm, unsigned(stm->input_buffer_frames));
-  r = AudioUnitSetProperty(stm->input_unit,
-                           kAudioDevicePropertyBufferFrameSize,
-                           kAudioUnitScope_Output,
-                           AU_IN_BUS,
-                           &stm->input_buffer_frames,
-                           sizeof(UInt32));
-  if (r != noErr) {
-    PRINT_ERROR_CODE("AudioUnitSetProperty/input/kAudioDevicePropertyBufferFrameSize", r);
+  r = audiounit_set_buffer_size(stm, stm->input_buffer_frames, true);
```
padenot: See how it's hard to understand what the `true` means here ?

src/cubeb_audiounit.cpp
padenot: Also the default value is probably unnecessary, there are going to be a couple callsites for this.

src/cubeb_audiounit.cpp
```diff
@@ -535,19 +535,44 @@ audiounit_get_input_device_id(AudioDeviceID * device_id)
   return CUBEB_OK;
 }
 
+static int
+audiounit_reinit_stream(cubeb_stream * stm, bool is_started)
+{
+  // This means the callback won't be called again.
```
achronop: I think we can live without that comment, the code is clean enough. Nevertheless stop method guarantees that user audio callback will not be called after it has returned.

src/cubeb_audiounit.cpp
```diff
@@ -66,6 +68,7 @@ struct cubeb {
   /* Differentiate input from output devices. */
   cubeb_device_type collection_changed_devtype = CUBEB_DEVICE_TYPE_UNKNOWN;
   std::vector<AudioObjectID> devtype_device_array;
+  dispatch_queue_t serial_queue = dispatch_queue_create(DISPATC_QUEUE_LABEL, nullptr);
```
achronop: The queue is asynchronously deallocated once all references to it are released. This happens automatically on destroy. `dispatch_release` is used if you want to to deallocated the object before that in case you are sure you do not need it any more.

src/cubeb_audiounit.cpp
achronop: By "main" thread I meant the thread that executes `cubeb_stream_init`. It's not clear I will change it.

The magic number is an arbitrary value big enough to ensure that there is a failure and there no point to wait any more. Better to return a failure in that case rather than waiting for ever.

src/cubeb_audiounit.cpp
```diff
@@ -1234,15 +1243,20 @@ buffer_size_changed_callback(void * inClientData,
   }
 }
 
+enum set_buffer_size_side {
+  set_input,
+  set_output,
+};
```
padenot: Constants (stand-alone and part of an enum) are usually in all caps.

PR #209, opened 2017-01-11
Define KSAUDIO_SPEAKER_5/7POINT1_SURROUND if it doesn't exist

ChunMinChang:
When I try to integrate the multiple channel support of cubeb on Windows, the error says: ```KSAUDIO_SPEAKER_5POINT1_SURROUND is not defined```, so I think it's better to define it in cubeb_wasapi.cpp itself.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -400,6 +400,17 @@ double stream_to_mix_samplerate_ratio(cubeb_stream_params & stream, cubeb_stream
 
 /* Convert the channel layout into the corresponding KSAUDIO_CHANNEL_CONFIG.
    See more: https://msdn.microsoft.com/en-us/library/windows/hardware/ff537083(v=vs.85).aspx */
+#ifndef KSAUDIO_SPEAKER_5POINT1_SURROUND
+#define KSAUDIO_SPEAKER_5POINT1_SURROUND (SPEAKER_FRONT_LEFT | SPEAKER_FRONT_RIGHT | \
+                                          SPEAKER_FRONT_CENTER | SPEAKER_LOW_FREQUENCY | \
+                                          SPEAKER_SIDE_LEFT  | SPEAKER_SIDE_RIGHT)
+#endif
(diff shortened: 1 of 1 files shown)


padenot:
Chun-Min, in what configuration does that happen ? Looks good in any case.

kinetiknz:
This shouldn't be necessary and I'd prefer not to add it unless we understand why.  Looking at `ksmedia.h` where these are defined, the `_SURROUND` variants are only defined for XP SP2 and up, but that shouldn't be a problem for us since that's all we support in Gecko anyway.

kinetiknz:
Digging in to this further, using https://msdn.microsoft.com/en-us/library/windows/desktop/aa383745(v=vs.85).aspx as a reference...

> the `_SURROUND` variants are only defined for XP SP2 and up

More specifically, they're defined for `NTDDI_VERSION` `0x05010200` (XP SP2) up to but not including `0x05020000` (WS03) and then from `0x05020100` (WS03 SP1) onwards.

`NTDDI_VERSION ` is derived from `_WIN32_WINNT` in `sdkddkver.h` in the Windows SDK, and _WIN32_WINNT is set to `0x0502` (`_WIN32_WINNT_WS03` aka "Windows Server 2003 with SP1, Windows XP with SP2") in the Gecko build.

The trap is... `sdkddkver.h` derives `NTDDI_VERSION` by simply appending `0000` to the value of `_WIN32_WINNT`, so despite `_WIN32_WINNT_WS03` claiming to represent XP SP2 and WS03 SP1, the `NTDDI_VERSION` derived will correspond to `NTDDI_WS03,` which explicitly excludes the required `_SURROUND` definitions.

We don't hit this on AppVeyor because libcubeb's own build system uses the SDK defaults, which is the latest Windows version supported by the selected SDK.

So I think the simplest fix is to add `#define NTDDI_VERSION 0x05020100` to the top of cubeb_wasapi.cpp.

ChunMinChang:
Thanks for the explicit diagnosis. I've modified the code by your suggestion. Thanks : )

kinetiknz:
Thanks! I added a short comment when I merged the commit.

PR #210, opened 2017-01-16
Prevent double-free when doing an emergency bailout from the rendering thread

padenot:
This caused gecko bug 1326176.

This was caused by the fact that we would null out `stm->thread` when in
fact it was still running, so we would delete `stm->emergency_bailout`
twice, because we would return true from `stop_and_join_thread`.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -1230,13 +1230,18 @@ bool stop_and_join_render_thread(cubeb_stream * stm)
     rv = false;
   }
 
-  LOG("Closing thread.");
 
-  CloseHandle(stm->thread);
-  stm->thread = NULL;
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -1225,13 +1225,18 @@ bool stop_and_join_render_thread(cubeb_stream * stm)
     rv = false;
   }
 
-  LOG("Closing thread.");
 
-  CloseHandle(stm->thread);
-  stm->thread = NULL;
+  // Only attempt to close and null out the thread
+  // and event if the WaitForSingleObject above succeeded,
+  // so that calling this function again attemps to close
```
kinetiknz: Typo in attempts, and seems incomplete after "close"?

padenot:
I've fixed the comment, thanks !

PR #212, opened 2017-01-19
Multiple channel support on OS X

ChunMinChang:
This is the implementation for audio 5.1 support on OS X. It only support the output side. If we also need to implement the input side, I'll add later.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -182,6 +182,11 @@ private:
   auto_array<T> ar;
 };
 
+enum io_side {
+  INPUT,
+  OUTPUT,
+};
+
 struct cubeb_stream {
   explicit cubeb_stream(cubeb * context);
 
@@ -1127,7 +1132,7 @@ audio_stream_desc_init(AudioStreamBasicDescription * ss,
 
 static int
 audiounit_layout_init(AudioUnit * unit,
-                      bool is_input,
+                      io_side side,
                       const cubeb_stream_params * stream_params)
 {
   assert(stream_params->layout != CUBEB_LAYOUT_UNDEFINED);
@@ -1184,8 +1189,8 @@ audiounit_layout_init(AudioUnit * unit,
 
   r = AudioUnitSetProperty(*unit,
                            kAudioUnitProperty_AudioChannelLayout,
-                           is_input ? kAudioUnitScope_Output : kAudioUnitScope_Input,
-                           is_input ? AU_IN_BUS : AU_OUT_BUS,
+                           side == INPUT ? kAudioUnitScope_Output : kAudioUnitScope_Input,
+                           side == INPUT ? AU_IN_BUS : AU_OUT_BUS,
                            layout.get(),
                            layout.size());
   if (r != noErr) {
@@ -1418,20 +1423,15 @@ buffer_size_changed_callback(void * inClientData,
   }
 }
 
-enum set_buffer_size_side {
-  INPUT,
-  OUTPUT,
-};
-
 static int
-audiounit_set_buffer_size(cubeb_stream * stm, uint32_t new_size_frames, set_buffer_size_side set_side)
+audiounit_set_buffer_size(cubeb_stream * stm, uint32_t new_size_frames, io_side side)
 {
   AudioUnit au = stm->output_unit;
   AudioUnitScope au_scope = kAudioUnitScope_Input;
   AudioUnitElement au_element = AU_OUT_BUS;
   const char * au_type = "output";
 
-  if (set_side == INPUT) {
+  if (side == INPUT) {
     au = stm->input_unit;
     au_scope = kAudioUnitScope_Output;
     au_element = AU_IN_BUS;
@@ -1447,7 +1447,7 @@ audiounit_set_buffer_size(cubeb_stream * stm, uint32_t new_size_frames, set_buff
                                &buffer_frames,
                                &size);
   if (r != noErr) {
-    if (set_side == INPUT) {
+    if (side == INPUT) {
       PRINT_ERROR_CODE("AudioUnitGetProperty/input/kAudioDevicePropertyBufferFrameSize", r);
     } else {
       PRINT_ERROR_CODE("AudioUnitGetProperty/output/kAudioDevicePropertyBufferFrameSize", r);
@@ -1465,7 +1465,7 @@ audiounit_set_buffer_size(cubeb_stream * stm, uint32_t new_size_frames, set_buff
                                    buffer_size_changed_callback,
                                    stm);
   if (r != noErr) {
-    if (set_side == INPUT) {
+    if (side == INPUT) {
       PRINT_ERROR_CODE("AudioUnitAddPropertyListener/input/kAudioDevicePropertyBufferFrameSize", r);
     } else {
       PRINT_ERROR_CODE("AudioUnitAddPropertyListener/output/kAudioDevicePropertyBufferFrameSize", r);
@@ -1482,7 +1482,7 @@ audiounit_set_buffer_size(cubeb_stream * stm, uint32_t new_size_frames, set_buff
                            &new_size_frames,
                            sizeof(new_size_frames));
   if (r != noErr) {
-    if (set_side == INPUT) {
+    if (side == INPUT) {
       PRINT_ERROR_CODE("AudioUnitSetProperty/input/kAudioDevicePropertyBufferFrameSize", r);
     } else {
       PRINT_ERROR_CODE("AudioUnitSetProperty/output/kAudioDevicePropertyBufferFrameSize", r);
@@ -1493,7 +1493,7 @@ audiounit_set_buffer_size(cubeb_stream * stm, uint32_t new_size_frames, set_buff
                                                     buffer_size_changed_callback,
                                                     stm);
     if (r != noErr) {
-      if (set_side == INPUT) {
+      if (side == INPUT) {
         PRINT_ERROR_CODE("AudioUnitAddPropertyListener/input/kAudioDevicePropertyBufferFrameSize", r);
       } else {
         PRINT_ERROR_CODE("AudioUnitAddPropertyListener/output/kAudioDevicePropertyBufferFrameSize", r);
@@ -1519,7 +1519,7 @@ audiounit_set_buffer_size(cubeb_stream * stm, uint32_t new_size_frames, set_buff
                                                   buffer_size_changed_callback,
                                                   stm);
   if (r != noErr) {
-    if (set_side == INPUT) {
+    if (side == INPUT) {
       PRINT_ERROR_CODE("AudioUnitAddPropertyListener/input/kAudioDevicePropertyBufferFrameSize", r);
     } else {
       PRINT_ERROR_CODE("AudioUnitAddPropertyListener/output/kAudioDevicePropertyBufferFrameSize", r);
@@ -1713,7 +1713,7 @@ audiounit_configure_output(cubeb_stream * stm)
     return CUBEB_ERROR;
   }
 
-  audiounit_layout_init(&stm->output_unit, false, &stm->output_stream_params);
+  audiounit_layout_init(&stm->output_unit, OUTPUT, &stm->output_stream_params);
 
   LOG("(%p) Output audiounit init successfully.", stm);
   return CUBEB_OK;
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -2720,14 +2720,13 @@ audiounit_get_channel_layout(bool preferred)
 
   // Get the channel layout
   UInt32 size = 0;
-  Boolean writable;
   rv = AudioUnitGetPropertyInfo(output_unit,
                                 preferred ? kAudioDevicePropertyPreferredChannelLayout :
                                             kAudioUnitProperty_AudioChannelLayout,
                                 kAudioUnitScope_Output,
                                 AU_OUT_BUS,
                                 &size,
-                                &writable);
+                                nullptr);
   if (rv != noErr) {
     if (preferred) {
       PRINT_ERROR_CODE("AudioUnitGetPropertyInfo/kAudioDevicePropertyPreferredChannelLayout", rv);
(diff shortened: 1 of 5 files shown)


padenot:
@achronop, I'll have a look as well, but you should look at this.

src/cubeb_audiounit.cpp
```diff
@@ -83,6 +88,53 @@ struct auto_array_wrapper {
...
+class auto_channel_layout
+{
+public:
+  auto_channel_layout()
+    : layout(nullptr)
+  {
+  }
+
+  auto_channel_layout(size_t size)
+    : layout(reinterpret_cast<AudioChannelLayout*>(malloc(size)))
```
ChunMinChang: The size depends on the channel counts of the connected audio device, so we don't know what's the size of the used AudioChannelLayout before asking via CoreAudio API.

Or do you mean ```void* operator new (std::size_t size)```? AFAIK, we still need to initialize the member value if we use it.

src/cubeb_audiounit.cpp
```diff
@@ -83,6 +88,53 @@ struct auto_array_wrapper {
...
+  size_t size()
+  {
+    return sizeof(*layout);
+  }
+
+private:
+  void release()
+  {
+    if (layout) {
+      free(layout);
```
ChunMinChang: OK. Thanks for the reminder : )

src/cubeb_audiounit.cpp
```diff
@@ -958,6 +1046,22 @@ audiounit_get_preferred_sample_rate(cubeb * /* ctx */, uint32_t * rate)
...
+  // (e.g. ASUS Xonar U7), has preferred layout setting.
+  // For default output on Mac, there is no preferred channel layout,
+  // so it might return UNDEFINED.
+  // In that case, we should get the channel configuration directly.
+  *layout = audiounit_get_channel_layout(true);
+  if (*layout == CUBEB_LAYOUT_UNDEFINED) {
+    *layout = audiounit_get_channel_layout(false);
+  }
+
+  return CUBEB_OK;
```
ChunMinChang: Hmm... sometime we can still set the channel layout if it's predefined layout is ```UNDEFINED```.

We only support ```SMPTE``` layout for now. The SMPTE's channel order for audio 5.1 is _Left, Right, Center, Low Frequency Effect, Left Surround, Right Surround_. However, if it's predefined layout is not in SMPTE's format(e.g., _Left, Right, Left Surround, Right Surround, Center, Low Frequency Effect_), then we will get ```UNDEFINED```.  

However, we still able to set its layout into SMPTE's format by CoreAudio's API. In such case, will it confuse users if we return ```CUBEB_ERROR``` when the device can be set and play sound well?

src/cubeb_audiounit.cpp
```diff
@@ -1022,6 +1126,77 @@ audio_stream_desc_init(AudioStreamBasicDescription * ss,
 }
 
 static int
+audiounit_layout_init(AudioUnit * unit,
+                      bool is_input,
```
ChunMinChang: Good idea! I'll reuse the ```enum set_buffer_size_side``` to do that.

src/cubeb_audiounit.cpp
```diff
@@ -1022,6 +1126,77 @@ audio_stream_desc_init(AudioStreamBasicDescription * ss,
...
+                      const cubeb_stream_params * stream_params)
+{
+  assert(stream_params->layout != CUBEB_LAYOUT_UNDEFINED);
+  assert(stream_params->channels == CUBEB_CHANNEL_LAYOUT_MAPS[stream_params->layout].channels);
+
+  OSStatus r;
+  auto_channel_layout layout(sizeof(AudioChannelLayout));
+
+  switch (stream_params->layout) {
+    case CUBEB_LAYOUT_DUAL_MONO:
```
ChunMinChang: I think the ```CUBEB_LAYOUT_DUAL_MONO``` is different from ```MONO```. In SMPTE's spec, [dual mono](http://searchfox.org/mozilla-central/source/dom/media/MediaInfo.cpp#19) is same as __stereo__.

src/cubeb_audiounit.cpp
```diff
@@ -1022,6 +1126,77 @@ audio_stream_desc_init(AudioStreamBasicDescription * ss,
...
+  }
+
+  r = AudioUnitSetProperty(*unit,
+                           kAudioUnitProperty_AudioChannelLayout,
+                           is_input ? kAudioUnitScope_Output : kAudioUnitScope_Input,
+                           is_input ? AU_IN_BUS : AU_OUT_BUS,
+                           layout.get(),
+                           layout.size());
+  if (r != noErr) {
+    PRINT_ERROR_CODE("AudioUnitSetProperty/kAudioUnitProperty_AudioChannelLayout", r);
```
ChunMinChang: Thanks for the reminder! I'll add that.

src/cubeb_audiounit.cpp
```diff
@@ -2528,13 +2705,69 @@ int audiounit_register_device_collection_changed(cubeb * context,
...
+  // Get the channel layout
+  UInt32 size = 0;
+  Boolean writable;
+  rv = AudioUnitGetPropertyInfo(output_unit,
+                                (preferred) ? kAudioDevicePropertyPreferredChannelLayout :
+                                              kAudioUnitProperty_AudioChannelLayout,
+                                kAudioUnitScope_Output,
+                                AU_OUT_BUS,
+                                &size,
+                                &writable);
```
ChunMinChang: OK. I'll replace it.

src/cubeb_audiounit.cpp
```diff
@@ -2528,13 +2705,69 @@ int audiounit_register_device_collection_changed(cubeb * context,
...
+  if (rv != noErr) {
+    PRINT_ERROR_CODE("AudioUnitGetPropertyInfo/kAudioDevicePropertyPreferredChannelLayout||kAudioUnitProperty_AudioChannelLayout", rv);
```
ChunMinChang: OK, I will add that info.

ChunMinChang:
@achronop 
Thanks for your review :)
Travis CI shows some error before I rebase the commits, so I am afraid that some review comments may be relocated after rebasing. Sorry for that.

src/cubeb_audiounit.cpp
achronop: No `malloc` is fine, since we cannot use normal `new`using `operator new` instead of `malloc` does not make any difference to me.

src/cubeb_audiounit.cpp
achronop: Let me chat more about this one. If cubeb does not offer a mechanism to set a non-SMPTE layout I am not sure if you should care what the user can do by using the CoreAudio API directly.

Moreover, even if we allow it, again the assert in line 1141 (1133 before) will be hit. The assert there says that we never expect an `UNDEFINED` layout which is not the case always when we use the  `audiounit_get_channel_layout()` method.

src/cubeb_audiounit.cpp
achronop: Right I did not notice the `_DUAL_`.

src/cubeb_audiounit.cpp
ChunMinChang: OK. If it will confuse users(they may think that set ```UNDEFINED``` to ```stream_init``` is ok) without returning ```CUBEB_ERROR```, I can change it.

The current returned value of ```cubeb_get_preferred_channel_layout``` on Windows and Linux should also be modified. I will push another requests to do that.

test/test_latency.cpp
```diff
@@ -27,7 +27,6 @@ TEST(cubeb, latency)
   }
 
   r = cubeb_get_preferred_channel_layout(ctx, &layout);
-  ASSERT_TRUE(r == CUBEB_OK || r == CUBEB_ERROR_NOT_SUPPORTED);
```
achronop: Why did you remove the assert? Do you hit the assert after the change above?

ChunMinChang: The assert will be hit when applying same behavior(return ```ERROR``` when the preferred layout is ```UNDEFINED```).

Actually, the assert on OSX could be passed after the change. To keep same behavior of ```cubeb_get_preferred_channel_layout``` on every backend, the implementation of getting preferred layout on Windows and Linux also need to return ```ERROR``` when the preferred layout is ```UNDEFINED```. However, the preferred 5.1 layout on my __Ubuntu__ is _Left, Right, Center, Low Frequency Effect, Rear Left Surround, Rear Right Surround_ instead of SMPTE's _Left, Right, Center, Low Frequency Effect, Left Surround, Right Surround_, so it will get ```UNDEFINED```.

I can leave the assert for now, but it will __still__ be removed after changing the implementation of getting preferred layout on Windows and Linux.

achronop: > However, the preferred 5.1 layout on my Ubuntu is Left, Right, Center, Low Frequency Effect, Rear Left Surround, Rear Right Surround instead of SMPTE's Left, Right, Center, Low Frequency Effect, Left Surround, Right Surround, so the preferred layout is UNDEFINED.

So, does the test pass on your laptop without the return error change? My understanding is that it should hit the assert in cubeb_pulse.c:800. Am I correct?

ChunMinChang: @achronop : Sorry for late reply.
I think I don't explain the code well. Let me try again : )

If we need to return ```CUBEB_ERROR``` when system's preferred layout is ```CUBEB_LAYOUT_UNDEFINED```(non-SMPTE spec), then I will need to add following code after [cubeb_pulse.c:719](https://github.com/kinetiknz/cubeb/blob/c7088a13c0288dbcd94ef7f312ab6a6cb66f8faa/src/cubeb_pulse.c#L719)
```c
if (*layout == CUBEB_LAYOUT_UNDEFINED) {
    return CUBEB_ERROR;
}
```

Then, it will hit the assertion in [test_latency.cpp:30](https://github.com/kinetiknz/cubeb/blob/068ba2f1d8c1ec941cce374ed125c0150ea513d3/test/test_latency.cpp#L30) on my ubuntu
```cpp
r = cubeb_get_preferred_channel_layout(ctx, &layout);
ASSERT_TRUE(r == CUBEB_OK || r == CUBEB_ERROR_NOT_SUPPORTED);
```
since the preferred 5.1 layout on my Ubuntu is _Left, Right, Center, Low Frequency Effect, Rear Left Surround, Rear Right Surround_, instead of SMPTE's 5.1 layout: _Left, Right, Center, Low Frequency Effect, Left Surround, Right Surround_. Thus, the returned preferred layout of ```cubeb_get_preferred_channel_layout```  is ```CUBEB_LAYOUT_UNDEFINED``` on my ubuntu. That's why it hit the assertion.

On the other hand, the same 5.1 audio device on my OS X doesn't hit any assertion because it's returned preferred layout of ```cubeb_get_preferred_channel_layout```  is ```CUBEB_LAYOUT_3F2_LFE```. It's exactly what we want. Nevertheless, AFAIK, the preferred channel layout is __different from device(driver) to device(driver)__, so it's possible that [```ASSERT_TRUE(r == CUBEB_OK...``` in test_latency.cpp](https://github.com/kinetiknz/cubeb/pull/212/files/b211071a2f4f9435c6eb00ea95ac603e38aed5c9#diff-5bf2cbe3856cd9ebcaf0ec2d8b051a98L30)
```
ASSERT_TRUE(r == CUBEB_OK || r == CUBEB_ERROR_NOT_SUPPORTED);
```
will be hit if we [return ```ERROR``` when the preferred layout is ```UNDEFINED```](https://github.com/kinetiknz/cubeb/pull/212/files/b211071a2f4f9435c6eb00ea95ac603e38aed5c9#diff-148fe3e6ebf13351ba4d886b1b4898d2R1070)
```cpp
if (*layout == CUBEB_LAYOUT_UNDEFINED) {
  return CUBEB_ERROR;
}
```
when the device's default layout is non-SMPTE spec.

However, I am OK to keep ```ASSERT_TRUE(r == CUBEB_OK || r == CUBEB_ERROR_NOT_SUPPORTED);``` in ```test_latency.cpp``` for now and change it when we face a real use case.

achronop: Thank you for the detailed explanation. I understand why you removed the assert in test_latency.cpp and it's fine. 

Let me make a different question just to clarify the case:
In your Linux box if you execute the following code, as it is right now:
```
r = cubeb_get_preferred_channel_layout(ctx, &layout); 
// As you said returned layout will be CUBEB_LAYOUT_UNDEFINED, r will be CUBEB_OK 
ASSERT_TRUE(r == CUBEB_OK || r == CUBEB_ERROR_NOT_SUPPORTED);
// Feed that layout to the stream params
cubeb_stream_params params = { CUBEB_SAMPLE_FLOAT32NE, preferred_rate, max_channels, layout};
// Use the stream params to init a stream
cubeb_stream_init(ctx, &stream, "test", NULL, NULL, NULL, &params,
512, data_cb, state_cb, nullptr);
```
Will that code hit the assert in cubeb_pulse.c:800 ?

ChunMinChang: Yes, it will hit the assert.

We restrict that the channel layout must match with the channel counts. Thus, if user want to use ```cubeb_get_preferred_channel_layout``` to get the _layout_ as input parameter of ```cubeb_stream_init```, it means that the user only know audio's channel counts but have no knowledge of its layout. In that case, ```cubeb_get_preferred_channel_layout``` could __not__ help user to get the audio's layout information they want. It only indicates the device's preferred layout, instead of the audio data's layout.

To warn user that the ```cubeb_get_preferred_channel_layout``` does not always return a valid layout, we could add following code in test_latency.cpp.
```cpp
ASSERT_TRUE(r == CUBEB_OK || r == CUBEB_ERROR_NOT_SUPPORTED || r == CUBEB_ERROR);
if (r == CUBEB_ERROR) {
  ASSERT_EQ(layout, CUBEB_LAYOUT_UNDEFINED);
}
```

What do you think?

achronop: I agree, we can do it in one line:
ASSERT_TRUE(r == CUBEB_OK || r == CUBEB_ERROR_NOT_SUPPORTED ||( r == CUBEB_ERROR && layout == CUBEB_LAYOUT_UNDEFINED));

So lets return CUBEB_ERROR for now when layout is UNDEFINED.

ChunMinChang: OK! Thanks for reviewing this!

achronop:
@padenot, @kinetiknz: I finished the review. I do not know if you plan to go through it. If you do not have any comment I can merge it.

kinetiknz:
If you've done a full review and everything has been addressed, I'm fine with merging.  Thanks!

PR #214, opened 2017-01-20
Move jack_init below pulse_init in cubeb.c.

kinetiknz:
Gecko only officially supports the PulseAudio backend, so it makes sense to
prefer that over JACK and ALSA, which are supported by contributors.

This is a partial fix for issues like [BMO #1331526](https://bugzilla.mozilla.org/show_bug.cgi?id=1331526).  A better long term fix
would allow overriding the backend selection with requests for a specific
backend.

### DIFF
--- a/src/cubeb.c
+++ b/src/cubeb.c
@@ -108,11 +108,11 @@ int
 cubeb_init(cubeb ** context, char const * context_name)
 {
   int (* init[])(cubeb **, char const *) = {
-#if defined(USE_JACK)
-    jack_init,
-#endif
 #if defined(USE_PULSE)
     pulse_init,
+#endif
+#if defined(USE_JACK)
+    jack_init,
 #endif
 #if defined(USE_ALSA)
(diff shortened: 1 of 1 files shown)


kinetiknz:
r? padenot

h1z1:
How did this ever get commited?  You're compiling with jack and pulse, expecting pulse to have priority?   Would have thought the accepted fix allowed one to specify the backend at runtime.

kinetiknz:
> Would have thought the accepted fix allowed one to specify the backend at runtime.

That exists: cubeb_init takes a backend name - see https://github.com/kinetiknz/cubeb/blob/master/include/cubeb/cubeb.h#L430

h1z1:
Oh wow, I've used Firefox and Jack for quite a while, had no idea media.cubeb.backend even existed.. Nor why Mozilla insists on releases without jack given this issue is addressed?

And I see you've updated BMO, thank you!

PR #217, opened 2017-01-20
Stop inserting silence when we don't have input data and need to render some audio, on Windows

padenot:
We use the same strategy on OSX. Testing locally with a setup where I could repro a drift after some time, this fixes it: no drift in 18 hours (I let it run overnight).

The plan is to come up with a full-fledged drift compensation mechanism, but it's not finished, and this does the trick for now.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -747,15 +747,6 @@ refill_callback_duplex(cubeb_stream * stm)
     return true;
   }
 
-  // When WASAPI has not filled the input buffer yet, send silence.
-  double output_duration = double(output_frames) / stm->output_mix_params.rate;
-  double input_duration = double(stm->linear_input_buffer.length() / stm->input_stream_params.channels) / stm->input_mix_params.rate;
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -1651,6 +1642,14 @@ int setup_wasapi_stream(cubeb_stream * stm)
                                       stm->input_available_event,
                                       stm->capture_client,
                                       &stm->input_mix_params);
+
+    // We initializing an input stream, buffer ahead two buffers worth of silence.
+    // This delays the input side slightly, but allow to not glitch when no input
+    // is available when calling into the resampler to call the callback: the input
+    // refill event will be set shortly after to compensate for this lack of data.
+    stm->linear_input_buffer.push_silence(stm->input_buffer_frame_count *
+                                          stm->input_stream_params.channels * 2);
```
achronop: Why not to move the statement after checking the rv value.

Just note that, in audiounit, when there is a steady drift, the initial silence can be consumed completely in long term.

Issue #185, opened 2016-11-07
Resampler fill method unused frames "trap"


achronop:
The unused frames from method `cubeb_resampler_fill` depending on the `cubeb_resampler` instance. The `noop_resampler` resampler does not buffer any of the frames internally so it is expected to have remaining frames. On the other hand the `cubeb_resampler_speex` resampler buffers internally all input frames and uses them in following iterations as needed. In that case it is not expected to have any left over frames after fill.

I raise this issue first to let you know about the case and second to find a way to make resampler behavior more uniform.

PR #218, opened 2017-01-20
resampler: return unsed frames in all cases

achronop:
Address #185


achronop:
I have a number of fixes depending on this change. I will close this one and push them all together.

PR #221, opened 2017-01-23
Audiounit drift with c920 webcam (Bug 1329190)

achronop:
This patch is solving a build up latency when c920 webcam is used as external audio input. The problem occurs after some time (~10 mins) by getting extra output callbacks, consequently, extra silence in pushed in the input that creates the delay. In order to solve that I did the followings:
- Change the resampler not to store any remaining input frame, instead, return the number of the used frames. 
- Change audiounit backend to save the remaining input and use them in the following cycles. Moreover extra silence is pushed, if only the input is not enough and not when there are more output callbacks than expected. This does not allow the latency to grow.

Finally, in order to study the case I created and used 2 convenient log macros for normal and verbose async logging. This can be done easily with dispatch queues.

On going try push: https://treeherder.mozilla.org/#/jobs?repo=try&revision=4aa24728339713ed18b2ef07881bcbb8d124ae0b

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -54,6 +54,11 @@ const char * DISPATCH_QUEUE_LABEL = "org.mozilla.cubeb";
 #endif
 #define ALOGV(msg, ...) dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0), ^{LOGV(msg, ##__VA_ARGS__);})
 
+#ifdef ALOG
+#undef ALOG
+#endif
+#define ALOG(msg, ...) dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0), ^{LOG(msg, ##__VA_ARGS__);})
+
 /* Testing empirically, some headsets report a minimal latency that is very
  * low, but this does not work in practice. Lie and say the minimum is 256
  * frames. */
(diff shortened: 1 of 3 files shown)


achronop:
Since review has not started yet I push 2 more (small) corrections. Sorry for the inconvenience.

src/cubeb_resampler.cpp
```diff
@@ -162,7 +162,7 @@ ::fill_internal_input(T * input_buffer, long * input_frames_count,
   /* process the input, and present exactly `output_frames_needed` in the
   * callback. */
   input_processor->input(input_buffer, *input_frames_count);
-  resampled_input = input_processor->output(resampled_frame_count);
+  resampled_input = input_processor->output(resampled_frame_count, (size_t*)input_frames_count);
```
padenot: Maybe it would be better to use a reference, if we go with a design where the resamplers keep their data and don't return them with the caller.

src/cubeb_audiounit.cpp
```diff
@@ -314,12 +315,13 @@ audiounit_input_callback(void * user_ptr,
...
+  // many output (switched device etc), we can clear the input buffer now,
+  // only keeping the new rendered data.
   if (stm->output_callback_in_a_row > stm->expected_output_callbacks_in_a_row) {
-    stm->input_linear_buffer->pop(
-        stm->input_linear_buffer->length() -
-        input_frames * stm->input_stream_params.channels);
+    stm->input_linear_buffer->clear();
+    /* Paranoia: The 10 extra frames are not used but if they do not exist resampler
+     * will assert. */
+    stm->input_linear_buffer->push_silence(10 * stm->input_desc.mChannelsPerFrame);
```
padenot: Can we compute this number based on the resampling ratio ?

src/cubeb_audiounit.cpp
```diff
@@ -353,6 +355,12 @@ audiounit_input_callback(void * user_ptr,
   return noErr;
 }
 
+static uint32_t
+min_input_frames_required(cubeb_stream * stm)
```
padenot: Required for what ?

src/cubeb_resampler_internal.h
```diff
@@ -237,7 +237,8 @@ class cubeb_resampler_speex_one_way : public processor {
 
     /* This shifts back any unresampled samples to the beginning of the input
        buffer. */
-    resampling_in_buffer.pop(nullptr, frames_to_samples(in_len));
+    *input_frames_used = in_len;
+    resampling_in_buffer.clear();
```
padenot: This will probably break Windows/WASAPI. It might be best to let the frames sitting in the resampler.

src/cubeb_resampler_internal.h
```diff
@@ -385,6 +386,7 @@ class delay_line : public processor {
     delay_output_buffer.push(delay_input_buffer.data(),
                              frames_to_samples(frames_needed));
     delay_input_buffer.pop(nullptr, frames_to_samples(frames_needed));
+    *input_frames_used = frames_needed;
```
padenot: Here, you don't pop all the frames from `delay_input_buffer` like the the case where we resample, is that expected ?

achronop: No, it's not intentional, I just missed it. However since we keep the buffering in the resampler (based on previous comment) this ends to be the desired behavior.

achronop:
Great! I really need this fix! I will start a cubeb update right away.

PR #222, opened 2017-01-24
Allow using a different release assert function in cubeb.

padenot:
See [bug 1332937](https://bugzilla.mozilla.org/show_bug.cgi?id=1332937), where the crash reporter is not called.

This allow using `MOZILLA_RELEASE_ASSERT` or something like that, and integrate with the crash reporter.

@kinetiknz, do you remember why we have both a null-dereference AND a call to `abort` in `cubeb_crash` ?

### DIFF
--- a/src/cubeb-internal.h
+++ b/src/cubeb-internal.h
@@ -9,6 +9,7 @@
 
 #include "cubeb/cubeb.h"
 #include "cubeb_log.h"
+#include "cubeb_assert.h"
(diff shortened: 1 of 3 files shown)


kinetiknz:
> @kinetiknz, do you remember why we have both a null-dereference AND a call to abort in cubeb_crash ?

IIRC it was based on MOZ_CRASH at some point.

PR #223, opened 2017-02-01
Allow buffering some input when using a noop resampler, and rename it because it's not strictly no-op anymore

padenot:
@achronop and I have witnessed a number of scenarios where sometimes, instead of getting an input callback before the output callback, we have no input data when we want to render, and then right after, we get an input callback with double the amount of frames. This happens on OSX and Windows. 

Because we delay the input side by a few buffers, this works fine when resampling happens. When using a no-op resampler, because we simply forward the call to the user callback, data getting lost.

This adds a buffer in the no-op path (renamed for the occasion), and handles this case, adjusting the assert. Also, a number of tests that would have caught this are added, and another test case that simulates the scenario described above has been added.

### DIFF
--- a/src/cubeb_resampler.cpp
+++ b/src/cubeb_resampler.cpp
@@ -35,29 +35,54 @@ to_speex_quality(cubeb_resampler_quality q)
   }
 }
 
-long noop_resampler::fill(void * input_buffer, long * input_frames_count,
-                          void * output_buffer, long output_frames)
+template<typename T>
+passthrough_resampler<T>::passthrough_resampler(cubeb_stream * s,
+                                                cubeb_data_callback cb,
+                                                void * ptr,
+                                                uint32_t input_channels)
+  : processor(input_channels)
+  , stream(s)
+  , data_callback(cb)
+  , user_ptr(ptr)
+{
+}
+
+template<typename T>
+long passthrough_resampler<T>::fill(void * input_buffer, long * input_frames_count,
+                                    void * output_buffer, long output_frames)
 {
   if (input_buffer) {
     assert(input_frames_count);
   }
   assert((input_buffer && output_buffer &&
-         *input_frames_count >= output_frames) ||
+         *input_frames_count + static_cast<int>(samples_to_frames(internal_input_buffer.length())) >= output_frames) ||
          (!input_buffer && (!input_frames_count || *input_frames_count == 0)) ||
          (!output_buffer && output_frames == 0));
+
+  long rv;
 
   if (output_buffer == nullptr) {
     assert(input_buffer);
     output_frames = *input_frames_count;
   }
 
-  if (input_buffer && *input_frames_count != output_frames) {
-    assert(*input_frames_count > output_frames);
-    *input_frames_count = output_frames;
+
+  if (input_buffer) {
+    internal_input_buffer.push(static_cast<T*>(input_buffer),
+                               frames_to_samples(*input_frames_count));
+    assert(samples_to_frames(internal_input_buffer.length()) >=
+           static_cast<uint32_t>(output_frames));
   }
 
-  return data_callback(stream, user_ptr,
-                       input_buffer, output_buffer, output_frames);
+  rv = data_callback(stream, user_ptr, internal_input_buffer.data(),
+                     output_buffer, output_frames);
+
+  if (input_buffer) {
+    internal_input_buffer.pop(nullptr, frames_to_samples(output_frames));
+    printf("remaining frames in noop: %lu\n", samples_to_frames(internal_input_buffer.length()));
+  }
+
+  return rv;
 }
 
 template<typename T, typename InputProcessor, typename OutputProcessor>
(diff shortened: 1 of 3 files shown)


test/test_resampler.cpp
```diff
@@ -536,3 +543,201 @@ TEST(cubeb, resampler_drain)
...
+  ASSERT_TRUE(!!output_buffer);
+  ASSERT_TRUE(!!input_buffer);
+
+  for (int i = 0; i < frame_count; i++) {
+    // output is two channels, input is one channel, we upmix.
+    output_buffer[i] = output_buffer[i+1] = input_buffer[i];
+  }
+}
+
+long cb_noop_resampler_duplex(cubeb_stream * /*stm*/, void * /*user_ptr*/,
```
achronop: Nit: passthrough in the name

test/test_resampler.cpp
```diff
@@ -536,3 +543,201 @@ TEST(cubeb, resampler_drain)
...
+void check_duplex(const T * input_buffer,
+                  T * output_buffer, long frame_count)
+{
+  ASSERT_EQ(frame_count, 256);
+  ASSERT_TRUE(!!output_buffer);
+  ASSERT_TRUE(!!input_buffer);
+
+  for (int i = 0; i < frame_count; i++) {
+    // output is two channels, input is one channel, we upmix.
+    output_buffer[i] = output_buffer[i+1] = input_buffer[i];
```
achronop: in line 684 you have `const int output_channels = 1;`. Here you are setting 2 channels. Moreover you are setting in check method. It might be more intuitive to set inside the callback.

test/test_resampler.cpp
```diff
@@ -536,3 +543,201 @@ TEST(cubeb, resampler_drain)
...
+    long int frames = BUF_BASE_SIZE;
+    // Simulate that sometimes, we don't have the input callback on time
+    if (i != 0 && (i % 100) == 0) {
+      long zero = 0;
+      got = cubeb_resampler_fill(resampler, input_buffer_normal /* unused here */,
+                                 &zero, output_buffer, BUF_BASE_SIZE);
+    } else if (i != 0 && (i % 100) == 1) {
+      // if this is the case, the on the next iteration, we'll have twice the
+      // amount of input frames
+      seq_idx = seq(input_buffer_prebuffer, seq_idx, BUF_BASE_SIZE * 2);
```
achronop: It would be better for the reader to create a new buffer here with a better name instead of using the existing`input_buffer_prebuffer`. It would be easier to follow with a name `new_input_data_twice_long` or anything similar you can come up.

test/test_resampler.cpp
```diff
@@ -536,3 +543,201 @@ TEST(cubeb, resampler_drain)
...
+      frames = 2 * BUF_BASE_SIZE;
+      got = cubeb_resampler_fill(resampler, input_buffer_prebuffer, &frames, output_buffer, BUF_BASE_SIZE);
+    } else {
+       // normal case
+      seq_idx = seq(input_buffer_normal, seq_idx, BUF_BASE_SIZE);
+      long normal_input_frame_count = 256;
+      got = cubeb_resampler_fill(resampler, input_buffer_normal, &normal_input_frame_count, output_buffer, BUF_BASE_SIZE);
+      is_seq(output_buffer, 2, BUF_BASE_SIZE, output_seq_idx);
```
achronop: I do not know if I get it wrong but the `2` here is for the out channels, however you have configured one channel for the output.

src/cubeb_resampler.cpp
```diff
@@ -35,29 +35,54 @@ to_speex_quality(cubeb_resampler_quality q)
...
+    assert(samples_to_frames(internal_input_buffer.length()) >=
+           static_cast<uint32_t>(output_frames));
+  }
+
+  rv = data_callback(stream, user_ptr, internal_input_buffer.data(),
+                     output_buffer, output_frames);
+
+  if (input_buffer) {
+    internal_input_buffer.pop(nullptr, frames_to_samples(output_frames));
+    printf("remaining frames in noop: %lu\n", samples_to_frames(internal_input_buffer.length()));
```
padenot: I'm removing it, it's just leftover from debugging.

src/cubeb_resampler.cpp
```diff
@@ -35,29 +35,54 @@ to_speex_quality(cubeb_resampler_quality q)
...
   }
 
-  return data_callback(stream, user_ptr,
-                       input_buffer, output_buffer, output_frames);
+  return rv;
```
padenot: This new code always consumes all the frames. Eventually, `input_frames_count` will be able to be passed by value.

src/cubeb_resampler.cpp
```diff
@@ -35,29 +35,54 @@ to_speex_quality(cubeb_resampler_quality q)
...
+         *input_frames_count + static_cast<int>(samples_to_frames(internal_input_buffer.length())) >= output_frames) ||
          (!input_buffer && (!input_frames_count || *input_frames_count == 0)) ||
          (!output_buffer && output_frames == 0));
 
+  long rv;
+
   if (output_buffer == nullptr) {
     assert(input_buffer);
     output_frames = *input_frames_count;
   }
```
padenot: Nope because you need to update `output_frames` iff output_buffer` is nullptr. But I agree, I'll do something along those lines.

Issue #149, opened 2016-09-04
Overly-aggressive filtering-out of PulseAudio sources and sinks


patrakov:
Cubeb has some logic in src/cubeb_pulse.c, functions pulse_get_state_from_sink_port() and pulse_get_state_from_source_port() that tries to filter out unplugged devices from enumeration. Unfortunately, it is too aggressive. Namely, it returns CUBEB_DEVICE_STATE_DISABLED for sinks and sources with no ports at all.

By accident, both ALSA and Bluetooth sinks created by module-alsa-card and module-bluez5-device contain ports, so common setups work. However, less common setups involve sources and sinks without ports:
- Manually-loaded module-alsa-source and module-alsa-sink (without module-alsa-card - this is useful e.g. to get SPDIF output in addition to the regular analog output)
- Module-tunnel-source, module-tunnel-sink
- Module-ladspa-sink
- Module-null-sink's monitor is useful for capturing the output of some other application without letting it play to real speakers

One of the important setups broken by upgrading Firefox to a version that contains cubeb with this filter is the headless xrdp setup where the only sinks and sources available are provided by xrdp for forwarding the client's speakers and microphone to the session.

kinetiknz:
Thanks for reporting this.  As a temporary workaround, you can revert to the older audio enumeration and capture code by setting media.navigator.audio.full_duplex to false in about:config.

It looks like the old code didn't do much filtering beyond excluding sink monitors.  @achronop do you mind investigating?

achronop:
Sure I can take it. Please assign it to me and let me know what is the priority.

patrakov:
@kinetiknz thanks for the workaround, it helps

kinetiknz:
> Sure I can take it. Please assign it to me and let me know what is the priority.

I can't assign it until you're a collaborator; I've sent you an invite.  Once you accept, you can assign yourself.

Priority will come from the WebRTC team, it doesn't affect media playback so I can't make that call.  Given it's a regression, probably reasonably high priority.  The fix is pretty simple (removing the filtering of devices with no ports), it's just that someone needs to verify the behaviour and ensure the UI looks okay with the unfiltered device list, etc.

kinetiknz:
This should be fixed by #224.

PR #224, opened 2017-02-03
pulse: return state enabled when port is NULL. (Bug 1332887)

achronop:
It solves #149. The device state is set as enabled when port is null. Tested by a contributor and works fine (Bug 1332887).

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -1270,7 +1270,8 @@ pulse_sink_info_cb(pa_context * context, const pa_sink_info * info,
 
   devinfo->type = CUBEB_DEVICE_TYPE_OUTPUT;
   devinfo->state = pulse_get_state_from_sink_port(info->active_port);
-  devinfo->preferred = strcmp(info->name, list_data->default_sink_name) == 0;
+  devinfo->preferred = (strcmp(info->name, list_data->default_sink_name) == 0) ?
+				 CUBEB_DEVICE_PREF_ALL : CUBEB_DEVICE_PREF_NONE;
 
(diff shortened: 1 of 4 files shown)


achronop:
Before that change the audio devices in Travis were reported as disabled. This does not bother the duplex and record tests which can skip testing and report success. After that change the devices are enabled so the test is executed. Since they are dummy devices the `stream_state.seen_noise` is never set and that's why it fails.

achronop:
I changed the way we check for input data. Instead of expecting for noise I expect buffers of float values between [-1.0, 1.0]. Moreover I changed the variable name from `seen_noise` to `seen_audio`. However the test_record still fails (in Linux). It seems that record callback is not executed at all. I raised issue #225 about it and disabled the check (in Linux). The new mechanism for checking audio works successfully in test_duplex.

achronop:
Since review hasn't started yet I pushed one more fix quite relevant to the previous one, it is reported as a side effect of the same bug [0]. The problem, in case of Linux (same in OSX) , is that we cannot differentiate the Preferred options described by the `enum cubeb_device_pref`. Till now we set preferred flag to true which corresponds to `CUBEB_DEVICE_PREF_MULTIMEDIA`. Inside gecko we are looking for `CUBEB_DEVICE_PREF_VOICE` [1] so we failed to report the device as preferred. I changed to set `CUBEB_DEVICE_PREF_ALL` instead of true.

[0] https://bugzilla.mozilla.org/show_bug.cgi?id=1332887#c25
[1] https://dxr.mozilla.org/mozilla-central/source/dom/media/webrtc/MediaEngineWebRTC.cpp#94

PR #227, opened 2017-02-06
Fix warnings in cubeb_alsa.c

padenot:
I'm a bit concerned about the substractions that are not checked for underflow, but this fixes the warnings.


kinetiknz:
Space between the cast and the target, e.g. `(snd_pcm_sframes_t) stm->bufframes`

Other than that, looks fine, go ahead and merge.

PR #229, opened 2017-02-08
Drop support for Windows versions lower than 7 (XP SP2 and Vista)

kinetiknz:
Now that https://bugzilla.mozilla.org/show_bug.cgi?id=1130266 is in effect, we no longer need to support XP SP2 and Vista.


padenot:
Cool!

PR #231, opened 2017-02-09
audiounit: Implement aggregate devices for duplex (Bug 1325577)

achronop:
This PR implements the aggregate device enhancement for full duplex operation. Aggregate device is a virtual device created from the given input and output devices and has the benefit of clock drift compensation etc. This new device is used to configure the existing AudioUnit(s).

One thing to note is that stream_init (duplex) will fail when the create of aggregate device fails. I could avoid that and fallback to the old code flow however I want to use it (probably setup telemetry in firefox) in order to figure out if (and how often) the create can fail. It never fails in my machine. I plan to remove it after a couple of weeks and allow the fallback mechanism for the Released versions.

The on going try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=623acf4fb244a2b51d2e9f1572b3423a657798b7

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1613,6 +1613,24 @@ audiounit_activate_clock_drift_compensation(cubeb_stream * stm)
 
 static int audiounit_destroy_aggregate_device(cubeb_stream * stm);
 
+/*
+ * Aggregate Device is a virtual audio interface which utilizes inputs and outputs
+ * of one or more physical audio interfaces. It is possible to use the clock of
+ * one of the devices as a master clock for all the combined devices and enable
+ * drift compensation for the devices that are not designated clock master.
+ *
+ * Creating a new aggregate device programmatically requires [0][1]:
+ * 1. Locate the base plug-in ("com.apple.audio.CoreAudio")
+ * 2. Create a dictionary that describes the aggregate device
+ *    (don't add sub-devices in that step, prone to fail [0])
+ * 3. Ask the base plug-in to create the aggregate device (blank)
+ * 4. Add the array of sub-devices.
+ * 5. Set the master device (1st output device in our case)
+ * 6. Enable drift compensation for the non-master devices
+ *
+ * [0] https://lists.apple.com/archives/coreaudio-api/2006/Apr/msg00092.html
+ * [1] https://lists.apple.com/archives/coreaudio-api/2005/Jul/msg00150.html
+ * */
 static int
 audiounit_create_aggregate_device(cubeb_stream * stm)
 {
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -64,7 +64,6 @@ void audiounit_stream_start_internal(cubeb_stream * stm);
 static void audiounit_close_stream(cubeb_stream *stm);
 static int audiounit_setup_stream(cubeb_stream *stm);
 
-
 extern cubeb_ops const audiounit_ops;
 
 struct cubeb {
@@ -256,7 +255,6 @@ struct cubeb_stream {
   /* This is true if a device change callback is currently running.  */
   std::atomic<bool> switching_device{ false };
   std::atomic<bool> buffer_size_change_state{ false };
-  // Aggregate devices data members
   AudioDeviceID aggregate_device_id = 0;    // the aggregate device id
   AudioObjectID plugin_id = 0;              // used to create aggregate device
 };
@@ -913,9 +911,7 @@ audiounit_uninstall_device_changed_callback(cubeb_stream * stm)
       return CUBEB_ERROR;
     }
 
-    /* Event to notify when the input is going away. */
-    AudioDeviceID dev = stm->input_device ? stm->input_device :
-                        audiounit_get_default_device_id(CUBEB_DEVICE_TYPE_INPUT);
+    AudioDeviceID dev = audiounit_get_input_device_id(stm);
     r = audiounit_remove_listener(stm, dev, kAudioDevicePropertyDeviceIsAlive,
                                   kAudioObjectPropertyScopeGlobal, &audiounit_property_listener_callback);
     if (r != noErr) {
@@ -1361,25 +1357,34 @@ audiounit_layout_init(cubeb_stream * stm, io_side side)
 static void
 audiounit_get_sub_devices(AudioDeviceID device_id, std::vector<AudioDeviceID> * sub_devices)
 {
+  assert(sub_devices->empty());
+  uint32_t count = 32;
+
   AudioObjectPropertyAddress property_address = { kAudioAggregateDevicePropertyActiveSubDeviceList,
                                                   kAudioObjectPropertyScopeGlobal,
                                                   kAudioObjectPropertyElementMaster };
-  AudioObjectID sub_device_array[32];
-  UInt32 size = sizeof(sub_device_array);
-  OSStatus r = AudioObjectGetPropertyData(device_id,
-                                          &property_address,
-                                          0,
-                                          nullptr,
-                                          &size,
-                                          sub_device_array);
-  if (r != noErr) {
+  UInt32 size = 0;
+  OSStatus rv = AudioObjectGetPropertyDataSize(device_id,
+                                               &property_address,
+                                               0,
+                                               nullptr,
+                                               &size);
+  if (rv == noErr) {
+    count = static_cast<uint32_t>(size / sizeof(AudioObjectID));
+  }
+  sub_devices->resize(count);
+
+  rv = AudioObjectGetPropertyData(device_id,
+                                  &property_address,
+                                  0,
+                                  nullptr,
+                                  &size,
+                                  sub_devices->data());
+  if (rv != noErr) {
+    sub_devices->clear();
     sub_devices->push_back(device_id);
   } else {
-    int num_devices = size / sizeof(AudioObjectID);
-    LOG("Found %d subdevices", num_devices);
-    for (int i = 0; i < num_devices; i++) {
-      sub_devices->push_back(sub_device_array[i]);
-    }
+    LOG("Found %u sub-devices", count);
   }
 }
 
@@ -1558,7 +1563,7 @@ audiounit_activate_clock_drift_compensation(cubeb_stream * stm)
 
   UInt32 qualifier_data_size = sizeof(AudioObjectID);
   AudioClassID class_id = kAudioSubDeviceClassID;
-  void* qualifier_data = &class_id;
+  void * qualifier_data = &class_id;
   UInt32 size = 0;
   OSStatus rv = AudioObjectGetPropertyDataSize(stm->aggregate_device_id,
                                                &address_owned,
@@ -1606,10 +1611,10 @@ audiounit_activate_clock_drift_compensation(cubeb_stream * stm)
   return CUBEB_OK;
 }
 
-static int audiounit_destroy_aggregate_device(cubeb_stream*);
+static int audiounit_destroy_aggregate_device(cubeb_stream * stm);
 
 static int
-audiounit_create_aggregate_device(cubeb_stream *stm)
+audiounit_create_aggregate_device(cubeb_stream * stm)
 {
   int r = audiounit_create_blank_aggregate_device(stm);
   if (r != CUBEB_OK) {
@@ -1762,7 +1767,6 @@ audiounit_create_unit(AudioUnit * unit, io_side side, AudioDeviceID device)
 
   if (!use_default_output) {
     switch (side) {
-
       case INPUT:
       {
         r = audiounit_enable_unit_scope(unit, INPUT, ENABLE);
@@ -1777,7 +1781,6 @@ audiounit_create_unit(AudioUnit * unit, io_side side, AudioDeviceID device)
         }
         break;
       }
-
       case OUTPUT:
       {
         r = audiounit_enable_unit_scope(unit, OUTPUT, ENABLE);
@@ -1792,7 +1795,6 @@ audiounit_create_unit(AudioUnit * unit, io_side side, AudioDeviceID device)
         }
         break;
       }
-
       default:
         assert(false);
     }
@@ -1804,8 +1806,10 @@ audiounit_create_unit(AudioUnit * unit, io_side side, AudioDeviceID device)
       devid = device;
     }
 
-    rv = AudioUnitSetProperty(*unit, kAudioOutputUnitProperty_CurrentDevice,
-                              kAudioUnitScope_Global, 0,
+    rv = AudioUnitSetProperty(*unit,
+                              kAudioOutputUnitProperty_CurrentDevice,
+                              kAudioUnitScope_Global,
+                              0,
                               &devid, sizeof(AudioDeviceID));
     if (rv != noErr) {
       LOG("AudioUnitSetProperty/kAudioOutputUnitProperty_CurrentDevice rv=%d", rv);
@@ -2254,7 +2258,7 @@ audiounit_setup_stream(cubeb_stream * stm)
                               INPUT,
                               in_dev);
     if (r != CUBEB_OK) {
-      LOG("(%p) Create AudioUnit for input failed.", stm);
+      LOG("(%p) AudioUnit creation for input failed.", stm);
       return r;
     }
   }
@@ -2264,11 +2268,10 @@ audiounit_setup_stream(cubeb_stream * stm)
                               OUTPUT,
                               out_dev);
     if (r != CUBEB_OK) {
-      LOG("(%p) Create AudioUnit for output failed.", stm);
+      LOG("(%p) AudioUnit creation for output failed.", stm);
       return r;
     }
   }
-
 
   /* Latency cannot change if another stream is operating in parallel. In this case
   * latecy is set to the other stream value. */
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1164,8 +1164,7 @@ audiounit_get_preferred_channel_layout()
   return audiounit_convert_channel_layout(layout.get());
 }
 
-static int
-audiounit_create_unit(AudioUnit * unit, io_side side, AudioDeviceID device);
+static int audiounit_create_unit(AudioUnit * unit, io_side side, AudioDeviceID device);
 
 static int
 audiounit_get_preferred_channel_layout(cubeb * ctx, cubeb_channel_layout * layout)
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -1211,16 +1229,427 @@ audiounit_layout_init(AudioUnit * unit,
   return CUBEB_OK;
 }
 
+static int audiounit_create_ggregate_device_aux(cubeb_stream*, std::vector<AudioDeviceID>,
```
kinetiknz: Typo in the function name and everywhere else it occurs (callers and log messages)

src/cubeb_audiounit.cpp
```diff
@@ -1211,16 +1229,427 @@ audiounit_layout_init(AudioUnit * unit,
...
+audiounit_create_ggregate_device_aux(cubeb_stream * stm,
+                                     std::vector<AudioDeviceID> input_subdevice_id_array,
+                                     std::vector<AudioDeviceID> output_subdevice_id_array,
+                                     AudioDeviceID* out_aggregate_device)
+{
+  OSStatus osErr = noErr;
+  UInt32 out_size;
+
+  // Prepare sub-devices for clock drift compensation
+  // Workaround for bug in the HAL : until 10.6.2
```
kinetiknz: What bug? Is there a reference/description?

But there's a larger issue here.  From a brief look it appears chunks of code have been copied from somewhere else (JackCoreAudioAdapter.cpp?) without credit to the original author and (assuming that's the original source) without relicensing from GPL v2+.  The formatting of the comments makes it quite obvious since the new additions are in a completely different style to the existing libcubeb code.

achronop: Indeed the import is from JACK2 code. I made changes in order to adapt it to our code style, make use of the latest CoreAudio API and remove stuff we do not need. 

Sorry, I am not familiar with the process, can you please elaborate a little more. Is there compatibility with GPL V2+? By attribution you mean to give credits, provide license link and indicate changes. Where should I do that?

kinetiknz: > Is there compatibility with GPL V2+?

No, not for libcubeb's ISC license or Gecko's MPL license.

> By attribution you mean to give credits, provide license link and indicate changes. Where should I do that?

Aside from the license issue (which can only be dealt with by rewriting the code or asking the original author if it's possible to relicense the parts you need to ISC), you should reference the original source and author when importing (license compatible) code.  If it's a chunk imported to an existing file, then a comment near that code is fine.  If it's an entire file, it's better to add a note to the README to avoid modifying the original file if possible.

achronop:
I removed the imported part and wrote it from the scratch. Can you please have a look?

kinetiknz:
> I removed the imported part and wrote it from the scratch. Can you please have a look?

Thanks, will do.

src/cubeb_audiounit.cpp
```diff
@@ -1211,16 +1229,328 @@ audiounit_layout_init(AudioUnit * unit,
...
+                                          nullptr,
+                                          &size,
+                                          sub_device_array);
+  if (r != noErr) {
+    sub_devices->push_back(device_id);
+  } else {
+    int num_devices = size / sizeof(AudioObjectID);
+    LOG("Found %d subdevices", num_devices);
+    for (int i = 0; i < num_devices; i++) {
+      sub_devices->push_back(sub_device_array[i]);
```
kinetiknz: And create the vector in audiounit_get_sub_devices and return it rather than requiring the caller to pass one in, like `audiounit_get_devices_of_type`.

src/cubeb_audiounit.cpp
```diff
@@ -1211,16 +1229,328 @@ audiounit_layout_init(AudioUnit * unit,
...
+  }
+  return CUBEB_OK;
+}
+
+static int audiounit_destroy_aggregate_device(cubeb_stream*);
+
+static int
+audiounit_create_aggregate_device(cubeb_stream *stm)
+{
+  int r = audiounit_create_blank_aggregate_device(stm);
```
kinetiknz: The only state from `stm` this needs is `plugin_id` and `aggregate_device_id`, so just pass those.

src/cubeb_audiounit.cpp
```diff
@@ -1211,16 +1229,328 @@ audiounit_layout_init(AudioUnit * unit,
...
+  if (r != CUBEB_OK) {
+    LOG("Failed to create blank aggregate device");
+    audiounit_destroy_aggregate_device(stm);
+    return CUBEB_ERROR;
+  }
+
+  r = audiounit_set_aggregate_sub_device_list(stm);
```
kinetiknz: Same: this only needs `aggregate_device_id` and `input_device` (you'll need to refactor `audiounit_get_input_device_id` to not require `stm` also)

src/cubeb_audiounit.cpp
```diff
@@ -1211,16 +1229,328 @@ audiounit_layout_init(AudioUnit * unit,
...
+  if (r != CUBEB_OK) {
+    LOG("Failed to set aggregate sub-device list");
+    audiounit_destroy_aggregate_device(stm);
+    return CUBEB_ERROR;
+  }
+
+  r = audiounit_set_master_aggregate_device(stm);
```
kinetiknz: Only needs `aggregate_device_id`

src/cubeb_audiounit.cpp
```diff
@@ -1211,16 +1229,328 @@ audiounit_layout_init(AudioUnit * unit,
...
+  if (r != CUBEB_OK) {
+    LOG("Failed to set master sub-device for aggregate device");
+    audiounit_destroy_aggregate_device(stm);
+    return  CUBEB_ERROR;
+  }
+
+  r = audiounit_activate_clock_drift_compensation(stm);
```
kinetiknz: Only needs `aggregate_device_id`

src/cubeb_audiounit.cpp
```diff
@@ -1211,16 +1229,328 @@ audiounit_layout_init(AudioUnit * unit,
   return CUBEB_OK;
 }
 
+static void
+audiounit_get_sub_devices(AudioDeviceID device_id, std::vector<AudioDeviceID> * sub_devices)
```
achronop: I intended to do it by reference in the initial version of the code, however debugging an error I realized that the problem was I had forgotten the reference (`&`). Passing by pointer prevents from problems like this. In addition is more obvious inside the body of the method that `sub_devices` is being updated. If the above does not persuade you to keep the pointer, let me know and I will update it.

src/cubeb_audiounit.cpp
```diff
@@ -1211,16 +1229,328 @@ audiounit_layout_init(AudioUnit * unit,
...
+                                           nullptr,
+                                           size,
+                                           &aggregate_sub_devices_array);
+  CFRelease(aggregate_sub_devices_array);
+  if (rv != noErr) {
+    PRINT_ERROR_CODE("AudioObjectSetPropertyData/kAudioAggregateDevicePropertyFullSubDeviceList", rv);
+    return CUBEB_ERROR;
+  }
+
+  CFRunLoopRunInMode(kCFRunLoopDefaultMode, 0.1, false);
```
achronop: Provide to framework sometime to perform the current updates. It's a measurement I take to avoid deadlock etc. It is not documented anywhere and I cannot prove it is necessary.

src/cubeb_audiounit.cpp
```diff
@@ -1211,16 +1229,328 @@ audiounit_layout_init(AudioUnit * unit,
...
+                                           0,
+                                           NULL,
+                                           size,
+                                           &master_sub_device);
+  if (rv != noErr) {
+    PRINT_ERROR_CODE("AudioObjectSetPropertyData/kAudioAggregateDevicePropertyMasterSubDevice", rv);
+    return CUBEB_ERROR;
+  }
+
+  CFRunLoopRunInMode(kCFRunLoopDefaultMode, 0.1, false);
```
achronop: Same as above.

src/cubeb_audiounit.cpp
achronop: I copied that pattern from wasapi backend. I found it easier to use a method that way. You do not have to worry about which AudioDevideID to use. Moreover there is no need to check if the arguments are the correct. Since all these methods `audiounit_create_blank_aggregate_device`, `audiounit_set_aggregate_sub_device_list` and `audiounit_get_input_device_id` cannot be used with different arguments I find it easier that way. I don't mean to reject the comment, I am discussing the reasons I like it. If you do not agree I can change it.

src/cubeb_audiounit.cpp
```diff
@@ -1253,38 +1583,110 @@ audiounit_create_unit(AudioUnit * unit,
...
+      assert(side == INPUT);
       devid = audiounit_get_default_device_id(CUBEB_DEVICE_TYPE_INPUT);
     } else {
       devid = device;
     }
+
     rv = AudioUnitSetProperty(*unit, kAudioOutputUnitProperty_CurrentDevice,
-                              kAudioUnitScope_Global,
-                              is_input ? AU_IN_BUS : AU_OUT_BUS,
+                              kAudioUnitScope_Global, 0,
```
achronop: Yes, it does not matter whether you set it IN or OUT. I confirmed with the following [sample code](https://developer.apple.com/library/content/technotes/tn2091/_index.html). I do not use the defined AU_{IN,OUT}_BUS to indicate that.

My guess is that there are ... I will take a look.

src/cubeb_audiounit.cpp
```diff
@@ -1253,38 +1583,110 @@ audiounit_create_unit(AudioUnit * unit,
...
                               &devid, sizeof(AudioDeviceID));
     if (rv != noErr) {
-      PRINT_ERROR_CODE("AudioUnitSetProperty/kAudioOutputUnitProperty_CurrentDevice", rv);
+      if (side == INPUT)
+        PRINT_ERROR_CODE("AudioUnitSetProperty/input/kAudioOutputUnitProperty_CurrentDevice", rv);
+      else
+        PRINT_ERROR_CODE("AudioUnitSetProperty/output/kAudioOutputUnitProperty_CurrentDevice", rv);
```
achronop: I raised  #240 to fix that in a separate PR.

achronop:
I pushed the update according to the first set of comments. I posted some answers and I am waiting your feedback. I will start writing the documentation for how aggregate devices works.

achronop:
I applied (or responded) all comments, looking forward for more. Moreover branch rebased to latest master.

src/cubeb_audiounit.cpp
```diff
@@ -1346,16 +1354,355 @@ audiounit_layout_init(cubeb_stream * stm, io_side side)
   stm->context->layout = audiounit_get_current_channel_layout(stm->output_unit);
 }
 
+static void
+audiounit_get_sub_devices(AudioDeviceID device_id, std::vector<AudioDeviceID> * sub_devices)
+{
+  assert(sub_devices->empty());
```
kinetiknz: As I suggested in my previous set of comments, rather than taking a vector argument create it within the function and return it was a result.  See audiounit_get_devices_of_type for an example.

src/cubeb_audiounit.cpp
```diff
@@ -1346,16 +1354,355 @@ audiounit_layout_init(cubeb_stream * stm, io_side side)
...
+  UInt32 size = 0;
+  OSStatus rv = AudioObjectGetPropertyDataSize(device_id,
+                                               &property_address,
+                                               0,
+                                               nullptr,
+                                               &size);
+  if (rv == noErr) {
+    count = static_cast<uint32_t>(size / sizeof(AudioObjectID));
+  }
+  sub_devices->resize(count);
```
kinetiknz: In what situation would AudioObjectGetPropertyDataSize fail but AudioObjectGetPropertyData succeed in this case?

It seems like it'd be clearer to push_back(device_id) and return if AudioObjectGetPropertyDataSize failed, otherwise go ahead with the resize and query.  That also removes the weird count = 32 default and questions about what happens if AudioObjectGetPropertyDataSize but AudioObjectGetPropertyData succeeds and the result needs more space than we allocated with the 32 element default.

src/cubeb_audiounit.cpp
kinetiknz: Yes, please change it to what I suggested.  I don't think the current approach is well factored and the fact the pattern exists in the WASAPI backends suggests we have more places to clean up.

src/cubeb_audiounit.cpp
```diff
@@ -1388,35 +1735,99 @@ audiounit_create_unit(AudioUnit * unit,
...
-      return CUBEB_ERROR;
+enum enable_state {
+  DISABLE,
+  ENABLE,
+};
+
+static int
+audiounit_enable_unit_scope(AudioUnit * unit,
+                            io_side side,
+                            enable_state state)
```
kinetiknz: Params all on one line.

src/cubeb_audiounit.cpp
```diff
@@ -1346,16 +1354,355 @@ audiounit_layout_init(cubeb_stream * stm, io_side side)
...
+    return CUBEB_ERROR;
+  }
+
+  return CUBEB_OK;
+}
+
+static int
+audiounit_new_unit_instance(AudioUnit * unit,
+                            io_side side,
+                            AudioDeviceID device)
```
kinetiknz: Params all on one line.

src/cubeb_audiounit.cpp
```diff
@@ -1346,16 +1354,355 @@ audiounit_layout_init(cubeb_stream * stm, io_side side)
...
+      return CUBEB_OK;
+    }
+  }
+  return CUBEB_OK;
+}
+
+static int audiounit_destroy_aggregate_device(cubeb_stream * stm);
+
+/*
+ * Aggregate Device is a virtual audio interface which utilizes inputs and outputs
```
kinetiknz: Maybe add a reference to AudioHardware.h too, since that's where the (little) info available seems to be

src/cubeb_audiounit.cpp
```diff
@@ -1388,35 +1735,99 @@ audiounit_create_unit(AudioUnit * unit,
...
+#if TARGET_OS_IPHONE
+  bool use_default_output = false;
+#else
+  bool use_default_output = device == 0 && (side == OUTPUT);
+#endif
+
+  if (!use_default_output) {
+    switch (side) {
+      case INPUT:
+      {
```
kinetiknz: These braces aren't needed since you're not declaring new vars.

src/cubeb_audiounit.cpp
```diff
@@ -1474,7 +1886,7 @@ audiounit_clamp_latency(cubeb_stream * stm, uint32_t latency_frames)
 
   UInt32 input_buffer_size = 0;
   if (stm->input_unit) {
-    r = AudioUnitGetProperty(stm->input_unit,
+    r = AudioUnitGetProperty(stm->output_unit,
```
kinetiknz: The test above checks input_unit and you've changed the code to use output_unit, is that intentional?

kinetiknz:
> Ours are configured as private, so they don't show up in the AMS list, but it seems like they're still persistent. Have you tested the situation where libcubeb/Gecko is shut down in an unclean way (e.g. by crashing) so doesn't clean up the aggregate and then a later run tries to (re)use it?

According to AudioHardware.h private aggregate devices aren't persistent, so we don't need to worry about this, which is a huge relief.

src/cubeb_audiounit.cpp
achronop: Done, looks much better that way. In your previous set of comments you had mention to take sub devices by reference.

src/cubeb_audiounit.cpp
```diff
@@ -1346,16 +1354,355 @@ audiounit_layout_init(cubeb_stream * stm, io_side side)
...
+                                           nullptr,
+                                           size,
+                                           &aggregate_sub_devices_array);
+  CFRelease(aggregate_sub_devices_array);
+  if (rv != noErr) {
+    LOG("AudioObjectSetPropertyData/kAudioAggregateDevicePropertyFullSubDeviceList, rv=%d", rv);
+    return CUBEB_ERROR;
+  }
+
+  CFRunLoopRunInMode(kCFRunLoopDefaultMode, 0.1, false);
```
achronop: This kind of errors do not repro locally. It does work without. I removed it for now, Let's reconsider if we see strange deadlocks.

achronop:
> I've had a bit more of a look at the aggregate devices stuff in general to get my head around it. Why don't we use the AudioHardwareCreateAggregateDevice API added in 10.9? It looks much simpler and we don't need to support earlier versions of macOS in libcubeb or Gecko. If we can use AudioHardwareCreateAggregateDevice, let's file an issue and plan to do that in a follow-up PR.

Issue #244 raised. I will follow up when this one is landed.

achronop:
> According to AudioHardware.h private aggregate devices aren't persistent, so we don't need to worry about this, which is a huge relief.

Testing also confirms that they are not persistent.

achronop:
> Are you able to explain why Chromium doesn't seem to need aggregate devices for their full duplex implementation? I see they used to have some aggregate device code, but it was removed a few years ago. How are they dealing with clock drift?

The aggregate device code added with [Issue 13403002](https://chromiumcodereview.appspot.com/13403002) and removed with [Issue 163343002](https://codereview.chromium.org/460373002) as unused code. From a short research in their code I see that that they move IO data to [FIFO buffers](https://cs.chromium.org/chromium/src/media/base/audio_pull_fifo.h). They render output [here](https://cs.chromium.org/chromium/src/media/renderers/audio_renderer_impl.cc?type=cs&l=759) and they [push zero](https://cs.chromium.org/chromium/src/media/filters/audio_clock.cc?type=cs&l=52) to fill the buffers in case they do not have enough data. I might miss something it was a quick look ...

achronop:
I finished the update according to your comments. Just a short note, Firefox's merge day is on Monday. In case (without any push) you feel we can make it for 54 I can apply any new comment and do the import with priority. I am sure Maire wants it in 54. If not it's totally fine.

src/cubeb_audiounit.cpp
kinetiknz: You didn't update this one, I think we should refactor it the same way... but it can be done in a follow-up bug.

src/cubeb_audiounit.cpp
```diff
@@ -1346,16 +1353,356 @@ audiounit_layout_init(cubeb_stream * stm, io_side side)
...
+    LOG("(%p) Failed to activate clock drift compensation for aggregate device", stm);
+    audiounit_destroy_aggregate_device(stm);
+    return  CUBEB_ERROR;
+  }
+
+  return CUBEB_OK;
+}
+
+static int
+audiounit_destroy_aggregate_device(cubeb_stream * stm)
```
kinetiknz: In a follow-up: this one can be refactored to take plugin_id and aggregate_device_id too

src/cubeb_audiounit.cpp
```diff
@@ -1838,20 +2247,38 @@ audiounit_setup_stream(cubeb_stream * stm)
...
+  AudioDeviceID out_dev = stm->output_device;
+  if (has_input(stm) && has_output(stm)) {
+    r = audiounit_create_aggregate_device(stm);
+    if (r != CUBEB_OK) {
+      stm->aggregate_device_id = 0;
+      LOG("(%p) Create aggregate devices failed.", stm);
+      // !!!NOTE: It is not necessary to return here. If it does not
+      // return it will fallback to the old implementation. The intention
+      // is to investigate how often it fails. I plan to remove
+      // it after a couple of weeks.
```
kinetiknz: Please file a bug/issue to track the result of this experiment and re-enabling the fallback behaviour.

PR #234, opened 2017-02-16
Downmix audio on OSX

ChunMinChang:
The audio rendering mechanism on OS X will drop the extra channels beyond the channels that audio device can provide. To keep all information, we need to downmix the audio data by ourselves.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -195,6 +195,45 @@ private:
   auto_array<T> ar;
 };
 
+struct mixing_wrapper {
+  virtual void downmix(void * output_buffer, long output_frames, cubeb_channel_layout output_layout, cubeb_channel_layout mixing_layout) = 0;
+  virtual ~mixing_wrapper() {};
+};
+template <typename T>
+struct mixing_impl : public mixing_wrapper {
+
+  using downmix_func_ptr = void (*)(T * const, long, T *, unsigned int, unsigned int, cubeb_channel_layout, cubeb_channel_layout);
+  using remap_func_ptr = bool (*)(T const * const, unsigned long, T *, cubeb_channel_layout, cubeb_channel_layout);
+  mixing_impl(downmix_func_ptr dfptr, remap_func_ptr rfptr) {
+    downmix_wrapper = dfptr;
+    remap_wrapper = rfptr;
+    // If we allocate the mixed buffer here, what is its size ?
+  }
+
+  ~mixing_impl() {}
+
+  void downmix(void * output_buffer, long output_frames, cubeb_channel_layout output_layout, cubeb_channel_layout mixing_layout) override {
+    uint32_t output_channels = CUBEB_CHANNEL_LAYOUT_MAPS[output_layout].channels;
+    uint32_t using_channels = CUBEB_CHANNEL_LAYOUT_MAPS[mixing_layout].channels;
+    buffer = std::vector<T>(using_channels * output_frames);
+    T* out = static_cast<T *>(output_buffer);
+    (*downmix_wrapper)(out, output_frames, buffer.data(), output_channels, using_channels, output_layout, mixing_layout);
+
+    // The mixed buffer returned from downmixing is shrunk by its channel counts.
+    // Although the extra channels beyonds the channels that the audio device
+    // can provide will be dropped, we still need to feed the unused channel
+    // buffer to the backend for processing. Thus, we need to assign the
+    // mixed buffer's data to the original output buffer.
+    (*remap_wrapper)(buffer.data(), output_frames, out, mixing_layout, output_layout);
+  }
+
+  /* Mixed buffer */
+  std::vector<T> buffer;
+  /* Function pointer for downmixing and remapping */
+  void (*downmix_wrapper)(T * const, long, T *, unsigned int, unsigned int, cubeb_channel_layout, cubeb_channel_layout);
+  bool (*remap_wrapper)(T const * const, unsigned long, T *, cubeb_channel_layout, cubeb_channel_layout);
+};
+
 enum io_side {
   INPUT,
   OUTPUT,
@@ -257,6 +296,8 @@ struct cubeb_stream {
   /* This is true if a device change callback is currently running.  */
   std::atomic<bool> switching_device{ false };
   std::atomic<bool> buffer_size_change_state{ false };
+  /* Mixing interface */
+  std::unique_ptr<mixing_wrapper> mixing;
 };
 
 bool has_input(cubeb_stream * stm)
@@ -483,21 +524,6 @@ is_extra_input_needed(cubeb_stream * stm)
          stm->available_input_frames < minimum_resampling_input_frames(stm);
 }
 
-template <typename T>
-void
-audiounit_copy_through_mixed_buffer_to_output(T* output_buffer,
-                                              long output_frames,
-                                              T* mixed_buffer,
-                                              UInt32 output_channels,
-                                              UInt32 mixed_channels)
-{
-  for (long i = 0, o = 0 ; i < output_frames ; ++i, o += output_channels) {
-    for (UInt32 c = 0 ; c < mixed_channels ; ++c) {
-      output_buffer[o + c] = mixed_buffer[i * mixed_channels + c];
-    }
-  }
-}
-
 void
 audiounit_mix_output_buffer(cubeb_stream * stm,
                             void * output_buffer,
@@ -517,27 +543,7 @@ audiounit_mix_output_buffer(cubeb_stream * stm,
 
   // We only handle downmixing for now.
   if (cubeb_should_downmix(&stm->output_stream_params, &mixed_params)) {
-    if (stm->output_desc.mFormatFlags & kAudioFormatFlagIsFloat) {
-      std::vector<float> mixed_buffer(mixed_params.channels * output_frames);
-      float* out = (float *)output_buffer;
-      cubeb_downmix_float(out, output_frames, mixed_buffer.data(),
-                          stm->output_stream_params.channels, mixed_params.channels,
-                          stm->output_stream_params.layout, mixed_params.layout);
-      audiounit_copy_through_mixed_buffer_to_output(out, output_frames,
-                                                    mixed_buffer.data(),
-                                                    stm->output_stream_params.channels,
-                                                    mixed_params.channels);
-    } else if (stm->output_desc.mFormatFlags & kAudioFormatFlagIsSignedInteger) {
-      std::vector<short> mixed_buffer(mixed_params.channels * output_frames);
-      short* out = (short *)output_buffer;
-      cubeb_downmix_short(out, output_frames, mixed_buffer.data(),
-                          stm->output_stream_params.channels, mixed_params.channels,
-                          stm->output_stream_params.layout, mixed_params.layout);
-      audiounit_copy_through_mixed_buffer_to_output(out, output_frames,
-                                                    mixed_buffer.data(),
-                                                    stm->output_stream_params.channels,
-                                                    mixed_params.channels);
-    }
+    stm->mixing->downmix(output_buffer, output_frames, stm->output_stream_params.layout, mixed_params.layout);
   }
 }
 
@@ -1225,6 +1231,16 @@ audio_stream_desc_init(AudioStreamBasicDescription * ss,
   return CUBEB_OK;
 }
 
+void
+audiounit_init_mixed_buffer(cubeb_stream * stm)
+{
+  if (stm->output_desc.mFormatFlags & kAudioFormatFlagIsFloat) {
+    stm->mixing.reset(new mixing_impl<float>(&cubeb_downmix_float, &mix_remap_float));
+  } else if (stm->output_desc.mFormatFlags & kAudioFormatFlagIsSignedInteger) {
+    stm->mixing.reset(new mixing_impl<short>(&cubeb_downmix_short, &mix_remap_short));
+  }
+}
+
 static int
 audiounit_layout_init(AudioUnit * unit,
                       io_side side,
@@ -1802,6 +1818,7 @@ audiounit_configure_output(cubeb_stream * stm)
   }
 
   audiounit_layout_init(&stm->output_unit, OUTPUT, &stm->output_stream_params);
+  audiounit_init_mixed_buffer(stm);
 
   LOG("(%p) Output audiounit init successfully.", stm);
   return CUBEB_OK;
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -196,14 +196,21 @@ private:
 };
 
 struct mixing_wrapper {
-  virtual void downmix(void * output_buffer, long output_frames, cubeb_channel_layout output_layout, cubeb_channel_layout mixing_layout) = 0;
+  virtual void downmix(void * output_buffer, long output_frames,
+                       cubeb_channel_layout output_layout,
+                       cubeb_channel_layout mixing_layout) = 0;
   virtual ~mixing_wrapper() {};
 };
+
 template <typename T>
 struct mixing_impl : public mixing_wrapper {
 
-  using downmix_func_ptr = void (*)(T * const, long, T *, unsigned int, unsigned int, cubeb_channel_layout, cubeb_channel_layout);
-  using remap_func_ptr = bool (*)(T const * const, unsigned long, T *, cubeb_channel_layout, cubeb_channel_layout);
+  using downmix_func_ptr = void (*)(T * const, long, T *,
+                                    unsigned int, unsigned int,
+                                    cubeb_channel_layout, cubeb_channel_layout);
+  using remap_func_ptr = bool (*)(T const * const, unsigned long, T *,
+                                  cubeb_channel_layout, cubeb_channel_layout);
+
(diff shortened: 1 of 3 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -483,6 +483,64 @@ is_extra_input_needed(cubeb_stream * stm)
...
+                                              T* mixed_buffer,
+                                              UInt32 output_channels,
+                                              UInt32 mixed_channels)
+{
+  for (long i = 0, o = 0 ; i < output_frames ; ++i, o += output_channels) {
+    for (UInt32 c = 0 ; c < mixed_channels ; ++c) {
+      output_buffer[o + c] = mixed_buffer[i * mixed_channels + c];
+    }
+  }
+}
```
padenot: I think we have something similar in `cubeb_wasapi.cpp`, can you consolidate the two function in `cubeb_mixer.cpp`

ChunMinChang: I found the above code is wrong. I think we should reuse ```mix_remap``` here because if the ```output channel```(the audio data's layout) is ```L, R, C, LFE, LS, RS``` and ```mixed_layout```(the used speaker's layout) is ```L, R, LS, RS``` then the above code will write the ```LS, RS``` to ```C, LFE```. That's not the expected result.

src/cubeb_audiounit.cpp
```diff
@@ -483,6 +483,64 @@ is_extra_input_needed(cubeb_stream * stm)
...
+      float* out = (float *)output_buffer;
+      cubeb_downmix_float(out, output_frames, mixed_buffer.data(),
+                          stm->output_stream_params.channels, mixed_params.channels,
+                          stm->output_stream_params.layout, mixed_params.layout);
+      audiounit_copy_through_mixed_buffer_to_output(out, output_frames,
+                                                    mixed_buffer.data(),
+                                                    stm->output_stream_params.channels,
+                                                    mixed_params.channels);
+    } else if (stm->output_desc.mFormatFlags & kAudioFormatFlagIsSignedInteger) {
+      std::vector<short> mixed_buffer(mixed_params.channels * output_frames);
```
ChunMinChang: OK, thanks for noticing the inappropriate design. I would like to modify it but I don't know what is the buffer size we should allocate. Could we calculate the buffer size in advance before getting the ```output_frames``` from the callback?

ChunMinChang:
@padenot 
Thanks for reviewing the patch! 
I'd like to remove the allocated buffer in callback, but I don't know how we could calculate the buffer size for mixing the audio data before getting ```output_frames``` from the callback. So, I add a wrapper using template to process downmixing beyond data types first.

src/cubeb_audiounit.cpp
```diff
@@ -483,6 +524,29 @@ is_extra_input_needed(cubeb_stream * stm)
...
+void
+audiounit_mix_output_buffer(cubeb_stream * stm,
+                            void * output_buffer,
+                            long output_frames)
+{
+  // The audio rendering mechanism on OS X will drop the extra channels beyond
+  // the channels that audio device can provide, so we need to downmix the
+  // audio data by ourselves to keep all the information.
+
+  cubeb_channel_layout usedLayout = audiounit_get_channel_layout(false);
```
ChunMinChang: Thanks for giving inspection!

I did try to use ```AudioObjectGetPropertyDataSize``` to get preferred channel layout on my [draft](https://gist.github.com/ChunMinChang/ea74c8228745449873716e1d98ba956e#file-channellayout-cpp-L325), but it doesn't work yet with ```kAudioDevicePropertyPreferredChannelLayout```. I think we should follow your advice to use ```cubeb_stream*``` first. I'll open another PR to do that, thanks!

achronop: If this is the call that did not work:
https://gist.github.com/ChunMinChang/ea74c8228745449873716e1d98ba956e#file-channellayout-cpp-L332
I wonder if you tested for scope `kAudioObjectPropertyScopeOutput`? Global scope does not seem appropriate when you set a specific `AudioObjectId` like the `id` arg here:
https://gist.github.com/ChunMinChang/ea74c8228745449873716e1d98ba956e#file-channellayout-cpp-L336

ChunMinChang: Yes, we could use ```kAudioObjectPropertyScopeOutput``` to get the ```kAudioDevicePropertyPreferredChannelLayout``` now. However, I forget to mention that using ```kAudioDevicePropertyPreferredChannelLayout``` to get channel layout is unpredictable. The ```mChannelDescriptions[<index of channel>].mChannelLabel``` of channel layout will be ```kAudioChannelLabel_Unknown``` if I apply ```kAudioDevicePropertyPreferredChannelLayout```, so it will be treated as ```UNDEFINED``` layout. Instead, if I use ```kAudioUnitProperty_AudioChannelLayout```, then the ```mChannelLabel``` for default output will be ```kAudioChannelLabel_Left``` and ```kAudioChannelLabel_Right```.

Maybe we need to add a function to guess the layout from channel counts or set a policy like
1) Use ```AudioObjectGetProperty``` to get ```kAudioDevicePropertyPreferredChannelLayout```
2) Use ```AudioUnitGetProperty``` to get ```kAudioUnitProperty_AudioChannelLayout``` with live ```cubeb_stream*```
3) Use ```AudioUnitGetProperty``` to get ```kAudioUnitProperty_AudioChannelLayout``` with new ```cubeb_stream*```

src/cubeb_audiounit.cpp
```diff
@@ -195,6 +195,45 @@ struct auto_array_wrapper_impl : public auto_array_wrapper {
...
+};
+template <typename T>
+struct mixing_impl : public mixing_wrapper {
+
+  using downmix_func_ptr = void (*)(T * const, long, T *, unsigned int, unsigned int, cubeb_channel_layout, cubeb_channel_layout);
+  using remap_func_ptr = bool (*)(T const * const, unsigned long, T *, cubeb_channel_layout, cubeb_channel_layout);
+  mixing_impl(downmix_func_ptr dfptr, remap_func_ptr rfptr) {
+    downmix_wrapper = dfptr;
+    remap_wrapper = rfptr;
+    // If we allocate the mixed buffer here, what is its size ?
```
ChunMinChang: I tried to allocate the buffer with size ```stm->latency_frames * channels``` but it's not always larger than ```channels * output_frames(from audiounit_output_callback)```.

ChunMinChang: I think we could try reuse the output_buffer instead of creating a new one.

ChunMinChang:
```downmix_3f2``` now is modified to allow using same buffer for input and output for downmixing and it won't change to unrelated channel buffers. Thus, we don't need to allocate any buffer for downmixing audio. For example, if we have a buffer is ```[L_0, R_0, C_0, LFE_0, LS_0, RS_0, L_1, R_1, C_1, LFE_1, LS_1, RS_1, ...]```, then the downmixed buffer will be ```[L_0+ 0.707*(C_0 + LS_0), R_0+ 0.707*(C_0 + RS_0), C_0, LFE_0, LS_0, RS_0, L_1+ 0.707*(C_1 + LS_1), R_1+ 0.707*(C_1 + RS_1), C_1, LFE_1, LS_1, RS_1, ...]```. If the buffer for input and output is different, then the behavior is same as usual. That is, the downmixed buffer will be ```[L_0+ 0.707*(C_0 + LS_0), R_0+ 0.707*(C_0 + RS_0), L_1+ 0.707*(C_1 + LS_1), R_1+ 0.707*(C_1 + RS_1), ...]```. 

However, this is a little bit tricky, so it might confuse users when they call the function. If there is any way to make it better, please let me know. Thanks!

**Comment by ChunMinChang** [comment]
@achronop : I was wondering if you could give me some suggestions about this issue.

To avoid allocating buffers when downmixing audio, I reused the buffer returned from ```audiounit_output_callback``` to do the in-place downmixing.  Thus, we don't have any extra buffers for downming. However, this only works when converting audio from 3F2(-LFE) to 1F, 2F, 3F, 2F1, 3F1, 2F2 and their LFEs.

The latest modification to add a option ```bool resizing = (in != out);``` in ```downmix_3f2```. If resizing is true, it means that the output buffer has different buffer size from the input's one. For example, if the input buffer is for 3F2-LFE and output buffer is for stereo, then we will have 6 channels for input and 2 channels for output. Therefore, the size of input buffer and output buffer is 6K and 2K, where K is the number of frames. On the other hand, if the resizing is false, it implies we use the same buffer for input and output in in ```downmix_3f2```. If the layout for input and output is 3F2-LFE and stereo, then only the data for left and right channel in this buffer will be updated. The other data will stay same. Another detailed example is in [previous comment](https://github.com/kinetiknz/cubeb/pull/234#issuecomment-281285582).

What do you think? 

p.s. I am not sure if I should use merge or rebase after merging [pull 236](https://github.com/kinetiknz/cubeb/pull/236). I use merge for now. I could change to use rebase if it's more appropriate.

**Comment by achronop** [comment]
> If the layout for input and output is 3F2-LFE and stereo, then only the data for left and right channel in this buffer will be updated. The other data will stay same. Another detailed example is in previous comment.

I do not understand why do we need to change the way that channels are stored in the buffer? Since we are dowmixing there is enough space in the buffer to store channels in a row without gaps. Please correct if I get something wrong.

**Comment by achronop** [comment]
> p.s. I am not sure if I should use merge for depending on pull 236. I could change to use rebase if it's more appropriate.

It is not a mistake to merge, the benefit of rebasing (to latest) is that you will do it anyway, at the end so you have it ready.

**Comment by ChunMinChang** [comment]
> I do not understand why do we need to change the way that channels are stored in the buffer in case of (in == out)? Since we are dowmixing there is enough space in the buffer to store channels in a row without gaps. Please correct me if I get something wrong.

The trick here is that OSX allows us to set the layout to 3F2-LFE with 6 channels when the audio device is stereo, so the ```output_buffer``` returned from ```audiounit_output_callback``` expects a 6 channels' buffer data. That's why we leave the gap in the buffer. 

However, when the output device is stereo, OSX eventually will drop out the channels beyond left and right, so we need to mix the __will-be-dropped__ audio data into the left and right channels.

In other words, when the output device is stereo, we still can set the layout to 3F2-LFE successfully and feed the 6 channel audio data like the following example into the ```output_buffer``` from ```audiounit_output_callback```.
```
+---+---+---+-----+----+----+
| L | R | C | LFE | LS | RS | ...
+---+---+---+-----+----+----+
          ^    ^    ^    ^
          The data for these four channels will be dropped!
```

Nevertheless, only the audio data for left and right will be played. 
```
+---+---+---+---+---+---+
| L | R | 0 | 0 | 0 | 0 | ...
+---+---+---+---+---+---+
          ^   ^   ^   ^
          The played sound of above example is same as this(no sound for non-stereo channels). 
```
Nevertheless, only the audio data for left and right will be played. 

To keep the information in C, LFE, LS and RS channels, we need to mix their data into L and R channels.
```
+----+----+---+---+---+---+
| L' | R' | x | x | x | x | ...
+----+----+---+---+---+---+
            ^   ^   ^   ^
            Useless channel data ...
```
where L' and R' are defined as:
```
L' = L+ 0.707*(C + LS)
R' = R+ 0.707*(C + RS)
```

**Comment by ChunMinChang** [comment]
To leave the gaps(useless data) in the buffer, we add an option ```bool resizing = (in != out);``` in ```downmix_3f2```.

```
+----+----+---+---+---+---+----+----+---+---+---+---+
| L0 | R0 | x | x | x | x | L1 | R1 | x | x | x | x | ...
+----+----+---+---+---+---+----+----+---+---+---+---+
          |<--   gap   -->|         |<--   gap   -->|
```

In the normal case, the downmixed buffer must be __smaller__ than the original buffer because it has __less__ channels. Thus, the ```resizing``` must be true. In our special case on OSX, the buffer size should not be changed because OSX expects a 6 channels buffer in ```audiounit_output_callback```. So we set ```resizing``` to false in this case.

The modification is that we do the __in-place__ downming in the ```output_buffer``` from ```audiounit_output_callback``` when ```resizing``` is false. The in-place method is mentioned in the [previous comment](https://github.com/kinetiknz/cubeb/pull/234#issuecomment-283855601). We change the data from
```
+---+---+---+-----+----+----+
| L | R | C | LFE | LS | RS | ...
+---+---+---+-----+----+----+
          ^    ^    ^    ^
          The data for these four channels will be dropped!
```
into
```
          |<--   gap   -->|
+----+----+---+---+---+---+
| L' | R' | x | x | x | x | ...
+----+----+---+---+---+---+
            ^   ^   ^   ^
            Useless channel data ...
```

**Comment by ChunMinChang** [comment]
In fact, the option ```resizing```  could be removed. We could do downmixing as usual and then expand the buffer to the format that the OSX expects.

```
+----+----+----+------+-----+-----+----+----+----+------+-----+-----+
| L0 | R0 | C0 | LFE0 | LS0 | RS0 | L1 | R1 | C1 | LFE1 | LS1 | RS1 | ......
+----+----+----+------+-----+-----+----+----+----+------+-----+-----+

   | Downmix
   v

+----+----+----+----+
| L0 | R0 | L1 | R1 | ...
+----+----+----+----+

   | Expand
   v

+----+----+---+---+---+---+----+----+---+---+---+---+
| L0 | R0 | x | x | x | x | L1 | R1 | x | x | x | x | ...
+----+----+---+---+---+---+----+----+---+---+---+---+
          |<--   gap   -->|         |<--   gap   -->|
```

But it will cost an extra ```O(m)```, where ```m``` is the length of the buffer. If it's ok, then I can add a patch to implement it.

**Comment by achronop** [comment]
Thank you for your detailed explanation. I guess that in OSX this is always the case, by this I mean the fact that we need to have gaps in an expanded buffer. If that's true I see 2 options. 

The first is the one you have proposed in your last comment. I guess the expand will happen in audiounit not in mixer. This has off course the extra cost. 

The second is to create an version of `downmix_3f2()` method specific for the osx. Then use the `__APPLE__` (or any other suitable) macro to use that new method when building on OSX. In this case you can do in-place conversion without the extra cost.

**Comment by achronop** [comment]
For the second option you could use the same macro is used by `cubeb_init()` method in case of audiounit: `#if defined(USE_AUDIOUNIT)`

File: src/cubeb_mixer.cpp [hunk 5]
```diff
@@ -214,14 +238,96 @@ downmix_3f2(T const * const in, unsigned long inframes, T * out, cubeb_channel_l
...
+//
+// |<---------- 1 ---------->|<---------- 2 ---------->|
+// +----+----+---+---+---+---+----+----+---+---+---+---+
+// | L0 | R0 | x | x | x | x | L1 | R1 | x | x | x | x | ...
+// +----+----+---+---+---+---+----+----+---+---+---+---+
+//           |<--   gap   -->|         |<--   gap   -->|
+//
+// where L0 = L0 + 0.707 * (C0 + LS0), R0 = R0 + 0.707 * (C0 + RS0),
+//       L1 = L1 + 0.707 * (C1 + LS1), R1 = R1 + 0.707 * (C1 + RS1), ...
+//
```
achronop: All comments above are super explanatory and useful but tend to be really long. Can we have a shorter version of them?

src/cubeb_mixer.cpp
```diff
@@ -214,14 +238,96 @@ downmix_3f2(T const * const in, unsigned long inframes, T * out, cubeb_channel_l
...
+
+  // Conversion from 3F2 to 2F2-LFE or 3F1-LFE is allowed, so we use '<=' instead of '<'.
+  assert(out_channels <= in_channels);
 
-  unsigned long out_index = 0;
   auto & downmix_matrix = DOWNMIX_MATRIX_3F2_LFE[out_layout - CUBEB_LAYOUT_MONO]; // The matrix is started from mono.
-  for (unsigned long i = 0; i < inframes * in_channels; i += in_channels) {
+  // The difference between this method and the normal version is that we use
+  // "out_index += in_channels" instead of "out_index += out_channels".
+  for (unsigned long i = 0, out_index = 0; i < inframes * in_channels; i += in_channels, out_index += in_channels) {
```
achronop: I know you had proposed to add between the #ifdef/#end this one line and I disagreed. Now that I see it I believe it would be better. Can we do that now? Sorry for the inconvenience!

src/cubeb_mixer.cpp
```diff
@@ -312,15 +428,16 @@ cubeb_downmix(T const * const in, long inframes, T * out,
       return;
     }
 
-    if (in == out || mix_remap(in, inframes, out, in_layout, out_layout)) {
+  // We only support downmix from audio 5.1 for now.
+#if defined(USE_AUDIOUNIT)
+  return;
+#endif
```
achronop: Do we plan to support more combinations than 5.1?

src/cubeb_mixer.cpp
ChunMinChang: OK. I'll push a shorter version for it.

src/cubeb_mixer.cpp
ChunMinChang: OK, I'll change it back.

src/cubeb_mixer.cpp
ChunMinChang: I think we could decide it after we get the result from telemetry.

ChunMinChang:
tryserver: https://treeherder.mozilla.org/#/jobs?repo=try&revision=5fa3620ccf7b865ad1a82846e5644a7500c18b30&selectedJob=83618345

achronop:
Rebased and merged in 4ab4577

PR #235, opened 2017-02-17
audiounit: additional fix for drift cases

achronop:
Resolve more drift scenarios. This patch is complementary to the previous one - 581d4bc


achronop:
Just to add I have tested it all combinations of user rate, device rate with c920, c930 and Plantronics 648 for more than 10mins in each call.

src/cubeb_resampler.cpp
```diff
@@ -72,6 +72,7 @@ long passthrough_resampler<T>::fill(void * input_buffer, long * input_frames_cou
 
   if (input_buffer) {
     internal_input_buffer.pop(nullptr, frames_to_samples(output_frames));
+    *input_frames_count = output_frames;
```
padenot: I'd like to remove this output param at some point, but we can discuss later.

PR #236, opened 2017-02-20
Prevent deadlock by avoiding creating AU when getting channel layout

ChunMinChang:
Per [discussion in pull 234](https://github.com/kinetiknz/cubeb/pull/234#discussion_r101745194), we should avoid to create another AU when we already have one in cubeb_stream, or it might cause a deadlock.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -72,7 +72,7 @@ static int audiounit_setup_stream(cubeb_stream *stm);
 static int audiounit_create_unit(AudioUnit * unit, bool is_input,
                                  const cubeb_stream_params * /* stream_params */,
                                  AudioDeviceID device);
-static cubeb_channel_layout audiounit_get_channel_layout(AudioUnit & output_unit);
+static cubeb_channel_layout audiounit_get_current_channel_layout(AudioUnit * output_unit);
 static cubeb_channel_layout audiounit_get_preferred_channel_layout();
 
 extern cubeb_ops const audiounit_ops;
@@ -1084,7 +1084,7 @@ audiounit_get_preferred_channel_layout(cubeb * ctx, cubeb_channel_layout * layou
   // If the preferred channel layout is UNDEFINED, then we try access the
   // current applied channel layout.
   if (*layout == CUBEB_LAYOUT_UNDEFINED) {
-    // We will create a new AU below to use audiounit_get_channel_layout,
+    // We will create a new AU below to use audiounit_get_current_channel_layout,
     // which shares the same mutex with data callback,
     // so we need to prevent creating another AU when we already have one.
     if (ctx->active_streams) {
@@ -1094,7 +1094,7 @@ audiounit_get_preferred_channel_layout(cubeb * ctx, cubeb_channel_layout * layou
     // Get the default ouput unit
     AudioUnit output_unit;
     audiounit_create_unit(&output_unit, false, nullptr, 0);
-    *layout = audiounit_get_channel_layout(output_unit);
+    *layout = audiounit_get_current_channel_layout(&output_unit);
   }
 
   if (*layout == CUBEB_LAYOUT_UNDEFINED) {
@@ -2783,12 +2783,12 @@ audiounit_get_preferred_channel_layout()
 }
 
 static cubeb_channel_layout
-audiounit_get_channel_layout(AudioUnit & output_unit)
+audiounit_get_current_channel_layout(AudioUnit * output_unit)
 {
   OSStatus rv = noErr;
 
   UInt32 size = 0;
-  rv = AudioUnitGetPropertyInfo(output_unit,
+  rv = AudioUnitGetPropertyInfo(*output_unit,
                                 kAudioUnitProperty_AudioChannelLayout,
                                 kAudioUnitScope_Output,
                                 AU_OUT_BUS,
@@ -2801,7 +2801,7 @@ audiounit_get_channel_layout(AudioUnit & output_unit)
   assert(size > 0);
 
   auto_channel_layout layout(size);
-  rv = AudioUnitGetProperty(output_unit,
+  rv = AudioUnitGetProperty(*output_unit,
                             kAudioUnitProperty_AudioChannelLayout,
                             kAudioUnitScope_Output,
                             AU_OUT_BUS,
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -72,8 +72,6 @@ static int audiounit_setup_stream(cubeb_stream *stm);
 static int audiounit_create_unit(AudioUnit * unit, bool is_input,
                                  const cubeb_stream_params * /* stream_params */,
                                  AudioDeviceID device);
-static cubeb_channel_layout audiounit_get_current_channel_layout(AudioUnit * output_unit);
-static cubeb_channel_layout audiounit_get_preferred_channel_layout();
 
 extern cubeb_ops const audiounit_ops;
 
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -72,7 +72,8 @@ static int audiounit_setup_stream(cubeb_stream *stm);
 static int audiounit_create_unit(AudioUnit * unit, bool is_input,
                                  const cubeb_stream_params * /* stream_params */,
                                  AudioDeviceID device);
-static cubeb_channel_layout audiounit_get_channel_layout(bool preferred);
+static cubeb_channel_layout audiounit_get_channel_layout(AudioUnit & output_unit);
+static cubeb_channel_layout audiounit_get_preferred_channel_layout();
```
achronop: Why the functionality is broken into 2 methods. Both methods are used once in the same place. Would it be more simple to have just one method?

src/cubeb_audiounit.cpp
```diff
@@ -1072,16 +1073,28 @@ audiounit_get_preferred_sample_rate(cubeb * /* ctx */, uint32_t * rate)
...
   // For default output on Mac, there is no preferred channel layout,
   // so it might return UNDEFINED.
-  // In that case, we should get the channel configuration directly.
-  *layout = audiounit_get_channel_layout(true);
+  *layout = audiounit_get_preferred_channel_layout();
+
+  // If the preferred channel layout is UNDEFINED, then we try access the
+  // current applied channel layout.
   if (*layout == CUBEB_LAYOUT_UNDEFINED) {
-    *layout = audiounit_get_channel_layout(false);
```
achronop: This is not ideal and I know it is not your fault. The problem is, when one stream is operating we cannot get the preferred layout for any other independent new stream. 

One option is to add a new member in cubeb struct to store the stream layout. Every time we create/configure an output stream we update this cubeb member. If any active stream exists provide the layout from this data member if not create a new AU and get the value.

The second option is to simple create the AU every time which means we must be very careful not to use this api call under a mutex in a way that we create a deadlock.

If we know how often the `audiounit_get_preferred_channel_layout()` method fails would help to get a decision.  For example, if it is unlikely to fail then we can follow the simplistic method. If it fails in half of the cases the first option would be more appropriate.

src/cubeb_audiounit.cpp
```diff
@@ -2723,66 +2736,83 @@ int audiounit_register_device_collection_changed(cubeb * context,
...
+  AudioDeviceID id;
+
+  if (audiounit_get_output_device_id(&id) != CUBEB_OK) {
+    return CUBEB_LAYOUT_UNDEFINED;
+  }
+
+  AudioObjectPropertyAddress adr = { kAudioDevicePropertyPreferredChannelLayout,
+                                     kAudioObjectPropertyScopeOutput,
+                                     kAudioObjectPropertyElementMaster };
+  if (AudioObjectGetPropertyDataSize(id, &adr, 0, NULL, &size) != noErr) {
```
achronop: Nit: The code style is to declare an `OSStatus rv = noErr;` and check `if (rv != noErr)`. Following the same code style everywhere adds in readability. The same exist in other places.

src/cubeb_audiounit.cpp
ChunMinChang: The ```audiounit_get_channel_layout``` is prepared for calling when we already have a cubeb_stream like what we mentioned in [pull 234](https://github.com/ChunMinChang/cubeb/blob/0cb79ed2234c6387bbfc8cf2c00fece092fe4648/src/cubeb_audiounit.cpp#L535). 

```audiounit_get_channel_layout``` returns the current used layout by ```kAudioUnitProperty_AudioChannelLayout```, while ```audiounit_get_preferred_channel_layout``` returns the device's preferred layout by getting ```kAudioDevicePropertyPreferredChannelLayout```. The preferred layout and current used layout is two different value, so I implement two functions to call it.

To differentiate these two functions, it might be helpful by renaming ```audiounit_get_channel_layout``` to something like ```audiounit_get_current_channel_layout```.

File: src/cubeb_audiounit.cpp (hunk 2 above)
Comment (ChunMinChang) [inline src/cubeb_audiounit.cpp]: To avoid deadlock, I prefer first option. We could add a member in cubeb context to track the current layout. If there is no active stream, then it's ok to create one to get the current used layout. If we already have existing stream, then the current used layout must be updated.

File: src/cubeb_audiounit.cpp (hunk 3 above)
Comment (ChunMinChang) [inline src/cubeb_audiounit.cpp]: OK, I'll change the style back.

File: src/cubeb_audiounit.cpp [hunk 4]
```diff
@@ -1168,10 +1250,11 @@ audio_stream_desc_init(AudioStreamBasicDescription * ss,
 }
 
 static int
-audiounit_layout_init(AudioUnit * unit,
-                      io_side side,
-                      const cubeb_stream_params * stream_params)
+audiounit_set_channel_layout(AudioUnit * unit,
```
achronop: AudioUnit can be passed by value.

src/cubeb_audiounit.cpp
```diff
@@ -1225,22 +1308,36 @@ audiounit_layout_init(AudioUnit * unit,
...
 
   return CUBEB_OK;
 }
 
+void
+audiounit_layout_init(cubeb_stream * stm, io_side side)
+{
+  // We currently don't support the input layout setting.
+  if (side == INPUT) {
+    return;
```
achronop: Please log a message to indicate the case.

ChunMinChang:
Yes, I already pushed three tests to tryserver. 
1. Current modification
  - https://treeherder.mozilla.org/#/jobs?repo=try&revision=8cb99df0a9c33180333bc272dcc4a7a3373189e2
2. Current modification with undoing changes of [bug 1337805](https://bugzilla.mozilla.org/show_bug.cgi?id=1337805)
  - https://treeherder.mozilla.org/#/jobs?repo=try&revision=a1108ea4ea89eaffbe5c9731b7f68305d01fc424
3. Current patches with undoing changes of [bug 1337805](https://bugzilla.mozilla.org/show_bug.cgi?id=1337805) but without [```active_stream``` part](https://hg.mozilla.org/try/rev/ee1d37de083bf7f065103f0ccdc18591ce91d044)
  - https://treeherder.mozilla.org/#/jobs?repo=try&revision=ee1d37de083bf7f065103f0ccdc18591ce91d044

If we remove the ```active_stream``` part, I guess we could get the similar result of [bug 1337805](https://bugzilla.mozilla.org/show_bug.cgi?id=1337805). That's why I append the revert changes of its patches. The [result of test 3](https://treeherder.mozilla.org/#/jobs?repo=try&revision=ee1d37de083bf7f065103f0ccdc18591ce91d044&selectedJob=79554808) shows we have a [deadlock](https://archive.mozilla.org/pub/firefox/try-builds/cchang@mozilla.com-ee1d37de083bf7f065103f0ccdc18591ce91d044/try-macosx64-debug/try_yosemite_r7-debug_test-mochitest-gl-e10s-2-bm107-tests1-macosx-build7.txt.gz), so we indeed need use ```active_stream``` part to avoid it.

However, I am waiting the results of test 1 and 2. I'll update here ASAP.

achronop:
@ChunMinChang That's perfect thank you very much! 

ps. I don't think the try errors are due to our code.

ChunMinChang:
@achronop Thanks for your guidance and support : )

ok, maybe I need to do more investigation!

achronop:
BTW the test code you wrote is exactly what we need to verify (without the try server pain) most platform deadlocks. I'll copy locally and it will be nice to find a way to add it in automated tested in the future.

PR #237, opened 2017-02-20
Fixes for two crashes

padenot:
1. Don't try to re-do an emergency bailout after having done one: we don't know whether the bool has been freed or not. This has been found in the field, https://bugzilla.mozilla.org/show_bug.cgi?id=1340291. This is rather rare. One day we'll have proper atomic refptr of atomic memory maybe... 
2. `stm->output_client` and `stm->input_client` should be accessed only when the lock has been taken. During a reconfigure, we null them out ([here](https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_wasapi.cpp#L1797)). Simply taking the lock before doing the assert should be enough. This is very rare, captured in bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1340278

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -1893,10 +1893,10 @@ int stream_start_one_side(cubeb_stream * stm, StreamDirection dir)
 
 int wasapi_stream_start(cubeb_stream * stm)
 {
+  auto_lock lock(stm->stream_reset_lock);
+
   XASSERT(stm && !stm->thread && !stm->shutdown_event);
   XASSERT(stm->output_client || stm->input_client);
-
-  auto_lock lock(stm->stream_reset_lock);
 
   stm->emergency_bailout = new std::atomic<bool>(false);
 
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -1166,12 +1171,16 @@ bool stop_and_join_render_thread(cubeb_stream * stm)
     /* Something weird happened, leak the thread and continue the shutdown
      * process. */
     *(stm->emergency_bailout) = true;
+    // We give the ownership to the rendering thread.
+    stm->emergency_bailout.store(nullptr);
```
achronop: Since now `stm->emergency_bailout` can be null it might be a good idea to assert in line 833 that you got a non null pointer. I know that we can only hit it in extreme race conditions probably after very fast start and stop. I am not very familiar with wasapi weirdness. If you think it is not applicable ignore the comment.

padenot: This can happen in the following scenario:
- `cubeb_stream_start` is called, the rendering thread does not start immediately
- `cubeb_stream_stop` is called, and we can't join the thread because it's not being scheduled for 5 seconds, we null out the `emergency_bailout` pointer.
- The rendering thread finally start, gets the value, it's null, goes to the `WaitForMultipleEvents`, returns with a `shutdown_event`, dereferences the `emergency_bailout` ptr, that is `nullptr`.

Really if this happens, we should simply return from the rendering thread.

PR #239, opened 2017-02-21
api: Give cubeb_init a third param, backend name (Bug 1341238)

zamaudio:
This allows forcing of a particular backend by name,
*if* it was compiled in, otherwise default list is tried
in default order as before.

Tests updated to reflect new api.  (Gecko change required)

100% tests passed (using JACK backend)

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -414,10 +414,16 @@ typedef void (* cubeb_log_callback)(char const * fmt, ...);
                    context will be returned.
     @param context_name A name for the context. Depending on the platform this
                         can appear in different locations.
+    @param backend_name The name of the cubeb backend user desires to select.
+                        Accepted values self-documented in cubeb.c: init_oneshot
+                        If NULL, a default ordering is used for backend choice.
+                        A valid choice overrides all other possible backends,
+                        so long as the backend was included at compile time.
     @retval CUBEB_OK in case of success.
     @retval CUBEB_ERROR in case of error, for example because the host
                         has no audio hardware. */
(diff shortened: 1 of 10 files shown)


zamaudio:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1341238 for gecko side patch

src/cubeb.c
```diff
@@ -105,9 +106,57 @@ validate_latency(int latency)
 }
 
 int
-cubeb_init(cubeb ** context, char const * context_name)
+cubeb_init(cubeb ** context, char const * context_name, char const * backend_name)
 {
-  int (* init[])(cubeb **, char const *) = {
+  int (* init_oneshot)(cubeb **, char const *) = NULL;
```
kinetiknz: I think it'd be simpler if init_oneshot was a single element array so the code below can just switch on using default_init or init_oneshot in the loop rather than duplicating a bunch of code.

include/cubeb/cubeb.h
```diff
@@ -414,10 +414,13 @@ typedef void (* cubeb_log_callback)(char const * fmt, ...);
                    context will be returned.
     @param context_name A name for the context. Depending on the platform this
                         can appear in different locations.
+    @param backend_name The name of the forced backend.  This has a list of
+                        accepted values self-documented in cubeb.c
```
kinetiknz: Be more specific about the behaviour when a backend_name is specified, both when it's compiled in and when it isn't.

zamaudio:
Thanks, I addressed all of your points and force-pushed above ^ (new gecko patch in bugzilla)

kinetiknz:
Thanks!

PR #247, opened 2017-03-06
audiounit: aggregate devices follow up

achronop:
The patch addresses the follow up comments of PR #231

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1480,13 +1480,10 @@ get_device_name(AudioDeviceID id)
 }
 
 static int
-audiounit_set_aggregate_sub_device_list(cubeb_stream * stm)
+audiounit_set_aggregate_sub_device_list(AudioDeviceID aggregate_device_id, AudioDeviceID input_device_id,
+                                        AudioDeviceID output_device_id)
 {
-  AudioDeviceID input_device_id = audiounit_get_input_device_id(stm);
   const std::vector<AudioDeviceID> input_sub_devices = audiounit_get_sub_devices(input_device_id);
-
-
-  AudioDeviceID output_device_id = audiounit_get_default_device_id(CUBEB_DEVICE_TYPE_OUTPUT);
   const std::vector<AudioDeviceID> output_sub_devices = audiounit_get_sub_devices(output_device_id);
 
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -1451,12 +1406,7 @@ audiounit_create_blank_aggregate_device(AudioObjectID * plugin_id, AudioDeviceID
   CFDictionaryAddValue(aggregate_device_dict, CFSTR(kAudioAggregateDeviceIsPrivateKey), aggregate_device_private_key);
   CFRelease(aggregate_device_private_key);
 
-  r = AudioObjectGetPropertyData(*plugin_id,
-                                 &create_aggregate_device_address,
-                                 sizeof(aggregate_device_dict),
-                                 &aggregate_device_dict,
-                                 &size,
-                                 aggregate_device_id);
+  OSStatus r = AudioHardwareCreateAggregateDevice( aggregate_device_dict, aggregate_device_id);
```
kinetiknz: No space between ( and aggregate_device_dict

src/cubeb_audiounit.cpp
```diff
@@ -1480,13 +1430,10 @@ get_device_name(AudioDeviceID id)
 }
 
 static int
-audiounit_set_aggregate_sub_device_list(cubeb_stream * stm)
+audiounit_set_aggregate_sub_device_list(AudioDeviceID aggregate_device_id, AudioDeviceID input_device_id,
+                                        AudioDeviceID output_device_id)
```
kinetiknz: When you have to wrap function params, it's more readable if you put one on each line.

achronop:
Thanks for the review.

achronop:
It does not build for OSX 10.7 (as expected) because AudioHardwareCreateAggregateDevice and AudioHardwareDestroyAggregateDevice do not exist there.
https://treeherder.mozilla.org/#/jobs?repo=try&revision=89cd68a8f11981106a12d0a13659a88694168afe

You mentioned we do not need to support the audiounit backend for versions earlier that 10.9. What do you suggest for those versions?

kinetiknz:
It looks like our builders are still on 10.7 and building against the 10.7 API, despite Firefox supporting 10.9 as the minimum since 48 in August 2016.  I guess we'll have to use the old code paths until either bug 1270217 is fixed  or the builders are migrated to >= 10.9.

PR #248, opened 2017-03-06
audiounit: On unexpected number of returned frames go to error

achronop:
In audio callback, when returned frames are less that the expected we go to Drain. However if return frames are less than zero or more than expected we should go to error.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -550,8 +550,16 @@ audiounit_output_callback(void * user_ptr,
     stm->input_linear_buffer->pop(input_frames_before_fill * stm->input_desc.mChannelsPerFrame);
   }
 
-  if (outframes < 0) {
+  if (outframes < 0 || outframes > output_frames) {
     stm->shutdown = true;
+    OSStatus r = AudioOutputUnitStop(stm->output_unit);
+    assert(r == 0);
+    if (stm->input_unit) {
+      r = AudioOutputUnitStop(stm->input_unit);
+      assert(r == 0);
+    }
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -550,8 +550,16 @@ audiounit_output_callback(void * user_ptr,
     stm->input_linear_buffer->pop(input_frames_before_fill * stm->input_desc.mChannelsPerFrame);
   }
 
-  if (outframes < 0) {
+  if (outframes < 0 || outframes > output_frames) {
```
kinetiknz: When is `outframes > output_frames` true except when there's a bug?  It seems like it'd be better to assert so the failure is as close to the source of the bug as possible.

achronop: That said, I would expect that the same could occur for `outframes < 0`. I cannot think of any valid case that the user callback returns negative number of frames.

kinetiknz: Well, the question is: what motivated this change? A specific instance of this triggering, or that you just noticed that the error checking was incomplete?

Assert for both cases.  And make the same fix for all backends for consistency.

We probably need a to add debug-only assert that's non-fatal in release builds and maps to MOZ_ASSERT in Gecko and use that for this, falling back to the error handling for release builds.

achronop:
> Well, the question is: what motivated this change? A specific instance of this triggering, or that you just noticed that the error checking was incomplete?

Just noticed the incompleted check. I was wondering the same think during the development of full duplex for Android in which we decided to go to error so I did the same here.

achronop:
#256 raised to track that.

PR #249, opened 2017-03-06
audiounit: add synchronization in cubeb destroy (Bug 1344580)

achronop:
I am not able to repro the crash but I believe this patch will solve it. The truth is that we do not solve the root cause that would be not to call cubeb_destroy before every stream is stopped and destroyed (bug 1083664). In any case this patch fixes other cases like to destroy during device switch. 

Try link:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=1f819e60bed151b4ed84536aefc1c8c076b21cab

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1212,7 +1212,7 @@ audiounit_destroy(cubeb * ctx)
   {
     auto_lock lock(ctx->mutex);
     /* Unregister the callback if necessary. */
-    if(ctx->collection_changed_callback) {
+    if (ctx->collection_changed_callback) {
       audiounit_remove_device_listener(ctx);
     }
   }
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1209,10 +1209,12 @@ audiounit_destroy(cubeb * ctx)
     LOG("(%p) API misuse, %d streams active when context destroyed!", ctx, ctx->active_streams.load());
   }
 
-  /* Unregister the callback if necessary. */
-  if(ctx->collection_changed_callback) {
+  {
     auto_lock lock(ctx->mutex);
-    audiounit_remove_device_listener(ctx);
+    /* Unregister the callback if necessary. */
+    if(ctx->collection_changed_callback) {
+      audiounit_remove_device_listener(ctx);
+    }
   }
 
   delete ctx;


src/cubeb_audiounit.cpp
```diff
@@ -1203,16 +1203,18 @@ static OSStatus audiounit_remove_device_listener(cubeb * context);
...
+    // Disabling this assert for bug 1083664 -- we seem to leak a stream
+    // assert(ctx->active_streams == 0);
+    if (ctx->active_streams > 0) {
+      LOG("(%p) API misuse, %d streams active when context destroyed!", ctx, ctx->active_streams.load());
+    }
+
+    /* Unregister the callback if necessary. */
+    if(ctx->collection_changed_callback) {
+      audiounit_remove_device_listener(ctx);
+    }
```
kinetiknz: active_streams is atomic, so it seems like the only change effected by this modification is protecting the access of ctx->collection_changed_callback.  I can't see how that can fix bug 1344580.

achronop: I believe that crash of bug 1344580 is because `stm->context` became invalid during stream setup. If that happened due to a call in `cubeb_destroy` synchronization will save us from crashing. 

Even if this is not the case here the fix is valid. I am able to repro a crash when unplug a headset and call `cubeb_destroy` simultaneously. I could remove the reference to the bug.

kinetiknz: > I believe that crash of bug 1344580 is because stm->context became invalid during stream setup. If that happened due to a call in cubeb_destroy synchronization will save us from crashing.

It's an invalid state to have any streams still alive when cubeb_destroy is called.  If that's happening, the API misuse in the caller needs to be fixed.

src/cubeb_audiounit.cpp
```diff
@@ -1209,10 +1209,12 @@ audiounit_destroy(cubeb * ctx)
     LOG("(%p) API misuse, %d streams active when context destroyed!", ctx, ctx->active_streams.load());
   }
 
-  /* Unregister the callback if necessary. */
-  if(ctx->collection_changed_callback) {
+  {
     auto_lock lock(ctx->mutex);
-    audiounit_remove_device_listener(ctx);
+    /* Unregister the callback if necessary. */
+    if(ctx->collection_changed_callback) {
```
kinetiknz: Space between if and (

src/cubeb_audiounit.cpp
achronop: FWIW there were an assert to catch that case in the past, [here](https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_audiounit.cpp#L1176), but got disabled due to bug 1083664. I would love to re-enable that one. I did a try run and it is green (from that assert)

https://treeherder.mozilla.org/#/jobs?repo=try&revision=f2ba9281fdea9e8b25ca8dcc9012a2a7aa727079

PR #250, opened 2017-03-07
Getting correct size of allocated AudioChannelLayout array

ChunMinChang:
Instead of ```sizeof(*layout)```, we should keep track of the size for using malloc.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -147,10 +147,11 @@ public:
   {
   }
 
-  auto_channel_layout(size_t size)
-    : layout(reinterpret_cast<AudioChannelLayout*>(malloc(size)))
+  auto_channel_layout(size_t s)
+    : layout(reinterpret_cast<AudioChannelLayout*>(malloc(s)))
   {
-    memset(layout, 0, size);
+    sz = s;
+    memset(layout, 0, sz);
   }
 
   ~auto_channel_layout()
@@ -158,11 +159,12 @@ public:
     release();
   }
 
-  void reset(size_t size)
+  void reset(size_t s)
   {
     release();
-    layout = reinterpret_cast<AudioChannelLayout*>(malloc(size));
-    memset(layout, 0, size);
+    sz = s;
+    layout = reinterpret_cast<AudioChannelLayout*>(malloc(sz));
+    memset(layout, 0, sz);
   }
 
   AudioChannelLayout* get()
@@ -172,7 +174,7 @@ public:
 
   size_t size()
   {
-    return sizeof(*layout);
+    return sz;
   }
 
 private:
@@ -180,11 +182,13 @@ private:
   {
     if (layout) {
       free(layout);
+      sz = 0;
       layout = NULL;
     }
   }
 
   AudioChannelLayout * layout;
+  size_t sz;
 };
 
 enum io_side {
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -139,12 +139,13 @@ private:
   auto_array<T> ar;
 };
 
-template<typename T>
-using unique_ptr_with_free = std::unique_ptr<T, decltype(free)*>;
-
-template<typename T>
-inline unique_ptr_with_free<T> make_alloc_unique(size_t sz) {
-    return unique_ptr_with_free<T> { (T*)calloc(1, sz), free };
+static std::unique_ptr<AudioChannelLayout, decltype(&free)>
+make_sized_audio_channel_layout(size_t sz)
+{
+    assert(sz >= sizeof(AudioChannelLayout));
+    AudioChannelLayout * acl = reinterpret_cast<AudioChannelLayout *>(calloc(1, sz));
+    assert(acl); // Assert the allocation works.
+    return std::unique_ptr<AudioChannelLayout, decltype(&free)>(acl, free);
 }
 
 enum io_side {
@@ -1080,7 +1081,7 @@ audiounit_get_current_channel_layout(AudioUnit output_unit)
   }
   assert(size > 0);
 
-  auto layout = make_alloc_unique<AudioChannelLayout>(size);
+  auto layout = make_sized_audio_channel_layout(size);
   rv = AudioUnitGetProperty(output_unit,
                             kAudioUnitProperty_AudioChannelLayout,
                             kAudioUnitScope_Output,
@@ -1115,7 +1116,7 @@ audiounit_get_preferred_channel_layout()
   }
   assert(size > 0);
 
-  auto layout = make_alloc_unique<AudioChannelLayout>(size);
+  auto layout = make_sized_audio_channel_layout(size);
   rv = AudioObjectGetPropertyData(id, &adr, 0, NULL, &size, layout.get());
   if (rv != noErr) {
     return CUBEB_LAYOUT_UNDEFINED;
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -147,22 +147,24 @@ class auto_channel_layout
   {
   }
 
-  auto_channel_layout(size_t size)
-    : layout(reinterpret_cast<AudioChannelLayout*>(malloc(size)))
+  auto_channel_layout(size_t s)
+    : layout(reinterpret_cast<AudioChannelLayout*>(malloc(s)))
```
kinetiknz: Initialize sz in the member initialization list.

If you avoided renaming size -> s it'd make the patch smaller and easier to understand.

Also, the malloc could be a calloc and not need the memset(0).

It looks like auto_channel_layout could be replaced with a unique_ptr to simplify the code.  There's only one place you need the result of size() and the size is available in a local variable at that point anyway (and using that would've avoided this bug).

src/cubeb_audiounit.cpp
```diff
@@ -172,19 +174,21 @@ class auto_channel_layout
...
+    return sz;
   }
 
 private:
   void release()
   {
     if (layout) {
       free(layout);
+      sz = 0;
       layout = NULL;
```
kinetiknz: nullptr

src/cubeb_audiounit.cpp
```diff
@@ -147,22 +147,24 @@ class auto_channel_layout
...
   }
 
-  auto_channel_layout(size_t size)
-    : layout(reinterpret_cast<AudioChannelLayout*>(malloc(size)))
+  auto_channel_layout(size_t s)
+    : layout(reinterpret_cast<AudioChannelLayout*>(malloc(s)))
   {
-    memset(layout, 0, size);
+    sz = s;
+    memset(layout, 0, sz);
```
kinetiknz: If malloc failed, this memset would be a nullptr deref

src/cubeb_audiounit.cpp
```diff
@@ -147,22 +147,24 @@ class auto_channel_layout
...
 
-  void reset(size_t size)
+  void reset(size_t s)
   {
     release();
-    layout = reinterpret_cast<AudioChannelLayout*>(malloc(size));
-    memset(layout, 0, size);
+    sz = s;
+    layout = reinterpret_cast<AudioChannelLayout*>(malloc(sz));
+    memset(layout, 0, sz);
```
kinetiknz: As above, doesn't handle malloc failure

ChunMinChang:
I fixed the code by your comments and I tried to use ```unique_ptr``` but I am not sure if it's an appropriate usage.

src/cubeb_audiounit.cpp
```diff
@@ -139,53 +139,13 @@ struct auto_array_wrapper_impl : public auto_array_wrapper {
...
 
-class auto_channel_layout
-{
-public:
-  auto_channel_layout()
-    : layout(nullptr)
-  {
-  }
+template<typename T>
+using unique_ptr_with_free = std::unique_ptr<T, decltype(free)*>;
```
kinetiknz: You only use this alias in two places, so it'd be simpler to leave it out.

src/cubeb_audiounit.cpp
```diff
@@ -139,53 +139,13 @@ struct auto_array_wrapper_impl : public auto_array_wrapper {
...
-      layout = NULL;
-    }
-  }
-
-  AudioChannelLayout * layout;
-};
+template<typename T>
+inline unique_ptr_with_free<T> make_alloc_unique(size_t sz) {
+    return unique_ptr_with_free<T> { (T*)calloc(1, sz), free };
+}
```
kinetiknz: The `inline` is unnecessary, it's better not to add it unless you have a good reason.  The function should be `static`, though.  Also, code style is to put the return type on a separate line from the function name.

This function is quite unusual/specialized, so it's probably better not to make it generic over the type to reduce the chance of it being misused/reappropriated.  You should also assert that sz is at least sizeof(AudioChannelLayout), since any smaller request must be a bug.

Without the type alias, I'd write this as:

    std::unique_ptr<AudioChannelLayout, decltype(&free)>
    make_sized_audio_channel_layout(size_t sz)
    {
        assert(sz >= sizeof(AudioChannelLayout));
        AudioChannelLayout * acl = reinterpret_cast<AudioChannelLayout *>(calloc(1, sz));
        return std::unique_ptr<AudioChannelLayout, decltype(&free)>(acl, free);
    }

src/cubeb_audiounit.cpp
```diff
@@ -1120,7 +1080,7 @@ audiounit_get_current_channel_layout(AudioUnit output_unit)
   }
   assert(size > 0);
 
-  auto_channel_layout layout(size);
+  auto layout = make_alloc_unique<AudioChannelLayout>(size);
```
kinetiknz: Handle (or assert) case where layout is nullptr (e.g. alloc failure)

src/cubeb_audiounit.cpp
```diff
@@ -1155,7 +1115,7 @@ audiounit_get_preferred_channel_layout()
...
```
kinetiknz: Handle (or assert) case where layout is nullptr (e.g. alloc failure)

src/cubeb_audiounit.cpp
```diff
@@ -1274,7 +1234,8 @@ audiounit_set_channel_layout(AudioUnit unit,
   assert(stream_params->channels == CUBEB_CHANNEL_LAYOUT_MAPS[stream_params->layout].channels);
 
   OSStatus r;
-  auto_channel_layout layout(sizeof(AudioChannelLayout));
+  size_t size = sizeof(AudioChannelLayout);
+  auto layout = make_alloc_unique<AudioChannelLayout>(size);
```
kinetiknz: Just pass the sizeof result directly without the temp variable

src/cubeb_audiounit.cpp
```diff
@@ -1310,8 +1271,8 @@ audiounit_set_channel_layout(AudioUnit unit,
   // For those layouts that can't be matched to coreaudio's predefined layout,
   // we use customized layout.
   if (layout.get()->mChannelLayoutTag == kAudioChannelLayoutTag_Unknown) {
-    size_t size = offsetof(AudioChannelLayout, mChannelDescriptions[stream_params->channels]);
-    layout.reset(size);
+    size = offsetof(AudioChannelLayout, mChannelDescriptions[stream_params->channels]);
+    layout.reset((AudioChannelLayout*)calloc(1, size));
```
kinetiknz: Rather than using reset explicitly and exposing/duplicating the allocator logic, just do:

  layout = make_sized_audio_channel_layout(size);

src/cubeb_audiounit.cpp
kinetiknz: (and handle alloc failure)

src/cubeb_audiounit.cpp
kinetiknz: (and handle alloc failure)

src/cubeb_audiounit.cpp
```diff
@@ -1274,7 +1235,8 @@ audiounit_set_channel_layout(AudioUnit unit,
   assert(stream_params->channels == CUBEB_CHANNEL_LAYOUT_MAPS[stream_params->layout].channels);
 
   OSStatus r;
-  auto_channel_layout layout(sizeof(AudioChannelLayout));
+  size_t size = 0; // used when we need to customize the layout.
```
kinetiknz: Oh, sorry, it was right before, because you need it to be initialized to pass it to AudioUnitSetProperty!  I'll fix and merge.

kinetiknz:
Thanks! Pushed as f1a6eed0ef1829ea3ac589feefa0f46bf594cdaa.

ChunMinChang:
Thanks for your guidance!

PR #251, opened 2017-03-07
audiounit: preserve volume after stream restart (Bug 1343920)

achronop:
Get the volume level at the begin of reinit and set it to the newly created stream after setup.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -654,6 +654,9 @@ audiounit_get_input_device_id(AudioDeviceID * device_id)
   return CUBEB_OK;
 }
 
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -654,6 +654,9 @@ audiounit_get_input_device_id(AudioDeviceID * device_id)
   return CUBEB_OK;
 }
 
+static int audiounit_stream_get_volume(cubeb_stream * stm, float * volume);
+int audiounit_stream_set_volume(cubeb_stream * stm, float volume);
```
kinetiknz: audiounit_stream_set_volume should be static, please fix while you're touching this code.

achronop:
That was fast! Thanks for the review.

PR #253, opened 2017-03-07
audiounit: context synchronization in reinit stream

achronop:
Lock cubeb mutex during reinit stream. This add synchronization between the reinit and the stream_start/stop/init/destroy. I believe that causes the crash in Bug 1343930 but I did not manage to repro in order to verify.


kinetiknz:
Bouncing review to @padenot.

But can you please fix the `context_locl` typo in the other places this lock is taken?

PR #254, opened 2017-03-07
Make `stop_and_join_render_thread` more idempotent, and add a test.

padenot:
This fixes Gecko bug #1345147.

### DIFF
--- a/test/test_overload_callback.cpp
+++ b/test/test_overload_callback.cpp
@@ -1,0 +1,85 @@
+/*
+ * Copyright © 2017 Mozilla Foundation
+ *
+ * This program is made available under an ISC-style license.  See the
+ * accompanying file LICENSE for details.
+ */
+
+#include "gtest/gtest.h"
+#if !defined(_XOPEN_SOURCE)
+#define _XOPEN_SOURCE 600
+#endif
+#include <stdio.h>
+#include <stdlib.h>
+#include <math.h>
+#include <atomic>
+#include "cubeb/cubeb.h"
+#include "common.h"
+
+#define SAMPLE_FREQUENCY 48000
+#define STREAM_FORMAT CUBEB_SAMPLE_FLOAT32LE
+
+std::atomic<bool> load_callback{ false };
+
+long data_cb(cubeb_stream * stream, void * user, const void * inputbuffer, void * outputbuffer, long nframes)
+{
+  if (load_callback) {
+    printf("Sleeping...\n");
+    delay(100000);
+    printf("Sleeping done\n");
+  }
+  return nframes;
+}
+
+void state_cb(cubeb_stream * stream, void * /*user*/, cubeb_state state)
+{
+  assert(stream);
+
+  switch (state) {
+  case CUBEB_STATE_STARTED:
+    printf("stream started\n"); break;
+  case CUBEB_STATE_STOPPED:
+    printf("stream stopped\n"); break;
+  case CUBEB_STATE_DRAINED:
+    assert(false && "this test is not supposed to drain"); break;
+  case CUBEB_STATE_ERROR:
+    printf("stream error\n"); break;
+  default:
+    assert(false && "this test is not supposed to have a weird state"); break;
+  }
+}
+
+TEST(cubeb, overload_callback)
+{
+  cubeb * ctx;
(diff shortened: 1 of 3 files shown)


test/test_overload_callback.cpp
```diff
@@ -0,0 +1,98 @@
...
+    printf("Sleeping...\n");
+    delay(100000);
+    printf("Sleeping done\n");
+  }
+  return nframes;
+}
+
+void state_cb(cubeb_stream * stream, void * /*user*/, cubeb_state state)
+{
+  if (stream == NULL)
```
kinetiknz: Assert instead, passing a nullptr seems like a bug we'd want to catch.

test/test_overload_callback.cpp
```diff
@@ -0,0 +1,98 @@
...
+  case CUBEB_STATE_STARTED:
+    printf("stream started\n"); break;
+  case CUBEB_STATE_STOPPED:
+    printf("stream stopped\n"); break;
+  case CUBEB_STATE_DRAINED:
+    printf("stream drained\n"); break;
+  case CUBEB_STATE_ERROR:
+    printf("stream error\n"); break;
+  default:
+    printf("unknown stream state %d\n", state);
```
kinetiknz: Assert instead, passing an invalid state value seems like a bug.

test/test_overload_callback.cpp
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  return;
```
kinetiknz: Remove this

test/test_overload_callback.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include <stdlib.h>
+#include <math.h>
+#include <atomic>
+#include "cubeb/cubeb.h"
+#include "common.h"
+
+#define SAMPLE_FREQUENCY 48000
+#define STREAM_FORMAT CUBEB_SAMPLE_FLOAT32LE
+
+std::atomic<bool> load_callback = false;
```
kinetiknz: I think you need `std::atomic<bool> load_callback{ false };` to fix the Travis error

test/test_overload_callback.cpp
```diff
@@ -0,0 +1,98 @@
...
+  if (stream == NULL)
+    return;
+
+  switch (state) {
+  case CUBEB_STATE_STARTED:
+    printf("stream started\n"); break;
+  case CUBEB_STATE_STOPPED:
+    printf("stream stopped\n"); break;
+  case CUBEB_STATE_DRAINED:
+    printf("stream drained\n"); break;
```
kinetiknz: data_cb never triggers drain, so maybe assert if we hit this case

test/test_overload_callback.cpp
```diff
@@ -0,0 +1,98 @@
...
+    printf("unknown stream state %d\n", state);
+  }
+
+  return;
+}
+
+TEST(cubeb, overload_callback)
+{
+  cubeb *ctx;
+  cubeb_stream *stream;
```
kinetiknz: Space between * and variable name on both lines here.

test/test_overload_callback.cpp
```diff
@@ -0,0 +1,98 @@
...
+    ASSERT_EQ(r, CUBEB_OK);
+  }
+
+  output_params.format = STREAM_FORMAT;
+  output_params.rate = 48000;
+  output_params.channels = 2;
+  output_params.layout = CUBEB_LAYOUT_STEREO;
+
+  r = cubeb_get_min_latency(ctx, output_params, &latency_frames);
+
```
kinetiknz: No newline

test/test_overload_callback.cpp
```diff
@@ -0,0 +1,98 @@
...
+TEST(cubeb, overload_callback)
+{
+  cubeb *ctx;
+  cubeb_stream *stream;
+  cubeb_stream_params output_params;
+  int r;
+  uint32_t latency_frames = 0;
+
+  r = cubeb_init(&ctx, "Cubeb callback overload");
+  if (r != CUBEB_OK) {
```
kinetiknz: This can just be ASSERT_EQ(r, CUBEB_OK); without the if and fprintf, ditto all other cases

PR #258, opened 2017-03-17
Avoid generating unnecessary memory barriers in ring_buffer_base

kinetiknz:
This annotates ring_buffer_base's various loads and stores from the read and write indices with the narrowest memory ordering that particular load or store requires, rather than accepting the default  memory_order_seq_cst everywhere.  This avoids generating numerous unnecessary memory barriers (the x86 impl is now barrier free, and the ARMv7 impl has half as many barriers), which has a positive effect on performance.

Tested on RPi3 and i7-3720QM.

Note that I took the lazy path of removing the non-lock-free variants of ring_buffer_base since we don't use it (except in the test) and I don't know if there are specific plans to use it.  It's probably possible to keep it with more template magic to let us use the methods on std::atomic<T>, but the code is simpler with that aspect of genericity removed.

r? @padenot

### DIFF
--- a/src/cubeb_ringbuffer.h
+++ b/src/cubeb_ringbuffer.h
@@ -118,8 +118,8 @@ public:
     assert_correct_thread(producer_id);
 #endif
 
-    int rd_idx = read_index_;
-    int wr_idx = write_index_;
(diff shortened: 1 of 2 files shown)


padenot:
Cool, thanks !

PR #259, opened 2017-03-23
Assigning AudioUnit to nullptr before passing it to audiounit_create_unit

ChunMinChang:
AudioUnit is not initialized in audiounit_get_preferred_channel_layout since it's not assigned to nullptr before passing it to audiounit_create_unit.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1823,11 +1823,10 @@ audiounit_create_unit(AudioUnit * unit, io_side side, AudioDeviceID device)
   OSStatus rv;
   int r;
 
-  if (*unit == nullptr) {
-    int r = audiounit_new_unit_instance(unit, side, device);
-    if (r != CUBEB_OK) {
-      return r;
-    }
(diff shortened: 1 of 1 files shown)


ChunMinChang:
@achronop Is there any case that ```*unit``` in ```audiounit_create_unit(AudioUnit * unit, io_side side, AudioDeviceID device)``` is not ```nullptr```? I'd like to add ```assert(*unit == nullptr)``` in it to prevent the same problem if ```*unit``` should be always ```nullptr```.

achronop:
> s there any case that *unit in audiounit_create_unit(AudioUnit * unit, io_side side, AudioDeviceID device) is not nullptr? I'd like to add assert(*unit == nullptr) in it to prevent the same problem if *unit should be always nullptr.

You could add the assert but keep in your mind that I plan to change it. Now that we have aggregate devices we can configure just one audiounit so I plan to use that method to create if NULL or to enable the correct side of it's not. In any case you can add it for now but be aware that this might change in the future.

PR #263, opened 2017-03-28
WASAPI: make callback_input return !stm->draining

### LINKED ISSUE #262: Capture stream freeze on WASAPI when resampling
The `test_record.exe` command effectively freezes the stream after a single callback by returning false to WASAPI. There is no indication from cubeb, which doesn't pass any state updates or prematurely stop the stream.

This happens due to the fact that my system's native sample rate is 44.1KHz instead of the desired 48KHz, and the resampler doesn't guarantee that all frames it's given will be immediately passed on. When the WASAPI backend has, e.g., 448 frames, the resampler will sometimes produce 487 resampled frames and sometimes 488 resampled frames. The backend treats this as equivalent to the user callback returning less than the full amount of frames, since it doesn't match the number of frames in its linear input buffer, and returns false (telling WASAPI it doesn't want any more callbacks).

ligfx:
Fixes #262: Capture stream freezes on WASAPI when resampling.

The previous behavior would return false (and thus stop all future
callbacks) if the user didn't process all frames in linear_input_buffer.
This could erroneously happen when the resampler held back frames for
the next batch (e.g. 448 input frames at 44.1 KHz upsampled to 48 KHz,
which doesn't have a perfect integer scale), so the user callback didn't
have access to all of the current linear input frames.

The refill() function already handles the case where the user callback
returns less than the number of frames passed in (by settings
stm->draining), so this commit changes the callback_input() return value
to defer to that.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -757,7 +757,7 @@ refill_callback_duplex(cubeb_stream * stm)
 bool
 refill_callback_input(cubeb_stream * stm)
 {
-  bool rv, consumed_all_buffer;
+  bool rv;
 
   XASSERT(has_input(stm) && !has_output(stm));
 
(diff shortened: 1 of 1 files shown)


padenot:
Thanks !

PR #265, opened 2017-03-29
pulse: unref pa_operation in two places where their return was ignored.

kinetiknz:
pa_context_get_server_info and pa_context_get_sink_info_by_name both return
pa_operations, which require the caller to unref them.  In both cases we
were ignoring the return value and leaking the operations.

@djg This also fixes the missing wait that you noticed when porting to Rust.

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -142,7 +142,11 @@ sink_info_callback(pa_context * context, const pa_sink_info * info, int eol, voi
 static void
 server_info_callback(pa_context * context, const pa_server_info * info, void * u)
 {
-  WRAP(pa_context_get_sink_info_by_name)(context, info->default_sink_name, sink_info_callback, u);
+  pa_operation * o;
+  o = WRAP(pa_context_get_sink_info_by_name)(context, info->default_sink_name, sink_info_callback, u);
+  if (o) {
+    WRAP(pa_operation_unref)(o);
+  }
 }
 
 static void
@@ -576,6 +580,7 @@ pulse_init(cubeb ** context, char const * context_name)
 {
   void * libpulse = NULL;
   cubeb * ctx;
+  pa_operation * o;
 
   *context = NULL;
 
@@ -614,9 +619,17 @@ pulse_init(cubeb ** context, char const * context_name)
     return CUBEB_ERROR;
   }
 
+  /* server_info_callback performs a second async query, which is
+     responsible for initializing default_sink_info and signalling the
+     mainloop to end the wait. */
   WRAP(pa_threaded_mainloop_lock)(ctx->mainloop);
-  WRAP(pa_context_get_server_info)(ctx->context, server_info_callback, ctx);
+  o = WRAP(pa_context_get_server_info)(ctx->context, server_info_callback, ctx);
+  if (o) {
+    operation_wait(ctx, NULL, o);
+    WRAP(pa_operation_unref)(o);
+  }
   WRAP(pa_threaded_mainloop_unlock)(ctx->mainloop);
+  assert(ctx->default_sink_info);
 
   *context = ctx;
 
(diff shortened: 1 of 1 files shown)


djg:
@kinetiknz LGTM. I missed the pa_operation in the callback.

src/cubeb_pulse.c
```diff
@@ -142,7 +142,11 @@ sink_info_callback(pa_context * context, const pa_sink_info * info, int eol, voi
 static void
 server_info_callback(pa_context * context, const pa_server_info * info, void * u)
 {
-  WRAP(pa_context_get_sink_info_by_name)(context, info->default_sink_name, sink_info_callback, u);
+  pa_operation * o;
+  o = WRAP(pa_context_get_sink_info_by_name)(context, info->default_sink_name, sink_info_callback, u);
+  if (o) {
+    WRAP(pa_operation_unref)(o);
```
djg: Does this need to wait here? Otherwise, the operation_wait on the result of pa_context_get_server_info() will complete before the chain of callbacks does.

Actually, scratch that. The callbacks will occur on the same thread associated with pa_threaded_mainloop, so we can't block waiting for the callback of pa_context_get_sink_info_by_name().

Perhaps this is worthy of documenting in a comment?

kinetiknz: I added a comment in pulse_init at the "root" of the chain.

kinetiknz: I'll probably add a comment to server_info_callback as well, making the reason why it doesn't signal explicit.

src/cubeb_pulse.c
```diff
@@ -614,9 +619,17 @@ pulse_init(cubeb ** context, char const * context_name)
     return CUBEB_ERROR;
   }
 
+  /* server_info_callback performs a second async query, which is
```
kinetiknz: I might try to reword the comment, it's not clear enough as-is.

Basically we've got a chain of async callbacks: server_info_callback (to get the name) then sink_info_callback to get the info for that sink name.  We wait here for the chain to complete, and only the final callback in the chain signals.

src/cubeb_pulse.c
```diff
@@ -614,9 +619,17 @@ pulse_init(cubeb ** context, char const * context_name)
...
+     responsible for initializing default_sink_info and signalling the
+     mainloop to end the wait. */
   WRAP(pa_threaded_mainloop_lock)(ctx->mainloop);
-  WRAP(pa_context_get_server_info)(ctx->context, server_info_callback, ctx);
+  o = WRAP(pa_context_get_server_info)(ctx->context, server_info_callback, ctx);
+  if (o) {
```
achronop: Nit: The purposed pattern for waiting is:
```
    while (pa_operation_get_state(o) == PA_OPERATION_RUNNING)
        pa_threaded_mainloop_wait(m);
    pa_operation_unref(o);
```
I know this one works well in many other cases.

src/cubeb_pulse.c
```diff
@@ -142,7 +142,11 @@ sink_info_callback(pa_context * context, const pa_sink_info * info, int eol, voi
 static void
 server_info_callback(pa_context * context, const pa_server_info * info, void * u)
 {
-  WRAP(pa_context_get_sink_info_by_name)(context, info->default_sink_name, sink_info_callback, u);
+  pa_operation * o;
+  o = WRAP(pa_context_get_sink_info_by_name)(context, info->default_sink_name, sink_info_callback, u);
+  if (o) {
```
achronop: Nit: You can assert here instead of checking, it is not expected to be null.

src/cubeb_pulse.c
kinetiknz: I don't understand what you mean.  `operation_wait` is a wrapper around exactly that, plus a little error-checking.

I'll merge as-is, but happy to follow up with additional changes if necessary once you clarify.

achronop: Ignore my comment I got confused, it's fine as is.

PR #266, opened 2017-03-29
Add a deadlock test on OSX

ChunMinChang:
In CoreAudio, the data callback will holds a mutex shared with AudioUnit.
Thus, if the callback request another mutex M(resource) held by the another function, without releasing its holding one, then it will cause a deadlock when the another function, which holds the mutex M, request to use AudioUnit.

In the patch, we will reproduce the deadlock in [bug 1337805][b1337805] as follows:

```
                   holds
  data_callback <---------- mutext_AudioUnit
       |                            ^
       |                            |
       | request                    | request
       |                            |
       v           holds            |
  mutex_cubeb ------------> get_channel_layout
```

You can see more detail on [comment 0][b1350511-c0] and [1][b1350511-c1] of bug 1337805

The above deadlock problem is solve by #236. To to prevent it happens again, we add the test here in case someone without such knowledge misuses the AudioUnit in get_channel_layout. 

[b1337805]: https://bugzilla.mozilla.org/show_bug.cgi?id=1337805 "Bug 1337805"
[b1350511-c0]: https://bugzilla.mozilla.org/show_bug.cgi?id=1350511#c0 "Bug 1350511"
[b1350511-c1]: https://bugzilla.mozilla.org/show_bug.cgi?id=1350511#c1 "Bug 1350511"


test/test_deadlock.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+void * wait_to_get_layout(void *)
+{
+  while(!callbacking_before_getting_context) {
+    fprintf(stderr, "waiting for data callback ...\n");
+    usleep(1000); // Force to switch threads by sleeping 1 ms.
+  }
+
+  fprintf(stderr, "try getting channel layout ...\n");
+  get_preferred_channel_layout(); // Deadlock checkpoint.
```
achronop: Since we are doing this, if it's not too much work, can we test more getter methods from cubeb api. Right now we are testing the one that we know we have fixed. In any case we can add it with a follow up patch.

test/test_deadlock.cpp
```diff
@@ -0,0 +1,239 @@
...
+template<typename T>
+long data_cb(cubeb_stream * /*stream*/, void * /*user*/,
+             const void * /*inputbuffer*/, void * /*outputbuffer*/, long nframes)
+{
+  called = true;
+
+  if (!task_done) {
+    callbacking_before_getting_context = true;
+    fprintf(stderr, "time to switch thread!\n");
+    usleep(100000); // Force to switch threads by sleeping 100 ms.
```
achronop: 100ms in audio callback it's way too big. Anything over 10ms would create a glitch. I understand that the intention of the test is something else so we can leave it like that.

test/test_deadlock.cpp
```diff
@@ -0,0 +1,239 @@
...
+    callbacking_before_getting_context = false;
+  }
+
+  fprintf(stderr, "try getting backend id ...\n");
+
+  // Try requesting mutex of context when holding mutex of AudioUnit.
+  char const * backend_id = cubeb_get_backend_id(get_cubeb_context());
```
achronop: It would be nice to test more methods from cubeb api here.

test/test_deadlock.cpp
ChunMinChang: Thanks for the careful review! I wasn't aware of this issue.

Since 10ms is also available to switch thread from my test, I will change it to 10ms and leave some comment there to highlight this. I think it's closer to the real use case. However, if it's deadlocked, then it must be blocked longer than 10ms. The callback thread will be continued after unlocking.

test/test_deadlock.cpp
```diff
@@ -0,0 +1,239 @@
...
+// current used AudioUnit.
+template<typename T>
+long data_cb(cubeb_stream * /*stream*/, void * /*user*/,
+             const void * /*inputbuffer*/, void * /*outputbuffer*/, long nframes)
+{
+  called = true;
+
+  if (!task_done) {
+    callbacking_before_getting_context = true;
+    fprintf(stderr, "time to switch thread!\n");
```
ChunMinChang: I'll add more log for better readability

test/test_deadlock.cpp
ChunMinChang: The key is that we request cubeb's mutex by ```get_cubeb_context()```, when this thread is already holding a mutex for AudioUnit. So, we can test any API require the cubeb's mutex here.

test/test_deadlock.cpp
```diff
@@ -0,0 +1,239 @@
...
+{
+  auto_lock lock(context_mutex);
+  cubeb * context = get_cubeb_context_unlocked();
+  ASSERT_TRUE(!!context);
+
+  // We will cause a deadlock if cubeb_get_preferred_channel_layout requests
+  // mutext of AudioUnit when it holds mutex of context.
+  cubeb_channel_layout layout;
+  int r = cubeb_get_preferred_channel_layout(context, &layout);
+  ASSERT_EQ(r == CUBEB_OK, layout != CUBEB_LAYOUT_UNDEFINED);
```
ChunMinChang: The layout will be undefined if it's not SMPTE format. It doesn't affect the deadlock test, so I just check the following facts
- If layout is undefined, then the returned value must be CUBEB_ERROR
- If layout is no undefined, then the returned value must be CUBEB_OK

test/test_deadlock.cpp
```diff
@@ -0,0 +1,239 @@
...
+  pthread_t subject = *((pthread_t *) s);
+
+  sleep(2); // Sleep 2 seconds before detecting deadlock.
+
+  if (!task_done) {
+    pthread_kill(subject, CALL_THREAD_KILLER);
+    pthread_detach(subject);
+    // pthread_kill doesn't release mutexes held by the killed thread,
+    // so we need to unlock them manually.
+    context_mutex.unlock();
```
ChunMinChang: At this point, the ```context_mutex``` is held by ```wait_to_get_layout``` thread instead of ```watchdog``` thread, so we should not call ```assert_current_thread_owns()```.

The ```context_mutex.unlock()``` is used to break the deadlock between ```wait_to_get_layout``` thread and ```data_cb``` thread here.

PR #267, opened 2017-03-29
Bug 1349527 - Remove an extra +1 in cubeb_log_message::cubeb_log_message()

padenot:
Fix an out-of-bounds write by removing and extra +1 which was causing a \0
to be written out of |storage| in edge cases.

Patch by @BurningMind.


padenot:
This is obvious enough, I'll merge right now.

PR #269, opened 2017-03-30
Dont nest pulseaudio callbacks

djg:


djg:
Discussed it with @kinetiknz and this was silly.

PR #272, opened 2017-04-03
audiounit: make input_callback read correct number of frames handled

ligfx:
Same thing as #262/#263, but for the CoreAudio backend.

Might be a useful to change the definition of `cubeb_resampler_fill` at some point to make clear what the number of frames handled should be compared against.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -442,13 +442,10 @@ audiounit_input_callback(void * user_ptr,
                                         &total_input_frames,
                                         NULL,
                                         0);
+  assert(outframes >= 0);
+
   // Reset input buffer
   stm->input_linear_buffer->clear();
-
-  if (outframes < 0 || (UInt32) outframes != input_frames) {
-    stm->shutdown = true;
-    return noErr;
-  }
 
   return noErr;
 }


src/cubeb_audiounit.cpp
```diff
@@ -445,7 +445,7 @@ audiounit_input_callback(void * user_ptr,
   // Reset input buffer
   stm->input_linear_buffer->clear();
 
-  if (outframes < 0 || (UInt32) outframes != input_frames) {
+  if (outframes < 0 || (UInt32) outframes != total_input_frames) {
```
achronop: I do not think this is what we want here. On `cubeb_resampler_fill` return, value of `total_input_frames` will have been updated by the number of input frames _used_ by the resampler. This is not the number that we want to compare with the number of frames _produced_ by the resampler. 

I do not think that audiounit backend suffers by the same errors you see in WASAPI. In addition this part of the code corresponds to capture only functionality that is not used in Firefox yet. Can you please provide more context about the error you try to correct.

ligfx: The audiounit backend does suffer from the same error as #262/#263. You can see this by adding a message to the callback in `test_record`; in some situations it will only be called once.

The _used_ versus _produced_ terminology is unclear. The way I understand it is, this function passes a certain number of input frames (`input_frames`, or `total_input_frames` pre-call) to `cubeb_resampler_fill`. These frames are _all_ consumed by the resampler and added to its buffer.

It then takes some number of frames from that buffer (`total_input_frames` post-call), resamples them, and passes them to the user callback. The return value from the user callback is scaled to the original sample rate and returned by `cubeb_resampler_fill`. So, the logic here should compare the number of frames handled by the user callback against the number of frames given _to the user callback_, not the number of frames given to the resampler.

Does that sound right?

achronop: > The audiounit backend does suffer from the same error as #262/#263. You can see this by adding a message to the callback in test_record; in some situations it will only be called once.

That's just about right. We do not use capture only scenario in Firefox and some errors have not found yet. The confusion for me was that duplex resampling requests a specific number of output frames. On the other hand non-duplex resampling converts all provided frames. 

After discussing about it with Paul, the best solution for now is to stop draining at input only scenarios. You can replace the if clause with an assert to verify that returned frames are zero or more. 

If we want to restore drain functionality for input only scenario in the future we need to update resampler to return a specific number of frames similar to duplex case.

ligfx: Okay, that makes sense. Would you like me to submit a similar patch against WASAPI as well (replacing if-clause with assert)?

achronop: > Would you like me to submit a similar patch against WASAPI as well (replacing if-clause with assert)?

I do not think is necessary, wasapi has already an assert (cubeb_wasapi.cpp#L782) and returns `!draining` which is always true on capture.

PR #273, opened 2017-04-05
Using RAII helpers for cubeb context and stream in the test

ChunMinChang:
Apply RAII helpers for cubeb context and stream in most test cases but leaving test_sanity for explicit flow control.

### DIFF
--- a/test/common.h
+++ b/test/common.h
@@ -107,4 +107,18 @@ void print_log(const char * msg, ...)
   va_end(args);
 }
 
+struct cubeb_cleaner
+{
+  cubeb_cleaner(cubeb * context) : ctx(context) {}
+  ~cubeb_cleaner() { cubeb_destroy(ctx); }
+  cubeb * ctx;
+};
+
+struct cubeb_stream_cleaner
+{
+  cubeb_stream_cleaner(cubeb_stream * stream) : stm(stream) {}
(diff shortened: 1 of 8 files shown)


test/common.h
```diff
@@ -107,4 +107,18 @@ void print_log(const char * msg, ...)
   va_end(args);
 }
 
+struct cubeb_cleaner
```
kinetiknz: These could be replaced with a unique_ptr instead.  cubeb_audiounit has an example.

test/test_devices.cpp
```diff
@@ -112,22 +113,23 @@ TEST(cubeb, enumerate_devices)
   r = cubeb_init(&ctx, "Cubeb audio test", NULL);
   if (r != CUBEB_OK) {
     fprintf(stderr, "Error initializing cubeb library\n");
-    ASSERT_EQ(r, CUBEB_OK);
+    return r;
```
kinetiknz: It'd be better to assert everywhere rather than returning a result and asserting in the caller, since you lose the location of the failure this way.

kinetiknz:
Thanks! I think the Windows build wants #include <memory> to find unique_ptr.

ChunMinChang:
Thanks for the reminder and review! I've changed the files by your advice.

PR #274, opened 2017-04-06
Using gtest style's error message in cubeb

ChunMinChang:
- Perfer ```fprintf(stderr, ...)``` instead of ```printf(...)```
- Turn the error message into gtest's style

We have many patterns like the following:
```
if (r != CUBEB_OK) {
    fprintf(stderr, "Error initializing cubeb stream\n");
    ASSERT_EQ(r, CUBEB_OK);
}
```
We should replace them by:
```
ASSERT_EQ(r, CUBEB_OK) << "Error initializing cubeb stream";
```

### DIFF
--- a/test/test_tone.cpp
+++ b/test/test_tone.cpp
@@ -82,13 +82,13 @@ void state_cb_tone(cubeb_stream *stream, void *user, cubeb_state state)
 
   switch (state) {
   case CUBEB_STATE_STARTED:
-    printf("stream started\n"); break;
+    fprintf(stderr, "stream started\n"); break;
   case CUBEB_STATE_STOPPED:
-    printf("stream stopped\n"); break;
+    fprintf(stderr, "stream stopped\n"); break;
   case CUBEB_STATE_DRAINED:
-    printf("stream drained\n"); break;
+    fprintf(stderr, "stream drained\n"); break;
   default:
-    printf("unknown stream state %d\n", state);
+    fprintf(stderr, "unknown stream state %d\n", state);
   }
 
   return;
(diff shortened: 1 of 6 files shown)


kinetiknz:
This seems like a good change, thanks!

For reasons not immediately obvious to me, scan-build on Travis is complaining about one of the changes:

    /home/travis/build/kinetiknz/cubeb/test/test_tone.cpp:117:3: warning: Potential leak of memory pointed to by 'user_data'
      ASSERT_NE(user_data, nullptr) << "Error allocating user data";

kinetiknz:
> For reasons not immediately obvious to me, scan-build on Travis is complaining about one of the changes

I guess this is because the failure side of the assert returns from the test function and the analyzer can't see into (or doesn't understand) the guts of the success side (::testing::internal::CmpHelperNE) enough to realize that it should boil down to something equivalent to `!=` meaning the failure path is never taken if user_data is non-null.

Anyway, the simple solution is to change the allocation of user_data to a unique_ptr, which makes the code slightly nicer anyway.

ChunMinChang:
Thanks for the help! I've updated the code as per your suggestion

test/test_tone.cpp
```diff
@@ -113,13 +112,14 @@ TEST(cubeb, tone)
   params.channels = 1;
   params.layout = CUBEB_LAYOUT_MONO;
 
-  user_data = (struct cb_user_data *) malloc(sizeof(*user_data));
-  ASSERT_NE(user_data, nullptr) << "Error allocating user data";
+  std::unique_ptr<struct cb_user_data, decltype(&free)>
+    user_data((struct cb_user_data *)calloc(1, sizeof(struct cb_user_data)), free);
```
kinetiknz: FWIW, you could just use `new` and a plain `unique_ptr` instead.

kinetiknz:
Thanks!

PR #275, opened 2017-04-07
some cleanup

ChunMinChang:
- Fix following issue with #274
- Use auto-for-loop in ```text_mixer.cpp```
- Rename mixer stuff in ```cubeb_audiounit.cpp```


kinetiknz:
Thanks!

PR #276, opened 2017-04-07
Get channel mapping when AudioChannelLayout's tag is not UseChannelDescriptions

ChunMinChang:
For now, we can only read the channel layout when AudioChannelLayout's tag is ```kAudioChannelLayoutTag_UseChannelDescriptions```. We should get the layout mapping no matter what tag value is.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ if(USE_AUDIOUNIT)
     src/cubeb_audiounit.cpp
     src/cubeb_osx_run_loop.cpp)
   target_compile_definitions(cubeb PRIVATE USE_AUDIOUNIT)
-  target_link_libraries(cubeb PRIVATE "-framework AudioUnit" "-framework CoreAudio" "-framework CoreServices")
+  target_link_libraries(cubeb PRIVATE "-framework AudioToolbox" "-framework AudioUnit" "-framework CoreAudio" "-framework CoreServices")
 endif()
 
 check_include_files(pulse/pulseaudio.h USE_PULSE)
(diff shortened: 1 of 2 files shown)


ChunMinChang:
The code now is a draft. I need to figure out when ```mChannelLayoutTag``` is not ```kAudioChannelLayoutTag_UseChannelDescriptions``` if the ```AudioChannelLayout``` is returned from ```AudioUnitGetProperty``` with ```kAudioUnitProperty_AudioChannelLayout```.

I already sent an email on coreaudio-api@lists.apple.com to ask for the answer.

ChunMinChang:
Still waiting for the answer from coreaudio-api mailing list. I also asked on [stackoverflow](http://stackoverflow.com/questions/43734547/when-is-mchannellayouttag-not-kaudiochannellayouttag-usechanneldescriptions).

One way to know it is to add an assertion like ```assert(layout->mChannelLayoutTag == kAudioChannelLayoutTag_UseChannelDescriptions)``` and see if any firefox user hits it.

Another way is to add our convertion code to gecko(reference 2 below) to see if it works.

- References:
  1. https://chromium.googlesource.com/chromium/src/+/5285d00f561827da882c8faaa7bda1d571c4fc60/media/audio/mac/audio_manager_mac.cc#395
  2. http://searchfox.org/mozilla-central/rev/0079c7adf3b329bff579d3bbe6ac7ba2f6218a19/dom/media/platforms/apple/AppleATDecoder.cpp#481

kinetiknz:
Is this PR still relevant after https://github.com/kinetiknz/cubeb/pull/426?

ChunMinChang:
I think the changes are outdated. It might still be a problem. I saw chromium has [some similar code dealing with this situation](https://cs.chromium.org/chromium/src/media/audio/mac/audio_manager_mac.cc?l=391&rcl=60b409bdc94f4cfe3562c47e70a269e3be0bfb6d). However, I don't have a way to get a `AudioChannelLayoutTag` value that is not `kAudioChannelLayoutTag_UseChannelDescriptions`. We can close it until we firgure out how to hit this.

PR #277, opened 2017-04-11
wasapi: Implement S16NE-format streams

ligfx:
Brings the WASAPI backend up to parity with the other backends, making it easier to write cross-platform code for S16 audio. I tried to base my approach off of the AudioUnit backend (in fact, `auto_array_wrapper` is now in cubeb_utils and used by both) and a little bit from the Jack backend (the `bytes_per_sample` member).

The official documentation is (purposefully?) obtuse as to when S16 is a supported format, but it seems to hint that it's always available in non-exclusive  mode, and the rest of the internet seems to believe that that's true.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -310,6 +310,11 @@ struct cubeb_stream {
   /* WASAPI input works in "packets". We re-linearize the audio packets
    * into this buffer before handing it to the resampler. */
   std::unique_ptr<auto_array_wrapper> linear_input_buffer;
+  /* Bytes per sample. This multiplied by the number of channels is the number
+   * of bytes per frame. */
+  size_t bytes_per_sample = 0;
+  /* WAVEFORMATEXTENSIBLE sub-format: either PCM or float. */
+  GUID waveformatextensible_sub_format = GUID_NULL;
   /* Stream volume.  Set via stream_set_volume and used to reset volume on
      device changes. */
   float volume = 1.0;
@@ -548,8 +553,7 @@ frames_to_bytes_before_mix(cubeb_stream * stm, size_t frames)
 {
   // This is called only when we has a output client.
   XASSERT(has_output(stm));
-  size_t stream_frame_size = stm->output_stream_params.channels * sizeof(float);
-  return stream_frame_size * frames;
+  return stm->output_stream_params.channels * stm->bytes_per_sample * frames;
 }
 
 /* This function handles the processing of the input and output audio,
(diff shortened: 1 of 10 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -310,6 +310,9 @@ struct cubeb_stream {
   /* WASAPI input works in "packets". We re-linearize the audio packets
    * into this buffer before handing it to the resampler. */
   std::unique_ptr<auto_array_wrapper> linear_input_buffer;
+  /* Bytes per sample. This multiplied by the number of channels is the number
+   * of bytes per frame. */
+  size_t bytes_per_sample = 0;
```
kinetiknz: Looks like this was included in the wrong commit, should be in f5505f0.

src/cubeb_wasapi.cpp
```diff
@@ -160,6 +160,45 @@ struct auto_com {
...
+};
+
+template <typename T>
+struct mixing_impl : public mixing_wrapper {
+
+  typedef std::function<void(T * const, long, T *,
+                             unsigned int, unsigned int,
+                             cubeb_channel_layout, cubeb_channel_layout)> downmix_func;
+  typedef std::function<void(T * const, long, T *,
+                             unsigned int, unsigned int)> upmix_func;
```
kinetiknz: Please make these bare function pointers.  We had to do that in https://github.com/kinetiknz/cubeb/pull/271 to satisfy a restriction in Firefox downstream.

src/cubeb_wasapi.cpp
```diff
@@ -160,6 +160,45 @@ struct auto_com {
...
+    upmix_wrapper = umfunc;
+  }
+
+  ~mixing_impl() {}
+
+  void mix(void * const input_buffer, long input_frames, void * output_buffer,
+               cubeb_stream_params const * stream_params,
+               cubeb_stream_params const * mixer_params) override {
+    T* const in = static_cast<T *>(input_buffer);
+    T* out = static_cast<T *>(output_buffer);
```
kinetiknz: Space between `T` and `*` per the coding style

src/cubeb_wasapi.cpp
ligfx: Sure thing. Done.

src/cubeb_wasapi.cpp
ligfx: Done.

src/cubeb_wasapi.cpp
ligfx: Wow, good eyes! Moved to the correct commit.

kinetiknz:
Thanks!

PR #280, opened 2017-04-12
audiounit: Add unique id in aggregate device name and UID.

achronop:
Add a unique id for the device name and UID of the newly created aggregate device. This is necessary when more than one aggregate device functioning in the same time. If the name and UID are the same the second aggregate device fails to be configured. For unique id used the pointer address of the plug in id. BMO #1355306.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1494,11 +1494,15 @@ audiounit_create_blank_aggregate_device(AudioObjectID * plugin_id, AudioDeviceID
                                                                            &kCFTypeDictionaryKeyCallBacks,
                                                                            &kCFTypeDictionaryValueCallBacks);
 
-  CFStringRef aggregate_device_name = CFSTR("CubebAggregateDevice");
+  CFStringRef aggregate_device_name = CFStringCreateWithFormat(NULL, NULL, CFSTR("CubebAggregateDevice_%llx"),
+                                                               reinterpret_cast<uint64_t>(plugin_id));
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -1494,11 +1494,15 @@ audiounit_create_blank_aggregate_device(AudioObjectID * plugin_id, AudioDeviceID
                                                                            &kCFTypeDictionaryKeyCallBacks,
                                                                            &kCFTypeDictionaryValueCallBacks);
 
-  CFStringRef aggregate_device_name = CFSTR("CubebAggregateDevice");
+  CFStringRef aggregate_device_name = CFStringCreateWithFormat(NULL, NULL, CFSTR("CubebAggregateDevice_%llx"),
```
padenot: Maybe I'm being over-paranoid, but it might be best to not expose any info about the address space of the process. Can we use a global integer that is simply incremented instead of an address ?

Same for the other string.

achronop:
They are private AD so they do not appear in the lists but I'm sure there will be a way to find those names. I replaced the pointer id with timestamp.

src/cubeb_audiounit.cpp
achronop: Global/static variables does not work in Firefox. They start every time from zero. I am not sure why yet. Global/static variables do work in a stand alone binary statically linked to libcubeb.

padenot: Can you not put it in the `cubeb` struct ?

achronop: I could but it does not work with e10s. In that case the cubeb struct is new in every gum call.

PR #281, opened 2017-04-14
cubeb_sndio: Fix leak in sndio_stream_init() failure case

lioncash:
Fairly unlikely to happen, but may as well cover all bases anyways.


kinetiknz:
Thanks!

PR #282, opened 2017-04-14
cubeb_wasapi: Correct duplicate bitwise OR entry in wasapi_enumerate_devices()

lioncash:
Corrects an accidental duplicate, which would prevent enumerating both rendering and capturing endpoints.


kinetiknz:
Thanks!

PR #287, opened 2017-04-20
Enable multichannel recording if it's hardware-supported

ChunMinChang:
The ```mix_format``` returned from ```IAudioClient::GetMixFormat``` is the hardware-supported format for the stream, so we should not crash the stream on purpose if the user's audio sound card supports N-channel recording, where N > 2.

If we intend to crash a N-channel recording stream, we should check ```stream_params->channels``` instead of ```mix_format->nChannels```.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -1556,11 +1556,6 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
    * use the default setting retrieved from the stream format of the audio
    * engine's internal processing by GetMixFormat. */
   if (mix_format->nChannels > 2) {
-    /* Currently, we only support mono and stereo for capture stream. */
-    if (direction == eCapture) {
-      XASSERT(false && "Multichannel recording is not supported.");
-    }
-
     handle_channel_layout(stm, mix_format, stream_params);
   }
 


padenot:
> If we intend to crash a N-channel recording stream, we should check stream_params->channels instead of mix_format->nChannels.

We never want to crash on normal error paths.

ChunMinChang:
I mean, it's better to use 
```
if (stream_params->channels > 2 && direction == eCapture) {
   XASSERT(false && "Multichannel recording is not supported.");
}
```
to fail the assertion(if we intend to do so).

We should not fail the assertion only because user has a multi-channel recording device.

padenot:
What I mean is that release assertion are only to be reached in situation where we can't do anything else. This is very very easy to trigger, so we should fall back to something that works: either down/up/mix ourselves or fail to open the stream.

ChunMinChang:
With the patch, we don't prevent user to record audio when they have multi-channel device. Actually, when they are using Firefox, they still can't record more than 2 channels, even they have multi-channel device. The reason is that there is no case for setting ```stream_params->channels``` more than 2, even the ```mix_format->nChannels```(channel counts of hardware support) is more than 2.

In such case(with the patch), the channel mapping(```dwChannelMask``` and ```nChannels```) of ```mix_format``` wil be set in ```handle_channel_layout``` by our defined value(```stream_params->layout``` and ```stream_params->channels```, these two value should be valid). Then, this format will be checked via ```IAudioClient::IsFormatSupported```. If it's ok, then there is nothing to worry. If it's not, we will get a closest format and the channel mapping of ```mix_format``` will be set to the same as the closest format's one.

After out of the ```handle_channel_layout```, this final channel mapping will be saved in ```mix_params```. Here we use ```dwChannelMask``` to map to our defined layout instead of counting channels and its labels, so there is no memory-abuse issue addressed in pull [288](https://github.com/kinetiknz/cubeb/pull/288/commits/6494c91a1308a75b13391f48205c67d963cad9cb).

Finally, the ```mix_params``` is used for downmix or upmixing. If ```mix_params->channels > stream_params->channels```, then we will do upmixing. Otherwise, we will do downmixing.

I believe there is only few people have multi-channel recording device. I try my 8-track recording device, however, it's configured to 4 input devices with stereo layout on my PC. I guess other N-track recording device will also be configured into several 2 channel devices on Windows.

I don't have a multi-channel recording device and I can't find a soundflower alternative for Windows now, so the above is only my inference.

padenot:
> With the patch, we don't prevent user to record audio when they have multi-channel device. Actually, when they are using Firefox, they still can't record more than 2 channels, even they have multi-channel device. The reason is that there is no case for setting stream_params->channels more than 2, even the mix_format->nChannels(channel counts of hardware support) is more than 2.

@achronop is changing that right now, the patch for Firefox is almost ready.

> I believe there is only few people have multi-channel recording device. I try my 8-track recording device, however, it's configured to 4 input devices with stereo layout on my PC. I guess other N-track recording device will also be configured into several 2 channel devices on Windows.

We have 80 crashes on this assertion, this is not a theoretical concern.

> I don't have a multi-channel recording device and I can't find a soundflower alternative for Windows now, so the above is only my inference.

<http://vb-audio.pagesperso-orange.fr/Cable/> is what people use, usually.

ChunMinChang:
> We have 80 crashes on this assertion, this is not a theoretical concern.

Yes, so that's why I'd like to remove this assertion. Without the assertion, people is allowed do multi-channel recording with N-channels input device(N > 2). If we don't remove the assertion, people even cannot open a 1-channel recording stream if there is a connected N-channels input device.

> http://vb-audio.pagesperso-orange.fr/Cable/ is what people use, usually.

I tried this on my PC, but it only allows me to record sounds with stereo channels, so I can't hit the assertion below:
```c++
if (mix_format->nChannels > 2) {
  /* Currently, we only support mono and stereo for capture stream. */
  if (direction == eCapture) {
    XASSERT(false && "Multichannel recording is not supported.");
  }
  ...
}
```

padenot:
Let's merge this, we'll fix things later if needed.

PR #288, opened 2017-04-20
Make cubeb work with high channel count devices

padenot:
Without this, trying to play audio on a device that has more than 8 output channels trashes some memory and crashes.

With this patch we can use [SoundFlower](https://github.com/mattingalls/Soundflower/releases) to test multi-channel support, and to do nice things like, generating audio in a fake input device, recording the output device using another program test, etc. 

@ChunMinChang, want to have a look ? I plan to do the same for Windows after this, reading the code, it must have the same issue. I will have my windows machine next week, though.

### DIFF
--- a/src/cubeb_mixer.cpp
+++ b/src/cubeb_mixer.cpp
@@ -31,7 +31,8 @@ cubeb_channel_layout cubeb_channel_map_to_layout(cubeb_channel_map const * chann
 {
   uint32_t channel_mask = 0;
   for (uint8_t i = 0 ; i < channel_map->channels ; ++i) {
-    if (channel_map->map[i] == CHANNEL_INVALID) {
+    if (channel_map->map[i] == CHANNEL_INVALID ||
+        channel_map->map[i] == CHANNEL_UNMAPPED) {
       return CUBEB_LAYOUT_UNDEFINED;
     }
     channel_mask |= 1 << channel_map->map[i];
--- a/src/cubeb_mixer.h
+++ b/src/cubeb_mixer.h
@@ -27,7 +27,8 @@ typedef enum {
   CHANNEL_RCENTER,
   CHANNEL_RRS,
   CHANNEL_LFE,
-  CHANNEL_MAX // Max number of supported channels.
+  CHANNEL_UNMAPPED,
+  CHANNEL_MAX = 256 // Max number of supported channels.
 } cubeb_channel;
 
 static cubeb_channel const CHANNEL_INDEX_TO_ORDER[CUBEB_LAYOUT_MAX][CHANNEL_MAX] = {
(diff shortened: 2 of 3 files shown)


src/cubeb_mixer.h
```diff
@@ -27,7 +27,8 @@ typedef enum {
   CHANNEL_RCENTER,
   CHANNEL_RRS,
   CHANNEL_LFE,
-  CHANNEL_MAX // Max number of supported channels.
+  CHANNEL_UNMAPPED,
+  CHANNEL_MAX = 256 // Max number of supported channels.
```
achronop: I have a little concern about `cubeb_channel_map_to_layout()` with that change. The `channel_mask` variable in cubeb_mixer.cpp:37 will become `channel_mask |= 1 << CHANNEL_UNMAPPED` for every unknown channel and this will result in returning `CUBEB_LAYOUT_UNDEFINED` from the method. Is that intentional? We may not want to update `channel_mask` for `CHANNEL_UNMAPPED` values.

ChunMinChang: If it's intentional to map the ```cubeb_channel_map``` containing ```CHANNEL_UNMAPPED``` to ```CUBEB_LAYOUT_UNDEFINED```. I think it's better to use

```c++
if (channel_map->map[i] == CHANNEL_INVALID ||
    channel_map->map[i] == CHANNEL_UNMAPPED) {
  return CUBEB_LAYOUT_UNDEFINED;
}
```
before ```channel_mask |= 1 << channel_map->map[i];```

ChunMinChang:
I am curious about what's the difference between ```CHANNEL_UNMAPPED``` and ```CHANNEL_INVALID```. If ```CHANNEL_UNMAPPED``` is replaced by ```CHANNEL_INVALID``` in this patch, the behavior will be same. They both are unsupportrd channels. Do they have different meaning?

padenot:
> I am curious about what's the difference between CHANNEL_UNMAPPED and CHANNEL_INVALID. If CHANNEL_UNMAPPED is replaced by CHANNEL_INVALID in this patch, the behavior will be same. They both are unsupportrd channels. Do they have different meaning?

Yeah I don't know, but what I was thinking:
- `CHANNEL_INVALID` semantically refers this is an invalid channel
- `CHANNEL_UNMAPPED` semantically refers to a channel that is not mapped to a particular physical speaker position, but that is valid (for example, it can be mapped to a speaker in a speaker array, with just an positional index and no meaning, "speaker 4" instead of "right surround").

Of course, it does not change anything to use `INVALID` or `UNMAPPED`, because we don't, for now, have a way or dealing with unmapped speaker set (like SoundFlower's 64 channels).

src/cubeb_mixer.h
padenot: Yeah you're right, I'll do that.

src/cubeb_mixer.cpp
```diff
@@ -31,7 +31,8 @@ cubeb_channel_layout cubeb_channel_map_to_layout(cubeb_channel_map const * chann
 {
   uint32_t channel_mask = 0;
   for (uint8_t i = 0 ; i < channel_map->channels ; ++i) {
-    if (channel_map->map[i] == CHANNEL_INVALID) {
+    if (channel_map->map[i] == CHANNEL_INVALID ||
+        channel_map->map[i] == CHANNEL_UNMAPPED) {
```
achronop: I am not convinced this is want we want here. My understanding is that a valid layout can contain extra unmapped channels and we try to make cubeb not to crash because of that. If that's the case I would expect that we ignore the unmapped channels rather than return undefined, which is more like an error fall out. In the code above I would expect something like: 
`else if (channel_map->map[i] == CHANNEL_UNMAPPED) {continue;}`
Let me know if I get it wrong.

ChunMinChang: The mixer now assumes its channel count is a [fixed number](https://github.com/kinetiknz/cubeb/blob/2bb3e25537042ad113f02f36bd1080dc8abeb3b4/src/cubeb_mixer.cpp#L62). If the layout has a dynamic channel count, we may need to update our [mixing policy](https://github.com/kinetiknz/cubeb/blob/2bb3e25537042ad113f02f36bd1080dc8abeb3b4/src/cubeb_mixer.cpp#L357).

padenot: No it's still static.

achronop: Issue clarified in IRC. It's good to go.

PR #290, opened 2017-04-25
Use same mixer interface for wasapi and audiounit

ChunMinChang:
Since we have similar patterns to call up/downmixing in ```cubeb_wasapi.cpp``` and ```cubeb_audiounit.cpp```, it will be easier to maintain the code if we have an identical interface and use it in both backends.

### DIFF
--- a/src/cubeb_mixer.cpp
+++ b/src/cubeb_mixer.cpp
@@ -468,7 +468,6 @@ cubeb_should_mix(cubeb_stream_params const * stream, cubeb_stream_params const *
   return cubeb_should_upmix(stream, mixer) || cubeb_should_downmix(stream, mixer);
 }
 
-// An identical mixer interface beyond every backend.
 struct cubeb_mixer {
   virtual void mix(void * const input_buffer, long frames, void * output_buffer,
                    cubeb_stream_params const * stream_params,
@@ -504,11 +503,11 @@ struct cubeb_mixer_impl : public cubeb_mixer {
   cubeb_mixer_direction const direction;
 };
 
-cubeb_mixer * cubeb_mixer_create(cubeb_stream_params const * stream_params,
+cubeb_mixer * cubeb_mixer_create(cubeb_sample_format format,
                                  cubeb_mixer_direction direction)
 {
-  assert(stream_params);
-  switch(stream_params->format) {
+  assert(direction | CUBEB_MIXER_DIRECTION_DOWNMIX | CUBEB_MIXER_DIRECTION_UPMIX);
+  switch(format) {
     case CUBEB_SAMPLE_S16NE:
       return new cubeb_mixer_impl<short>(direction);
     case CUBEB_SAMPLE_FLOAT32NE:
--- a/src/cubeb_mixer.cpp
+++ b/src/cubeb_mixer.cpp
@@ -477,7 +477,7 @@ struct cubeb_mixer {
 
 template<typename T>
 struct cubeb_mixer_impl : public cubeb_mixer {
-  explicit cubeb_mixer_impl(cubeb_mixer_direction d)
+  explicit cubeb_mixer_impl(unsigned int d)
     : direction(d)
   {
   }
@@ -500,11 +500,11 @@ struct cubeb_mixer_impl : public cubeb_mixer {
 
   ~cubeb_mixer_impl() {};
 
-  cubeb_mixer_direction const direction;
+  unsigned char const direction;
 };
 
 cubeb_mixer * cubeb_mixer_create(cubeb_sample_format format,
-                                 cubeb_mixer_direction direction)
+                                 unsigned char direction)
 {
   assert(direction | CUBEB_MIXER_DIRECTION_DOWNMIX | CUBEB_MIXER_DIRECTION_UPMIX);
   switch(format) {
--- a/src/cubeb_mixer.cpp
+++ b/src/cubeb_mixer.cpp
@@ -359,10 +359,10 @@ cubeb_downmix(T const * const in, long inframes, T * out,
               cubeb_stream_params const * stream_params,
               cubeb_stream_params const * mixer_params)
 {
-  assert(in && out && inframes &&
-         stream_params->channels >= mixer_params->channels &&
-         mixer_params->channels > 0 &&
-         stream_params->layout != CUBEB_LAYOUT_UNDEFINED);
+  assert(in && out && inframes);
+  assert(stream_params->channels >= mixer_params->channels &&
+         mixer_params->channels > 0);
+  assert(stream_params->layout != CUBEB_LAYOUT_UNDEFINED);
 
   unsigned int in_channels = stream_params->channels;
   cubeb_channel_layout in_layout = stream_params->layout;
--- a/src/cubeb_mixer.cpp
+++ b/src/cubeb_mixer.cpp
@@ -506,7 +506,8 @@ struct cubeb_mixer_impl : public cubeb_mixer {
 cubeb_mixer * cubeb_mixer_create(cubeb_sample_format format,
                                  unsigned char direction)
 {
-  assert(direction | CUBEB_MIXER_DIRECTION_DOWNMIX | CUBEB_MIXER_DIRECTION_UPMIX);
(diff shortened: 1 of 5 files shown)


padenot:
Can you fix the issue on Linux ? The linux backend is still C so you can't use C++ constructs there.

padenot:
I had the same issue for the resampler, I ended up keeping the C API, but doing static dispatch with templates underneath to have static polymorphism, look in `cubeb_resampler.h` and `cubeb_resampler{,_internal}.cpp`.

ChunMinChang:
Thanks for the hints!

src/cubeb_mixer.cpp
```diff
@@ -356,10 +356,19 @@ downmix_fallback(T const * const in, unsigned long inframes, T * out, unsigned i
 template<typename T>
 void
 cubeb_downmix(T const * const in, long inframes, T * out,
-              unsigned int in_channels, unsigned int out_channels,
-              cubeb_channel_layout in_layout, cubeb_channel_layout out_layout)
+              cubeb_stream_params const * stream_params,
+              cubeb_stream_params const * mixer_params)
 {
-  assert(in_channels >= out_channels && in_layout != CUBEB_LAYOUT_UNDEFINED);
+  assert(in && out && inframes &&
```
kinetiknz: For future reference, I feel like it's better to break huge asserts up into several smaller ones.  It's easier to reason about what each one is testing, plus it's more obvious (in a crash backtrace) which condition triggered the failure.

src/cubeb_mixer.cpp
```diff
@@ -453,33 +468,67 @@ cubeb_should_mix(cubeb_stream_params const * stream, cubeb_stream_params const *
...
 }
 
-void
-cubeb_downmix_float(float * const in, long inframes, float * out,
-                    unsigned int in_channels, unsigned int out_channels,
-                    cubeb_channel_layout in_layout, cubeb_channel_layout out_layout)
-{
-  cubeb_downmix(in, inframes, out, in_channels, out_channels, in_layout, out_layout);
-}
+// An identical mixer interface beyond every backend.
```
kinetiknz: Can you please reword this? It's a bit hard to parse as-is.

src/cubeb_mixer.cpp
```diff
@@ -453,33 +468,67 @@ cubeb_should_mix(cubeb_stream_params const * stream, cubeb_stream_params const *
...
+  virtual ~cubeb_mixer() {};
+};
 
-void
-cubeb_downmix_short(short * const in, long inframes, short * out,
-                    unsigned int in_channels, unsigned int out_channels,
-                    cubeb_channel_layout in_layout, cubeb_channel_layout out_layout)
+template<typename T>
+struct cubeb_mixer_impl : public cubeb_mixer {
+  cubeb_mixer_impl(cubeb_mixer_direction d)
```
kinetiknz: Needs to be marked `explicit`

src/cubeb_mixer.cpp
```diff
@@ -453,33 +468,67 @@ cubeb_should_mix(cubeb_stream_params const * stream, cubeb_stream_params const *
...
+    : direction(d)
+  {
+  }
+
+  void mix(void * const input_buffer, long frames, void * output_buffer,
+           cubeb_stream_params const * stream_params,
+           cubeb_stream_params const * mixer_params)
+  {
+    T * in = static_cast<T*>(input_buffer);
```
kinetiknz: Casting const away from `input_buffer` seems like a bad thing.  If input_buffer can ever be modified, the parameter shouldn't be const, otherwise don't cast it away here.

src/cubeb_mixer.h
```diff
@@ -68,21 +68,20 @@ bool cubeb_should_downmix(cubeb_stream_params const * stream, cubeb_stream_param
...
-                       unsigned int in_channels, unsigned int out_channels);
-
-void cubeb_upmix_short(short * const in, long inframes, short * out,
-                       unsigned int in_channels, unsigned int out_channels);
-
-
+typedef enum {
+  CUBEB_MIXER_DIRECTION_DOWNMIX = 0x01,
+  CUBEB_MIXER_DIRECTION_UPMIX   = 0x02,
+  CUBEB_MIXER_DIRECTION_ALL     = 0x03,
```
kinetiknz: What does DIRECTION_ALL actually mean?

src/cubeb_mixer.cpp
```diff
@@ -453,33 +468,67 @@ cubeb_should_mix(cubeb_stream_params const * stream, cubeb_stream_params const *
...
+    T * out = static_cast<T*>(output_buffer);
+
+    if ((direction & CUBEB_MIXER_DIRECTION_DOWNMIX) &&
+        cubeb_should_downmix(stream_params, mixer_params)) {
+      cubeb_downmix(in, frames, out, stream_params, mixer_params);
+    } else if ((direction & CUBEB_MIXER_DIRECTION_UPMIX) &&
```
kinetiknz: Along with my question about DIRECTION_ALL below, it seems like this only handles UP/DOWN, and doesn't need to treat the enum values as flags.

src/cubeb_mixer.cpp
```diff
@@ -453,33 +468,67 @@ cubeb_should_mix(cubeb_stream_params const * stream, cubeb_stream_params const *
...
+      cubeb_upmix(in, frames, out, stream_params, mixer_params);
+    }
+  }
+
+  ~cubeb_mixer_impl() {};
+
+  cubeb_mixer_direction const direction;
+};
+
+cubeb_mixer * cubeb_mixer_create(cubeb_stream_params const * stream_params,
```
kinetiknz: Maybe it'd be better if this took the cubeb_sample_format directly, since it doesn't need anything else from cubeb_stream_params.

src/cubeb_mixer.h
ChunMinChang: ```0x01(binary: 001)``` is used to enable downmixing 
```0x02(binary: 010)``` is used to enable upmixing. 
```0x03(binary: 011 = 001 | 010)``` is used to enable downmixing and upmixing.

If it's unclear for readability, we could remove ```CUBEB_MIXER_DIRECTION_ALL``` and leave this setting for the user. That is, user should use ```CUBEB_MIXER_DIRECTION_DOWNMIX | CUBEB_MIXER_DIRECTION_UPMIX``` instead of ```CUBEB_MIXER_DIRECTION_ALL```

File: src/cubeb_mixer.cpp (hunk 7 above)
Comment (ChunMinChang) [inline src/cubeb_mixer.cpp]: You're right. I'll change it.

File: src/cubeb_mixer.h (hunk 5 above)
Comment (kinetiknz) [inline src/cubeb_mixer.h]: Right, but what is the expect result from requesting upmixing and downmixing at the same time?

File: src/cubeb_mixer.cpp (hunk 6 above)
Comment (ChunMinChang) [inline src/cubeb_mixer.cpp]: The flags are used to enable upmixing and downmixing. We need to do up/downmixing on Windows but we only do donwmixing on OSX, so that's why we use a flag to set what we need.

File: src/cubeb_mixer.cpp [hunk 8]
```diff
@@ -453,33 +468,66 @@ cubeb_should_mix(cubeb_stream_params const * stream, cubeb_stream_params const *
...
+  ~cubeb_mixer_impl() {};
+
+  cubeb_mixer_direction const direction;
+};
+
+cubeb_mixer * cubeb_mixer_create(cubeb_sample_format format,
+                                 cubeb_mixer_direction direction)
 {
-  cubeb_downmix(in, inframes, out, in_channels, out_channels, in_layout, out_layout);
+  assert(direction | CUBEB_MIXER_DIRECTION_DOWNMIX | CUBEB_MIXER_DIRECTION_UPMIX);
```
kinetiknz: This will always be true, probably not what you intended?

ChunMinChang: Thanks for the correction!

src/cubeb_mixer.cpp
kinetiknz: Thanks, makes sense to me now.

PR #296, opened 2017-05-05
Add debugging info for setting channel layout

ChunMinChang:
To make sure we don't use the memory beyond our allocation, it's better to check the index of ```mChannelDescriptions``` and ```mChannelLayoutTag``` are initialized before assigning data.

Also, we need to check cubeb can work with 8 channels input.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1278,6 +1278,7 @@ audiounit_set_channel_layout(AudioUnit unit,
   OSStatus r;
   size_t size = sizeof(AudioChannelLayout);
   auto layout = make_sized_audio_channel_layout(size);
+  assert(!layout->mChannelLayoutTag);
 
   switch (stream_params->layout) {
     case CUBEB_LAYOUT_DUAL_MONO:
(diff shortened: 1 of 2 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -1315,11 +1316,15 @@ audiounit_set_channel_layout(AudioUnit unit,
   if (layout->mChannelLayoutTag == kAudioChannelLayoutTag_Unknown) {
     size = offsetof(AudioChannelLayout, mChannelDescriptions[stream_params->channels]);
     layout = make_sized_audio_channel_layout(size);
+    assert(!layout->mChannelLayoutTag);
+    assert(!layout->mNumberChannelDescriptions);
```
kinetiknz: Two questions:
- would it be better to assert this inside make_sized_audio_channel_layout rather than duplicate it in the callers?
- except: how could these asserts ever fire, since the structure is allocated all-zeros via calloc?

ChunMinChang: I could remove this two lines since it's redundant.

Actually, I just want to confirm the ```offsetof``` works as the expectation. 
The size returned from ```offsetof(AudioChannelLayout, mChannelDescriptions[stream_params->channels]``` must be equal to 
- ```offsetof(AudioChannelLayout, mChannelDescriptions) + stream_params->channels * sizeof(AudioChannelDescription)``` 
- or ```sizeof(AudioChannelLayout) + (stream_params->channels - 1) * sizeof(AudioChannelDescription)```

There is a [crash](https://crash-stats.mozilla.com/report/index/7d358191-c5dc-4062-9778-480840170505) happens when ```audiounit_layout_init(stm, OUTPUT)``` is called, so I guess it's caused by using the invalid memory. If the added assertions are fired, then it implies that we use some invalid address.  

Or could we have other ways to check that?

ChunMinChang: The crash might be caused by the same issue of the #288. The pull 288 is landed in FF55, I could postpone this request after there are crashes in FF55(All the crashes are in FF54).

kinetiknz: I understand what you're trying to detect but I don't think those asserts are the right approach.  AudioChannelLayout is variable-length depending on the space allocated for the final mChannelDescriptions[] array, but everything preceding that is a fixed size and at a fixed location.  The `assert(sz >= sizeof(AudioChannelLayout))` in make_sized_audio_channel_layout guarantees that there is enough space allocated for at least all of the fixed members of AudioChannelLayout plus a single entry in mChannelDescriptions (because the array size declared in the header is `[1]`).  So if `sz` is too small or the allocation fails, we would already be asserting.

kinetiknz:
Closing this on the basis of https://bugzilla.mozilla.org/show_bug.cgi?id=1358896#c18

PR #302, opened 2017-05-18
Push destroy fns to backend

djg:
In preparation for string interning of device id for selected backends, move the common cubeb_device_collection_destroy logic to backends. This allows for allocation other than malloc/free and having strings that statically allocated in cubeb_device_info.

To being with all backends use the same, default cleanup code from cubeb_utils.

### DIFF
--- a/src/cubeb-internal.h
+++ b/src/cubeb-internal.h
@@ -52,6 +52,7 @@ struct cubeb_ops {
   int (* get_preferred_channel_layout)(cubeb * context, cubeb_channel_layout * layout);
   int (* enumerate_devices)(cubeb * context, cubeb_device_type type,
                             cubeb_device_collection ** collection);
+  int (* device_collection_destroy)(cubeb * context, cubeb_device_collection * collection);
   void (* destroy)(cubeb * context);
   int (* stream_init)(cubeb * context,
                       cubeb_stream ** stream,
--- a/src/cubeb.c
+++ b/src/cubeb.c
@@ -571,12 +571,16 @@ int cubeb_enumerate_devices(cubeb * context,
   return rv;
 }
 
-int cubeb_device_collection_destroy(cubeb_device_collection * collection)
+int cubeb_device_collection_destroy(cubeb * context,
+                                    cubeb_device_collection * collection)
 {
-  if (collection == NULL)
+  if (context == NULL || collection == NULL)
     return CUBEB_ERROR_INVALID_PARAMETER;
 
-  return cubeb_utils_default_device_collection_destroy(collection);
+  if (!context->ops->device_collection_destroy)
+    return CUBEB_ERROR_NOT_SUPPORTED;
+
+  return context->ops->device_collection_destroy(context, collection);
 }
 
 int cubeb_register_device_collection_changed(cubeb * context,
--- a/src/cubeb.c
+++ b/src/cubeb.c
@@ -571,31 +571,12 @@ int cubeb_enumerate_devices(cubeb * context,
   return rv;
 }
 
-static void
-cubeb_device_info_destroy(cubeb_device_info * info)
-{
-  XASSERT(info);
-
-  free((void *) info->device_id);
-  free((void *) info->friendly_name);
-  free((void *) info->group_id);
-  free((void *) info->vendor_name);
-
-  free(info);
-}
-
 int cubeb_device_collection_destroy(cubeb_device_collection * collection)
 {
-  uint32_t i;
-
   if (collection == NULL)
     return CUBEB_ERROR_INVALID_PARAMETER;
 
-  for (i = 0; i < collection->count; i++)
-    cubeb_device_info_destroy(collection->device[i]);
-
-  free(collection);
-  return CUBEB_OK;
+  return cubeb_utils_default_device_collection_destroy(collection);
 }
 
 int cubeb_register_device_collection_changed(cubeb * context,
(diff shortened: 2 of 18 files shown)


djg:
@achronop can you review this? If not, @kinetiknz?

src/cubeb-internal.h
```diff
@@ -52,6 +52,7 @@ struct cubeb_ops {
   int (* get_preferred_channel_layout)(cubeb * context, cubeb_channel_layout * layout);
   int (* enumerate_devices)(cubeb * context, cubeb_device_type type,
                             cubeb_device_collection ** collection);
+  int (* device_collection_destroy)(cubeb_device_collection * collection);
```
achronop: Since we have the context at that point why don't we pass it internally to the backends. Maybe we do not need it now but it might be useful for some backends, for example to lock a mutex when `register_device_collection_changed` is on etc. If you think it's not necessary it's fine.

src/cubeb.c
```diff
@@ -11,6 +11,7 @@
 #include <string.h>
 #include "cubeb/cubeb.h"
 #include "cubeb-internal.h"
+#include "cubeb_utils.h"
```
achronop: I'm not sure why you added that header. We do not use anything from it in this file. I believe it can be removed.

src/cubeb_pulse.c
```diff
@@ -1223,7 +1224,7 @@ pulse_sink_info_cb(pa_context * context, const pa_sink_info * info,
   devinfo->type = CUBEB_DEVICE_TYPE_OUTPUT;
   devinfo->state = pulse_get_state_from_sink_port(info->active_port);
   devinfo->preferred = (strcmp(info->name, list_data->default_sink_name) == 0) ?
-				 CUBEB_DEVICE_PREF_ALL : CUBEB_DEVICE_PREF_NONE;
+                                 CUBEB_DEVICE_PREF_ALL : CUBEB_DEVICE_PREF_NONE;
```
achronop: Are all these spacing changes intentional or the result of an automated tool? Same question for all similar changes in other files.

src/cubeb_utils.c
```diff
@@ -0,0 +1,35 @@
...
+
+#include "cubeb_utils.h"
+
+#include <assert.h>
+#include <stdlib.h>
+
+static void
+device_info_destroy(cubeb_device_info * info)
+{
+  free((void *) info->device_id);
```
achronop: Nit: We could assert for `info`.

src/cubeb_utils.c
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright © 2016 Mozilla Foundation
+ *
+ * This program is made available under an ISC-style license.  See the
+ * accompanying file LICENSE for details.
+ */
+
+#include "cubeb_utils.h"
+
+#include <assert.h>
+#include <stdlib.h>
```
achronop: We could make the file .cpp but please discuss it also with @kinetiknz. In this way we can move in all the implementations from the header (not necessarily with this patch).

src/cubeb_utils.h
```diff
@@ -8,6 +8,10 @@
 #if !defined(CUBEB_UTILS)
 #define CUBEB_UTILS
 
+#include "cubeb/cubeb.h"
+
+#ifdef __cplusplus
```
achronop: That can be a good reason to make the file cpp. It's a little confusing the way that this header alters according to the type of compiler it get exposed. But this would require to move the implementations in the .cpp file now.

src/cubeb_pulse.c
djg: I have untabify-on-save in my editor for gecko standards. I believe these lines a falling foul of being reformatting. I've been manually unstaging these changes, but it's a constant battle. I'll setup my editor to ignore this file.

src/cubeb.c
djg: Good spot. I don't know either.

src/cubeb_pulse.c
kinetiknz: There shouldn't be tabs in any libcubeb sources, so feel free to remove the tabs in this file.

src/cubeb_utils.c
djg: We could make it it C++.

src/cubeb_utils.h
djg: But for now, the function I added needs to be called from C code so extern "C" {} would still be required. We could convert all the .c files to be compiled as .cpp, but that would be up to @kinetiknz to decide on. I'd rather see effort to port to rust than C++, but as an interim step, I'm happy to do that work.

kinetiknz: Yeah, let's keep the changes simple here.  We've put the C++ conversion on hold in favour of Rust work; doing anything more than the simplest thing is taking focus away from future efforts.

src/cubeb-internal.h
djg: I think that's fair.

src/cubeb_utils.h
djg: I think cubeb_utils.c is a step along the path anyway to each backend getting it's own, specific clean up function. I foresee it will go away as I continue refactoring.

src/cubeb.c
djg: It was needed for the function cubeb_utils_default_device_collection_destroy().

It's removed in the next commit as it's not needed in that one. Each commit compiles cleanly.

djg:
Hi @achronop  I've made appropriate changes as discussed with @kinetiknz. I force pushed the patched changes to this branch and I think that's confused GH.

PR #304, opened 2017-05-22
CMake: allow googletest and sanitizers-cmake from parent projects

ligfx:
This makes it easier to embed cubeb within larger projects that already provide googletest and/or sanitizers-cmake, by testing directly for what is used rather than indirectly for .git folders.


kinetiknz:
That's much nicer, thanks.  It failed on Travis due to the path to googletest/CMakeLists.txt, so the version I merge prepends PROJECT_SOURCE_DIR.

Merged with that change in 91f801d661d6fcad9d01ea5ba3bbf5d59fa9ac4c

ligfx:
Great, thanks!

Issue #286, opened 2017-04-18
Replace cubeb_device_collection interface.


djg:
In discussion with @kinetiknz, he stated the extending-array-via-mallocing-header+data should be revisited at some point.

padenot:
If/when you're doing this, there is another problem that we need to think about, is that the lifetime of the device themselves is weird, and we're doing all sorts of weird index mapping in gecko.

It would be good to have stable indexes or devices.

djg:
@padenot I've opened Issue #298 to track the problem you've raised. I have patches to improve things.

kinetiknz:
Fixed by #309

PR #309, opened 2017-05-23
Rework device collection

djg:
Rust compatiblity change to cubeb_device_collection.

Convert the C dynamic array with magic header into a structure that hold pointer + count of entries.  This is the same memory layout that rust uses for fat pointers for &[T], Box<[T]>, etc, which will make it much easier to return memory allocated in cubeb-pulse-rust backend.

(Fixes issue #286)

@kinetiknz and I discussed the design options off-line and decided that returning pointer + count was acceptible (breaking the requirement for one allocation for count + storage)

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -29,7 +29,6 @@
 #endif
 #include "cubeb_resampler.h"
 #include "cubeb_ring_array.h"
-#include "cubeb_utils.h"
 #include <algorithm>
 #include <atomic>
 #include <vector>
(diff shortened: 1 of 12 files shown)


djg:
Please flatten into one commit when landing.

src/cubeb_audiounit.cpp
```diff
@@ -2919,10 +2918,10 @@ audiounit_strref_to_cstr_utf8(CFStringRef strref)
 
   len = CFStringGetLength(strref);
   size = CFStringGetMaximumSizeForEncoding(len, kCFStringEncodingUTF8);
-  ret = static_cast<char *>(malloc(size));
+  ret = new char[size+1];
 
   if (!CFStringGetCString(strref, ret, size, kCFStringEncodingUTF8)) {
```
kinetiknz: Doesn't this need to be size+1 for the null byte also?

djg: The documentation for CFStringGetLength() says "Number of 16-bit- Unicode characters in the string.". CFStringGetCString() documentation says "These functions do zero-terminate or put the length byte; the provided bufferSize should include space for this (so pass 256 for Str255)". CFStringGetMaximumSizeForEncoding() returns 3 * length for kCFStringEncodingUTF8.

I'll update code to pass max size + 1 to CFStringGetCString.

Issue #305, opened 2017-05-23
audiounit: cubeb_mixer writes outside output array


djg:
This line looks suspicious and causes test_mixer to write outside the array in 3f2 -> mono downmix.

[heap array out-of-bound write](https://github.com/kinetiknz/cubeb/blob/0c74372fd5e24d3ed0316bbb656a5f5995fc25b0/src/cubeb_mixer.cpp#L280) 

@ChunMinChang please take a look.

ChunMinChang:
Thanks for the report. #310 should solve this!

djg:
@ChunMinChang thanks for the explanation.

kinetiknz:
Fixed by #310

PR #310, opened 2017-05-24
Add osx-specific setting for downmixing test

ChunMinChang:
The ```AudioUnit``` allows us to set the layout containing other channels that the output device can **not** provide. For example, setting *3F2-LFE* layout to a stereo device is fine. When the audio is played, those channels beyond stereo(*Center*, *LFE*, *Left Surround*, *Right Surround*) will be dropped. 

To keep all the information, we need to mix those audio data before it's dropped. That's what we do in #234. We mix the audio before it's played and dropped. In above case, we mixed the audio in *Center*, *LFE*, *Left Surround*, *Right Surround* into *Left* and *Right* channels before playing, while other channels are not be changed. That is,

```
+----+----+----+------+-----+-----+----+----+----+------+-----+-----+
| L0 | R0 | C0 | LFE0 | LS0 | RS0 | L1 | R1 | C1 | LFE1 | LS1 | RS1 | ......
+----+----+----+------+-----+-----+----+----+----+------+-----+-----+

   | Downmix
   v

+----+----+---+---+---+---+----+----+---+---+---+---+
| l0 | r0 | x | x | x | x | l1 | r1 | x | x | x | x | ...
+----+----+---+---+---+---+----+----+---+---+---+---+
            ^   ^   ^   ^             ^   ^   ^   ^
           will be dropped            will be dropped
where
l0 = L0+ 0.707*(C0 + LS0)
r0 = R0+ 0.707*(C0 + RS0)

l1 = L1+ 0.707*(C1 + LS1)
r1 = R1+ 0.707*(C1 + RS1)
...
```

Please see the following discussions for more details:
- https://github.com/kinetiknz/cubeb/pull/234#issuecomment-283855601
- https://github.com/kinetiknz/cubeb/pull/234#issuecomment-283858288
- https://github.com/kinetiknz/cubeb/pull/234#issuecomment-283881631

However, we do **not** add the corresponding changes in ```test_mixer.cpp```, so the invalid address beyond the array buffer can provide will be used and cause errors. That's the cause for #305.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -388,8 +388,9 @@ is_extra_input_needed(cubeb_stream * stm)
 
 static void
 audiounit_mix_output_buffer(cubeb_stream * stm,
+                            long output_frames,
                             void * output_buffer,
-                            long output_frames)
+                            unsigned long output_buffer_length)
 {
   cubeb_stream_params output_mixer_params = {
     stm->output_stream_params.format,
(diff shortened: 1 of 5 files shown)


kinetiknz:
Thanks for the explanation and fixes!

PR #312, opened 2017-05-26
wasapi: use number of frames instead of samples

achronop:
In duplex callback we used to pass through the number of samples instead of frames. This is ok on MONO but it does not work on STEREO capture.

P.S. I need this fix for the stereo gum bug. It will be great to get a quick review. I plan to import it to gecko as soon as possible.


kinetiknz:
LGTM.  The capture-only side has the same bug, so I pushed the equivalent fix for that.

PR #333, opened 2017-06-23
Introduce consistent types and parameters in cubeb public API.

djg:
Address issue 332.

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -37,7 +37,7 @@ extern "C" {
     int latency_frames;
     uint64_t ts;
 
-    rv = cubeb_get_min_latency(app_ctx, output_params, &latency_frames);
+    rv = cubeb_get_min_latency(app_ctx, &output_params, &latency_frames);
     if (rv != CUBEB_OK) {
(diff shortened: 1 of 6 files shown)


djg:
Request review from @kinetiknz since he commented on the issue.

include/cubeb/cubeb.h
```diff
@@ -455,7 +455,7 @@ CUBEB_EXPORT int cubeb_get_max_channel_count(cubeb * context, uint32_t * max_cha
     @retval CUBEB_ERROR_INVALID_PARAMETER
     @retval CUBEB_ERROR_NOT_SUPPORTED */
 CUBEB_EXPORT int cubeb_get_min_latency(cubeb * context,
-                                       cubeb_stream_params params,
+                                       cubeb_stream_params * params,
```
kinetiknz: This should be a const pointer, but I see it's already not for cubeb_stream_init.  We should fix that too.

kinetiknz: Pushed some const fixes to https://github.com/kinetiknz/cubeb/tree/const-fixes, but probably needs fixes on other platforms before merging.

Issue #335, opened 2017-07-06
Use undefined layout as no mapping


achronop:
Undefined layout should be legal and mean that it's not one of the classic mappings (2.0, 2.1, 5.1, etc.). When undefined layout is set the channels must be forwarded in the output without any processing (downmixing or upmixing). Right now undefined layout causes an assert error.

This is discussed in [Bug 1378070#c6](https://bugzilla.mozilla.org/show_bug.cgi?id=1378070#c6)

achronop:
Fixed by #337

PR #337, opened 2017-07-07
Skip mixing when layout is set to undefined

achronop:
This address #335 .

@ChunMinChang can you please have a look?


ChunMinChang:
Looks good to me. What is the use case of the undefined output? Web Audio?

achronop:
Thanks for the review. Yeah, the point of undefined layout is to be able to provide the raw audio to the speakers more useful in WebAudio.

PR #338, opened 2017-07-12
audiounit: default to output stream on aggregate devices

achronop:
StackedKey is set to 0 in order arrange sub-devices of the AudioAggregateDevice such that the output streams are all fed the same data.

Output device is listed before input in order to make output device's stream the default.

Bug 1375729

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1471,6 +1471,11 @@ audiounit_create_blank_aggregate_device(AudioObjectID * plugin_id, AudioDeviceID
   CFDictionaryAddValue(aggregate_device_dict, CFSTR(kAudioAggregateDeviceIsPrivateKey), aggregate_device_private_key);
   CFRelease(aggregate_device_private_key);
 
+  int stacked_key = 0;
+  CFNumberRef aggregate_device_stacked_key = CFNumberCreate(kCFAllocatorDefault, kCFNumberIntType, &stacked_key);
+  CFDictionaryAddValue(aggregate_device_dict, CFSTR(kAudioAggregateDeviceIsStackedKey), aggregate_device_stacked_key);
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -1471,6 +1471,11 @@ audiounit_create_blank_aggregate_device(AudioObjectID * plugin_id, AudioDeviceID
   CFDictionaryAddValue(aggregate_device_dict, CFSTR(kAudioAggregateDeviceIsPrivateKey), aggregate_device_private_key);
   CFRelease(aggregate_device_private_key);
 
+  int stacked_key = 0;
```
kinetiknz: Call this `stacked`.  You're setting the value of a logical entity called `stacked`. The key part is the `kAudioAggregateDeviceIsStackedKey` value when used in a dictionary.

(This applies to private_key above too.)

achronop:
With aggregate device we add devices and not the "side" of a device. This is not usually a problem because a device is either input or output. But this is not the case with soundflower which offer input and output channel on the same device. So we end up with 2 output sub-devices in total. By setting the stacked to 0 sub-devices are arranged to get the same data.

Getting that further I believe it is possible to choose how many (or none) channels you want from a sub-device using the kAudioSubDeviceInputChannelsKey/kAudioSubDeviceOutputChannelsKey during the creation of the AD. That's something I would like to try but it requires some re-arrangement of the methods.

src/cubeb_audiounit.cpp
kinetiknz: This wasn't addressed before merging. :(

achronop: Ooops sorry, I missed to apply this one before the merge. I have it locally but missed to push it. I will do a follow up patch.

PR #339, opened 2017-07-13
Bug 1375411 - [cubeb_openSLES] notify drained if we didn't write any …

alastor0325:
The time stretching would fail on very short audio file.

Since the input/output of time-stretcher is not a linear function, it only returns
result if we feed it enough samples.

When the resampler returns zero sample in this situation, cubeb should call drained
initiative.


kinetiknz:
Thanks!

PR #340, opened 2017-07-20
audiounit: refactor to use device id instead of default device

achronop:
In several places (listeners install/remove etc) we assume that the default device is used. This is correct in many cases but not always. This patch replace the use of the default device with the current device according to the following rules:
* `stream::{in|out}put_device` since the very beginning of stream_init and for the rest of the session contains the current device id. In duplex stream the device id of aggregate device is stored in `stream::aggregate_device_id` and the `stream::{in|out}put_device` contains the selected devices which have been combined to form the AD. 
* when a default device is selected the value of `stream::is_default_{in|out}put` is true
* Install or remove device change listeners will use the current device even in the case of aggregate device.
* Remove listener failure is not a fatal error. This is actually expected when we try to remove a listener on a device that is not alive during reinit. 
* The device that will be used after a device change is always the default (for now) and it is chosen inside reinit.

Finally I remove some duplicate code.

P.S. I plan to improve reinit mechanism on a following patch. I mention in case the above reinit handling does not sound sufficient.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -92,8 +92,8 @@ make_sized_audio_channel_layout(size_t sz)
 }
 
 enum io_side {
-  INPUT = 0x01,
-  OUTPUT = 0x02,
+  INPUT,
+  OUTPUT,
 };
 
 static char const *
@@ -107,6 +107,21 @@ to_string(io_side side)
   }
 }
 
+typedef uint32_t device_flags_value;
+
+enum device_flags {
+  DEV_UKNOWN            = 0x00, /* Unkown */
+  DEV_INPUT             = 0x01, /* Record device like mic */
+  DEV_OUTPUT            = 0x02, /* Playback device like speakers */
+  DEV_SYSTEM_DEFAULT    = 0x04, /* System default device */
+  DEV_SELECTED_DEFAULT  = 0x08, /* User selected to use the system default device */
+};
+
+struct device_info {
+  AudioDeviceID id = kAudioObjectUnknown;
+  device_flags_value flags = DEV_UKNOWN;
+};
+
 struct cubeb_stream {
   explicit cubeb_stream(cubeb * context);
 
@@ -118,10 +133,8 @@ struct cubeb_stream {
   /* Stream creation parameters */
   cubeb_stream_params input_stream_params = { CUBEB_SAMPLE_FLOAT32NE, 0, 0, CUBEB_LAYOUT_UNDEFINED };
   cubeb_stream_params output_stream_params = { CUBEB_SAMPLE_FLOAT32NE, 0, 0, CUBEB_LAYOUT_UNDEFINED };
-  std::atomic<bool> is_default_input{ false };
-  std::atomic<bool> is_default_output{ false };
-  AudioDeviceID input_device = 0;
-  AudioDeviceID output_device = 0;
+  device_info input_device;
+  device_info output_device;
   /* User pointer of data_callback */
   void * user_ptr = nullptr;
   /* Format descriptions */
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -118,7 +118,8 @@ struct cubeb_stream {
   /* Stream creation parameters */
   cubeb_stream_params input_stream_params = { CUBEB_SAMPLE_FLOAT32NE, 0, 0, CUBEB_LAYOUT_UNDEFINED };
   cubeb_stream_params output_stream_params = { CUBEB_SAMPLE_FLOAT32NE, 0, 0, CUBEB_LAYOUT_UNDEFINED };
-  bool is_default_input;
+  std::atomic<bool> is_default_input{ false };
+  std::atomic<bool> is_default_output{ false };
```
kinetiknz: Why were these made atomic?  They should be protected by stm->mutex already.

src/cubeb_audiounit.cpp
```diff
@@ -554,65 +555,14 @@ audiounit_get_backend_id(cubeb * /* ctx */)
...
 static int audiounit_stream_set_volume(cubeb_stream * stm, float volume);
 static int audiounit_uninstall_device_changed_callback(cubeb_stream * stm);
+static AudioObjectID audiounit_get_default_device_id(cubeb_device_type type);
 
 static int
-audiounit_reinit_stream(cubeb_stream * stm)
+audiounit_reinit_stream(cubeb_stream * stm, io_side side)
 {
   auto_lock context_lock(stm->context->mutex);
   assert(stm->input_unit || stm->output_unit);
```
kinetiknz: This assert could be more specific now you pass in io_side.

src/cubeb_audiounit.cpp
```diff
@@ -663,20 +625,21 @@ audiounit_property_listener_callback(AudioObjectID /* id */, UInt32 address_coun
 {
   cubeb_stream * stm = (cubeb_stream*) user;
   stm->switching_device = true;
+  int switch_side(0);
```
kinetiknz: Use the regular initialization syntax.

src/cubeb_audiounit.cpp
```diff
@@ -635,6 +585,18 @@ audiounit_reinit_stream(cubeb_stream * stm)
 
     audiounit_close_stream(stm);
 
+    if (side & INPUT) {
```
kinetiknz: Add a comment describing why we fall back to default devices on reinit.

src/cubeb_audiounit.cpp
```diff
@@ -688,7 +651,7 @@ audiounit_property_listener_callback(AudioObjectID /* id */, UInt32 address_coun
             return noErr;
           }
           // Allow restart to choose the new default. Event register only for input.
-          stm->input_device = 0;
+          switch_side |= INPUT;
```
kinetiknz: Why do we only handle the input side for this?  It looks like we only register when we've got an input_unit, but since it's global scope won't we get this for output as well?

src/cubeb_audiounit.cpp
```diff
@@ -720,7 +683,7 @@ audiounit_property_listener_callback(AudioObjectID /* id */, UInt32 address_coun
   // Use a new thread, through the queue, to avoid deadlock when calling
   // Get/SetProperties method from inside notify callback
   dispatch_async(stm->context->serial_queue, ^() {
-    if (audiounit_reinit_stream(stm) != CUBEB_OK) {
+    if (audiounit_reinit_stream(stm, (io_side)switch_side) != CUBEB_OK) {
```
kinetiknz: static_cast<io_side>(switch_side)

(or reinterpret_cast if the compiler requires it, I don't recall)

src/cubeb_audiounit.cpp
```diff
@@ -1790,71 +1726,61 @@ audiounit_enable_unit_scope(AudioUnit * unit, io_side side, enable_state state)
...
-        r = audiounit_enable_unit_scope(unit, INPUT, DISABLE);
-        if (r != CUBEB_OK) {
-          LOG("Failed to disable audiounit input scope ");
-          return r;
-        }
-        break;
-      default:
-        assert(false);
-    }
+  if (is_default_device && (side == OUTPUT)) {
```
kinetiknz: Don't need parens around side == OUTPUT

src/cubeb_audiounit.cpp
```diff
@@ -1743,7 +1679,7 @@ audiounit_new_unit_instance(AudioUnit * unit, io_side side, AudioDeviceID device
   // so we retain automatic output device switching when the default
   // changes.  Once we have complete support for device notifications
   // and switching, we can use the AUHAL for everything.
-  bool use_default_output = device == 0 && (side == OUTPUT);
+  bool use_default_output = is_default_device && (side == OUTPUT);
```
kinetiknz: Don't need parens around side == OUTPUT

src/cubeb_audiounit.cpp
```diff
@@ -663,20 +625,21 @@ audiounit_property_listener_callback(AudioObjectID /* id */, UInt32 address_coun
...
+  int switch_side(0);
 
   LOG("(%p) Audio device changed, %u events.", stm, (unsigned int) address_count);
   for (UInt32 i = 0; i < address_count; i++) {
     switch(addresses[i].mSelector) {
       case kAudioHardwarePropertyDefaultOutputDevice: {
           LOG("Event[%u] - mSelector == kAudioHardwarePropertyDefaultOutputDevice", (unsigned int) i);
           // Allow restart to choose the new default
-          stm->output_device = 0;
+          switch_side |= OUTPUT;
```
kinetiknz: I think it's a bad idea to treat io_side as flags in some places (i.e. here and inside audiounit_reinit_stream) but to treat it as a strict enum in others (i.e. in audiounit_create_unit, where it expects io_side can only be one of the enum values, not a combination).

To make it clearer, it'd be good to introduce a new type for the cases where io_side needs to be used as flags.  Alternatively, you could convert every use to expect io_side as flags and deal with them appropriately.  You could potentially then extend it with flags for specifying we're using the default in/out rather than passing a separate is_default_device bool everywhere.

src/cubeb_audiounit.cpp
```diff
@@ -1165,7 +1099,7 @@ audiounit_get_preferred_channel_layout()
   return audiounit_convert_channel_layout(layout.get());
 }
 
-static int audiounit_create_unit(AudioUnit * unit, io_side side, AudioDeviceID device);
+static int audiounit_create_unit(AudioUnit * unit, io_side side, AudioDeviceID device, bool is_default_device);
```
kinetiknz: It's kind of unfortunate we have to pass is_default_device everywhere.  I guess we can't determine that we want the default simply by comparing the passed-in device ID against the default, since that doesn't allow us to determine whether we want "the default" (which happens to have a particular ID) or "this particular ID and nothing else".  I wonder if there's a tidier way to deal with this?

src/cubeb_audiounit.cpp
achronop: The scope is not important here. We register the listener for the current input device so it will not listen any other device. We do not expect it for the output. If we want it for the output we have to register the listener for the output device too.

kinetiknz:
Thanks!

PR #345, opened 2017-08-02
Two small fixes for audiounit - aggregate devices

achronop:
- when we create the aggregate device we do not have to set drift compensation flag for the device  which is the master clock.
- when input and output is on the same device (same device id) we do not have to create an aggregate device.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1616,7 +1616,8 @@ audiounit_activate_clock_drift_compensation(const AudioDeviceID aggregate_device
                                                kAudioObjectPropertyScopeGlobal,
                                                kAudioObjectPropertyElementMaster };
 
-  for (UInt32 i = 0; i < subdevices_num; ++i) {
+  // Start from the second device since the first is the master clock
+  for (UInt32 i = 1; i < subdevices_num; ++i) {
     UInt32 drift_compensation_value = 1;
     rv = AudioObjectSetPropertyData(sub_devices[i],
                                     &address_drift,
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -1616,7 +1616,8 @@ audiounit_activate_clock_drift_compensation(const AudioDeviceID aggregate_device
                                                kAudioObjectPropertyScopeGlobal,
                                                kAudioObjectPropertyElementMaster };
 
-  for (UInt32 i = 0; i < subdevices_num; ++i) {
+  // Start from the second device since the first is the master clock
```
kinetiknz: (commenting here since I can't see how to comment on the commit message)

In the commit message, it'd be slightly clearer to say something like "don't unnecessarily set drift compensation on master clock".

src/cubeb_audiounit.cpp
```diff
@@ -1691,6 +1691,7 @@ audiounit_create_aggregate_device(cubeb_stream * stm)
 static int
 audiounit_destroy_aggregate_device(AudioObjectID plugin_id, AudioDeviceID * aggregate_device_id)
 {
+  assert(aggregate_device_id && *aggregate_device_id && plugin_id);
```
kinetiknz: For AudioObjectIDs (and typedefs like AudioDeviceID) it's clearer to compare against the kAudio*Unknown constants, since it's obvious from the source what's happening and doesn't require someone changing the code later to understand that an AudioObjectID of 0 is never a valid ID.  So I'd make this assert something like:

    assert(aggregate_device_id && *aggregate_device_id != kAudioDeviceUnknown && plugin_id != kAudioObjectUnknown);

(with appropriate wrapping, etc.)

kinetiknz:
I went ahead and made those minor changes, since I wanted to uplift another change into m-c.

Merged as https://github.com/kinetiknz/cubeb/commit/9e9678413c4fa263a9b851b240f1158e884b2dd0 and https://github.com/kinetiknz/cubeb/commit/0e10388459fe1e252c075e4a1fab7a9363dc13a1.

Thanks!

Issue #346, opened 2017-08-04
Remove unused CUBEB_STREAM_TYPE_* and cubeb_stream_params.stream_type field


kinetiknz:
These things were added for B2G call management and is now unused.  We can now remove them.

There's a bunch of related code in Gecko using this which is all ifdefed B2G.

kinetiknz:
Fix is in https://github.com/kinetiknz/cubeb/tree/issue346

PR #347, opened 2017-08-06
Remove unused B2G-only stream_type field from cubeb_stream_params.

kinetiknz:
See https://github.com/kinetiknz/cubeb/issues/346

### DIFF
--- a/src/cubeb_opensl.c
+++ b/src/cubeb_opensl.c
@@ -105,7 +105,6 @@ struct cubeb_stream {
   /* Flag indicating draining. Synchronized
    * by stream::mutex lock. */
   int draining;
-  cubeb_stream_type stream_type;
   /* Flags to determine in/out.*/
   uint32_t input_enabled;
   uint32_t output_enabled;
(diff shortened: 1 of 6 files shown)


src/cubeb_opensl.c
```diff
@@ -1241,29 +1214,6 @@ opensl_configure_playback(cubeb_stream * stm, cubeb_stream_params * params) {
...
-    SLAndroidConfigurationItf playerConfig;
-    res = (*stm->playerObj)->GetInterface(stm->playerObj,
-                                          stm->context->SL_IID_ANDROIDCONFIGURATION,
-                                          &playerConfig);
-    if (res != SL_RESULT_SUCCESS) {
-      LOG("Failed to get android configuration interface. Error code: %lu", res);
-      return CUBEB_ERROR;
-    }
-
-    res = (*playerConfig)->SetConfiguration(playerConfig,
```
kinetiknz: SL_ANDROID_STREAM_MEDIA (== CUBEB_STREAM_TYPE_MUSIC) is the default according to https://developer.android.com/ndk/guides/audio/opensl/android-extensions.html

kinetiknz:
@achronop Do you mind reviewing this as well? Mainly for the Android bits and just so you're aware we're doing this.

achronop:
Why did you decide to remove that? It may be better to try to use it in gecko side rather than remove it. For example we could use MEDIA for WebAudio or playback and VOICE_CALL for WebRTC. I see that chrome does something similar:
https://cs.chromium.org/chromium/src/media/audio/android/audio_manager_android.cc?l=234

achronop:
I reviewed it, it's good to go, I do not have any comment just the question above.

kinetiknz:
> Why did you decide to remove that?

It was added in a rush and never properly designed, leaving us with the ugly hack of a platform-specific field in the public API and exposing the Android/B2G stream type constants directly without consideration of how they are handled on other platforms.

The code is dead and untested since B2G development stopped.  Keeping dead code around has real maintenance costs and zero benefits.

> It may be better to try to use it in gecko side rather than remove it. For example we could use MEDIA for WebAudio or playback and VOICE_CALL for WebRTC.

If you think it's worth supporting a better designed (i.e. portable) and properly maintained implementation of this feature, file a bug in BMO and make a case for it there.

padenot:
I agree with @kinetiknz here, this is not ideal and was done too quickly. In practice, the following needs to happen:
- Have specs recognize the fact that authors need to be able to express intent when doing something that will end up creating an audio stream (i.e., is this AudioContext is going to be used for speech, music, something else ?)
- Make a proper API in cubeb, that will abstract this over different platforms (on Windows, you can use the "communication device", and on Android "VOICE_CALL" as noted. iirc there is something similar for pulse, and we could use a `kAudioUnitSubType_VoiceProcessingIO` on OSX)

PR #348, opened 2017-08-08
resample.c: remove unused variable

avindra:
minor change to remove an unused variable


kinetiknz:
Thanks for the patch!  This file (and everything in src/speex) is imported from the upstream speexdsp project (https://github.com/xiph/speexdsp), so I'd prefer not to merge this as-is.  Two options to consider:

1. Submit your fix upstream and then update cubeb's import once the fix is merged.
2. Change cubeb's build system to disable warnings for src/speex.

avindra:
Oh word. Thanks for pointing me the right way I'll head over there!

kinetiknz:
Thanks!  I've imported your fix from upstream via a8460cb504506e3c0e7dc4224d8993a8e64c0367.

PR #349, opened 2017-08-11
Fix a gcc warning with -Wparentheses

sylvestre:
suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]


padenot:
Thanks Sylvestre !

PR #350, opened 2017-08-14
Add a thread reset API for the ring buffer class.

### LINKED ISSUE #320: wasapi breaks threading assumption made by ring_buffer_base
cubeb uses multiple threads to implement a fast logger. The logger uses `ring_buffer_base` to pass messages from log producer to log consumer. `ring_buffer_base` [asserts that the thread_id of the producer and consumer are constant](https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_ringbuffer.h#L317).

However, the wasapi implementation (at least - i haven't tested others) will create and destroy producer threads dynamically. As an aside, the consumer is a detached thread which seems kind of gross.

I've made a write-up here: https://github.com/dolphin-emu/dolphin/pull/5256#issuecomment-305950712

padenot:
This allow reseting the consumer/producer thread id in the ring buffer
class, so that we can take into account the change in thread inherent to
the audio devices changes.

This fixes #320.

### DIFF
--- a/test/test_ring_buffer.cpp
+++ b/test/test_ring_buffer.cpp
@@ -155,6 +155,32 @@ void basic_api_test(T& ring)
   ASSERT_EQ(ring.available_write(), 128);
 }
 
+void test_reset_api() {
+	const size_t RING_BUFFER_SIZE = 128;
+	const size_t ENQUEUE_SIZE = RING_BUFFER_SIZE / 2;
+
+	lock_free_queue<float> ring(RING_BUFFER_SIZE);
+	std::thread t([&ring, ENQUEUE_SIZE] {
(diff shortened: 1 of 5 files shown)


test/test_ring_buffer.cpp
```diff
@@ -155,6 +155,33 @@ void basic_api_test(T& ring)
...
+	});
+
+	t.join();
+
+	ring.reset_thread_ids();
+
+	// Enqueue with a different thread. We have reset the thread ID
+	// in the ring buffer, this should work.
+	std::thread t2([&ring, ENQUEUE_SIZE] {
+		std::unique_ptr<float> in_buffer(new float[ENQUEUE_SIZE]);
```
kinetiknz: Needs to be `std::unique_ptr<float[]>` otherwise you have mismatched allocators (ASAN caught this on Travis)

PR #354, opened 2017-09-07
Implement some rudimentary underrun protection for duplex streams

padenot:
For now, it leaves at most 100ms in the buffer. It seem to behave well. I've used this to test:

```diff
diff --git a/test/test_duplex.cpp b/test/test_duplex.cpp
index 427b503..f17634d 100644
--- a/test/test_duplex.cpp
+++ b/test/test_duplex.cpp
@@ -17,11 +17,14 @@
 #include <memory>
 #include "cubeb/cubeb.h"
 #include "common.h"
+#include <iostream>
 #include <atomic>

 #define SAMPLE_FREQUENCY 48000
 #define STREAM_FORMAT CUBEB_SAMPLE_FLOAT32LE

+static std::atomic<int> rounds = 0;
+
 struct user_state_duplex
 {
   std::atomic<int> seen_audio{ 0 };
@@ -50,7 +53,11 @@ long data_cb_duplex(cubeb_stream * stream, void * user, const void * inputbuffer
     output_index += 2;
   }

-  u->seen_audio |= seen_audio;
+  for (uint32_t i = 0; i < nframes * rounds; i++) {
+    if (ib[i % nframes] <= -1.0 && ib[i % nframes] >= 1.0) {
+      u->seen_audio |= seen_audio;
+    }
+  }

   return nframes;
 }
@@ -118,8 +125,13 @@ TEST(cubeb, duplex)
     cleanup_stream_at_exit(stream, cubeb_stream_destroy);

   cubeb_stream_start(stream);
-  delay(500);
-  cubeb_stream_stop(stream);
+  std::string str;
+  while (getline(std::cin, str)) {
+    std::stringstream iss(str);
+    int temp;
+    iss >> temp;
+    rounds = temp;
+  }

   ASSERT_TRUE(stream_state.seen_audio.load());
 }

```

On my windows machine, I can make it glitch by typing, say, `5000<enter>`, and then bring it back to normal by doing `1000<enter>`.

If I backout the actual patch in this PR, delay accumulates between input and output.

### DIFF
--- a/src/cubeb_resampler.cpp
+++ b/src/cubeb_resampler.cpp
@@ -39,8 +39,10 @@ template<typename T>
 passthrough_resampler<T>::passthrough_resampler(cubeb_stream * s,
                                                 cubeb_data_callback cb,
                                                 void * ptr,
-                                                uint32_t input_channels)
+                                                uint32_t input_channels,
+                                                uint32_t sample_rate)
   : processor(input_channels)
+  , sample_rate(sample_rate)
   , stream(s)
   , data_callback(cb)
   , user_ptr(ptr)
@@ -73,6 +75,7 @@ long passthrough_resampler<T>::fill(void * input_buffer, long * input_frames_cou
   if (input_buffer) {
     internal_input_buffer.pop(nullptr, frames_to_samples(output_frames));
     *input_frames_count = output_frames;
+    drop_audio_if_needed();
   }
 
   return rv;
@@ -205,6 +208,7 @@ cubeb_resampler_speex<T, InputProcessor, OutputProcessor>
   /* The number of frames returned from the callback. */
   long got = 0;
 
+
   /* We need to determine how much frames to present to the consumer.
    * - If we have a two way stream, but we're only resampling input, we resample
    * the input to the number of output frames.
@@ -242,9 +246,18 @@ cubeb_resampler_speex<T, InputProcessor, OutputProcessor>
 
   output_processor->written(got);
 
+  /* If either the input buffer or the output buffer grew too big, drop some
+     audio. This will glitch, but will prevent delay buildups. We drop before
+     adding the new audio in, because we always want to read/write recent audio. */
+  input_processor->drop_audio_if_needed();
+
   /* Process the output. If not enough frames have been returned from the
    * callback, drain the processors. */
-  return output_processor->output(out_buffer, output_frames_needed);
+  got = output_processor->output(out_buffer, output_frames_needed);
+
+  output_processor->drop_audio_if_needed();
+
+  return got;
 }
 
 /* Resampler C API */
--- a/src/cubeb_resampler.cpp
+++ b/src/cubeb_resampler.cpp
@@ -43,9 +43,9 @@ passthrough_resampler<T>::passthrough_resampler(cubeb_stream * s,
                                                 uint32_t sample_rate)
   : processor(input_channels)
   , stream(s)
-  , sample_rate(sample_rate)
   , data_callback(cb)
   , user_ptr(ptr)
+  , sample_rate(sample_rate)
 {
 }
 
--- a/src/cubeb_resampler.cpp
+++ b/src/cubeb_resampler.cpp
@@ -42,8 +42,8 @@ passthrough_resampler<T>::passthrough_resampler(cubeb_stream * s,
                                                 uint32_t input_channels,
                                                 uint32_t sample_rate)
   : processor(input_channels)
-  , sample_rate(sample_rate)
   , stream(s)
+  , sample_rate(sample_rate)
   , data_callback(cb)
   , user_ptr(ptr)
 {
@@ -208,7 +208,6 @@ cubeb_resampler_speex<T, InputProcessor, OutputProcessor>
   /* The number of frames returned from the callback. */
   long got = 0;
 
-
   /* We need to determine how much frames to present to the consumer.
    * - If we have a two way stream, but we're only resampling input, we resample
    * the input to the number of output frames.
@@ -246,9 +245,6 @@ cubeb_resampler_speex<T, InputProcessor, OutputProcessor>
 
   output_processor->written(got);
 
-  /* If either the input buffer or the output buffer grew too big, drop some
-     audio. This will glitch, but will prevent delay buildups. We drop before
-     adding the new audio in, because we always want to read/write recent audio. */
   input_processor->drop_audio_if_needed();
 
   /* Process the output. If not enough frames have been returned from the
--- a/src/cubeb_resampler_internal.h
+++ b/src/cubeb_resampler_internal.h
@@ -38,6 +38,17 @@ MOZ_END_STD_NAMESPACE
 #include <stdio.h>
 
 /* This header file contains the internal C++ API of the resamplers, for testing. */
+namespace {
+// When dropping audio input frames to prevent building
+// an input delay, this function returns the number of frames
+// to keep in the buffer.
+// @parameter sample_rate The sample rate of the stream.
+// @return A number of frames to keep.
+uint32_t min_buffered_audio_frame(uint32_t sample_rate)
+{
+  return sample_rate / 20;
+}
+}
 
 int to_speex_quality(cubeb_resampler_quality q);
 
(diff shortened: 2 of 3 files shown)


kinetiknz:
@ChunMinChang Would you mind reviewing this PR please?

src/cubeb_resampler_internal.h
```diff
@@ -332,6 +353,8 @@ class cubeb_resampler_speex_one_way : public processor {
   SpeexResamplerState * speex_resampler;
   /** Source rate / target rate. */
   const float resampling_ratio;
+  const uint32_t source_rate;
+  const uint32_t target_rate;
```
achronop: There is no usage of the two new members

src/cubeb_resampler_internal.h
```diff
@@ -164,6 +174,8 @@ class cubeb_resampler_speex_one_way : public processor {
                                 int quality)
   : processor(channels)
   , resampling_ratio(static_cast<float>(source_rate) / target_rate)
+  , source_rate(source_rate)
+  , target_rate(target_rate)
```
achronop: No usage of the two new members.

ChunMinChang:
I think I am not able to help this now since I am not familiar with this part and I am working on seamless looping. Sorry for that.

src/cubeb_resampler.cpp
```diff
@@ -242,9 +246,18 @@ ::fill_internal_duplex(T * in_buffer, long * input_frames_count,
 
   output_processor->written(got);
 
+  /* If either the input buffer or the output buffer grew too big, drop some
+     audio. This will glitch, but will prevent delay buildups. We drop before
+     adding the new audio in, because we always want to read/write recent audio. */
+  input_processor->drop_audio_if_needed();
```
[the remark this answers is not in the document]
padenot: Yes. You always have a processor, regardless of if you have an input buffer.

padenot:
Updated, please have a look. I'm not sure how much we should keep. 100ms is pretty random. I'll do some testing next week.

padenot:
I've updated to 50ms. Note that a single callback usually has 10ms worth of audio on on Windows.

@achronop, can you have a final look here, so I can land and get this in Nightly ?

src/cubeb_resampler_internal.h
```diff
@@ -175,7 +186,6 @@ class cubeb_resampler_speex_one_way : public processor {
   : processor(channels)
   , resampling_ratio(static_cast<float>(source_rate) / target_rate)
   , source_rate(source_rate)
```
achronop: source_rate is also unused.

src/cubeb_resampler_internal.h
```diff
@@ -354,7 +365,6 @@ class cubeb_resampler_speex_one_way : public processor {
   /** Source rate / target rate. */
   const float resampling_ratio;
   const uint32_t source_rate;
```
achronop: source_rate is also unused.

src/cubeb_resampler_internal.h
```diff
@@ -88,9 +99,9 @@ class passthrough_resampler : public cubeb_resampler
 
   void drop_audio_if_needed()
   {
-    // only keep at most 100ms of audio
-    if (samples_to_frames(internal_input_buffer.length()) > sample_rate / 10) {
-      internal_input_buffer.pop(nullptr, samples_to_frames(internal_input_buffer.length()) - sample_rate / 10);
+    uint32_t to_keep = min_buffered_audio_frame(sample_rate);
+    if (samples_to_frames(internal_input_buffer.length()) > to_keep) {
+      internal_input_buffer.pop(nullptr, samples_to_frames(internal_input_buffer.length()) - to_keep);
```
achronop: I think you need:
`internal_input_buffer.length() - frames_to_samples(to_keep)`
or
`frames_to_samples(samples_to_frames(internal_input_buffer.length()) - to_keep)`

src/cubeb_resampler_internal.h
```diff
@@ -471,8 +482,10 @@ class delay_line : public processor {
 
   void drop_audio_if_needed()
   {
-    if (samples_to_frames(delay_input_buffer.length()) > sample_rate / 10) {
-      delay_input_buffer.pop(nullptr, delay_input_buffer.length() - frames_to_samples(length));
+    size_t available = samples_to_frames(delay_input_buffer.length());
+    uint32_t to_keep = min_buffered_audio_frame(sample_rate);
+    if (available > to_keep) {
+      delay_input_buffer.pop(nullptr, available - frames_to_samples(sample_rate));
```
achronop: I think you need: `frames_to_samples(available - to_keep)`

test/test_resampler.cpp
```diff
@@ -755,3 +767,89 @@ TEST(cubeb, resampler_passthrough_duplex_callback_reordering)
...
+      frames = 2 * BUF_BASE_SIZE * UNDERRUN_FACTOR;
+      got = cubeb_resampler_fill(resampler, input_buffer_glitch, &frames, output_buffer, BUF_BASE_SIZE);
+      is_seq(output_buffer, 2, BUF_BASE_SIZE, output_seq_idx);
+      output_seq_idx += BUF_BASE_SIZE;
+    } else if (i != 0 && (i % 100) == 2) {
+      // On the next iteration, the sequence should be broken
+      seq_idx = seq(input_buffer_normal, input_channels, seq_idx, BUF_BASE_SIZE);
+      long normal_input_frame_count = 256;
+      got = cubeb_resampler_fill(resampler, input_buffer_normal, &normal_input_frame_count, output_buffer, BUF_BASE_SIZE);
+      is_not_seq(output_buffer, 2, BUF_BASE_SIZE, output_seq_idx);
```
achronop: The most important think to verify here is that we have at most a difference of `min_buffered_audio_frame(sample_rate)` (2205 frames in that case). Since you hold the indexes you can do that, something like:
`ASSERT_EQ(seq_idx - output_buffer[BUF_BASE_SIZE * 2 - 1] -1, 2205)`

jesup:
FYI, from chatting on Slack:
The "right" way to do it is to a) resample to match rates, b) keep enough buffer to avoid underruns, but no more.  This obviously can fail if the environment/load changes, which is why it has to be adaptive (basically it has to be a form of jitter buffer).  In most cases, it should run at "no extra buffering" (i.e. we haven't seen underruns) which means max one capture "frame" (typ 10ms or less - and if you're not having to resample (no drift), it should be 0 in that case.  When the buffer starts to grow, you slew the resample rate to stop the growth and get it back near your target depth.  If you see underruns, you glitch, and you slew the resample rate to increase the amount of buffering (and bump the target for buffering probably). 

This is classic jitter-buffer design.  For "extra" fancy, you can minimize pitch warbles when slewing (and slew faster) by using some of the magic in NetEq to add or remove cycles of the dominant frequency (TDC - Time Distortion Compensation IIRC)

IMHO

test/test_resampler.cpp
```diff
@@ -755,3 +767,89 @@ TEST(cubeb, resampler_passthrough_duplex_callback_reordering)
...
+      // Reclock so that we can use is_seq again.
+      output_seq_idx = output_buffer[BUF_BASE_SIZE * 2 - 1] + 1;
+    } else {
+       // normal case
+      seq_idx = seq(input_buffer_normal, input_channels, seq_idx, BUF_BASE_SIZE);
+      long normal_input_frame_count = 256;
```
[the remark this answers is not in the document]
padenot: This needs to be mutable.

test/test_resampler.cpp
```diff
@@ -755,3 +767,89 @@ TEST(cubeb, resampler_passthrough_duplex_callback_reordering)
...
+                                  output_buffer, BUF_BASE_SIZE);
+
+  output_seq_idx += BUF_BASE_SIZE;
+
+  // prebuffer_frames will hold the frames used by the resampler.
+  ASSERT_EQ(prebuffer_frames, BUF_BASE_SIZE);
+  ASSERT_EQ(got, BUF_BASE_SIZE);
+
+  for (uint32_t i = 0; i < 300; i++) {
+    long int frames = BUF_BASE_SIZE;
```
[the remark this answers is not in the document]
padenot: I need a mutable variable here.

test/test_resampler.cpp
padenot: I'd rather not make assumption about the internals here.

achronop: Not a blocking point, just a friendly remark that I consider that useful.

PR #356, opened 2017-09-13
audiounit: reinit stream on DataSource event (Bug 1398788)

achronop:
Due to Bug 1398788 full reinit is performed on DataSource event like we did in the past.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -674,7 +674,7 @@ audiounit_reinit_stream(cubeb_stream * stm, device_flags_value flags)
 }
 
 static char const *
-to_string(AudioObjectPropertySelector selector)
+event_addr_to_string(AudioObjectPropertySelector selector)
 {
   switch(selector) {
     case kAudioHardwarePropertyDefaultOutputDevice:
@@ -697,7 +697,7 @@ audiounit_property_listener_callback(AudioObjectID id, UInt32 address_count,
 {
   cubeb_stream * stm = (cubeb_stream*) user;
   if (stm->switching_device) {
-    LOG("Switch is taking place. Skip Event %s for id=%d", to_string(addresses[0].mSelector), id);
+    LOG("Switching is already taking place. Skip Event %s for id=%d", event_addr_to_string(addresses[0].mSelector), id);
     return noErr;
   }
   stm->switching_device = true;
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -673,48 +673,72 @@ audiounit_reinit_stream(cubeb_stream * stm, device_flags_value flags)
...
           switch_side |= DEV_INPUT;
         }
       break;
       case kAudioDevicePropertyDeviceIsAlive: {
-          LOG("Event[%u] - mSelector == kAudioDevicePropertyDeviceIsAlive", (unsigned int) i);
+          LOG("Event[%u] - mSelector == kAudioDevicePropertyDeviceIsAlive for id=%d", (unsigned int) i, id);
           // If this is the default input device ignore the event,
           // kAudioHardwarePropertyDefaultInputDevice will take care of the switch
           if (stm->input_device.flags & DEV_SYSTEM_DEFAULT) {
             LOG("It's the default input device, ignore the event");
```
padenot: Does it really switch automatically if the default input device changes? Maybe we should just re-init in all cases.

src/cubeb_audiounit.cpp
```diff
@@ -673,48 +673,72 @@ audiounit_reinit_stream(cubeb_stream * stm, device_flags_value flags)
   return CUBEB_OK;
 }
 
+static char const *
+to_string(AudioObjectPropertySelector selector)
```
padenot: Change the name of this one, `to_string` is too generic.

src/cubeb_audiounit.cpp
```diff
@@ -673,48 +673,72 @@ audiounit_reinit_stream(cubeb_stream * stm, device_flags_value flags)
...
+
 static OSStatus
-audiounit_property_listener_callback(AudioObjectID /* id */, UInt32 address_count,
+audiounit_property_listener_callback(AudioObjectID id, UInt32 address_count,
                                      const AudioObjectPropertyAddress * addresses,
                                      void * user)
 {
   cubeb_stream * stm = (cubeb_stream*) user;
+  if (stm->switching_device) {
+    LOG("Switch is taking place. Skip Event %s for id=%d", to_string(addresses[0].mSelector), id);
```
padenot: Change the message "Switching is already taking place...".

src/cubeb_audiounit.cpp
achronop: We do reinit in all case now. The reason we ignore that event is to avoid double reinit when the default input is unplugged. In that case we get an IsAlive event and we know that a second event of type DefaultInputDevice will follow. So we allow the stream to reinit on that. Please note that this patch, other than resetting the switch flag properly, does not touch this scenario which works like that for a long time.

achronop:
Let me squash before merge, when it is ready.

PR #358, opened 2017-09-22
audiounit: expect error when set device info.

achronop:
Return cubeb error status from `audiounit_set_device_info` since `audiounit_get_default_device_id` can fail.

Error found in the following crash id:
https://crash-stats.mozilla.com/report/index/920b342c-e152-471d-ae16-fcb800170919

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -576,7 +576,7 @@ static int audiounit_stream_set_volume(cubeb_stream * stm, float volume);
 static int audiounit_uninstall_device_changed_callback(cubeb_stream * stm);
 static AudioObjectID audiounit_get_default_device_id(cubeb_device_type type);
 
-static void
+static int
 audiounit_set_device_info(cubeb_stream * stm, AudioDeviceID id, io_side side)
 {
   assert(stm);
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -645,12 +650,14 @@ audiounit_reinit_stream(cubeb_stream * stm, device_flags_value flags)
      * default system device change. In both cases cubeb switch on the new default
      * device. This is considered the most expected behavior for the user. */
     if (flags & DEV_INPUT) {
-      audiounit_set_device_info(stm, 0, INPUT);
+      r = audiounit_set_device_info(stm, 0, INPUT);
+      assert(r == CUBEB_OK);
```
padenot: Is it ok to assert here? This can happen, right?

src/cubeb_audiounit.cpp
```diff
@@ -645,12 +650,14 @@ audiounit_reinit_stream(cubeb_stream * stm, device_flags_value flags)
...
     }
     /* Always use the default output on reinit. This is not correct in every case
      * but it is sufficient for Firefox and prevent reinit from reporting failures.
      * It will change soon when reinit mechanism will be updated. */
-    audiounit_set_device_info(stm, 0, OUTPUT);
+    r = audiounit_set_device_info(stm, 0, OUTPUT);
+    assert(r == CUBEB_OK);
```
padenot: Same.

src/cubeb_audiounit.cpp
achronop: I see the failure reported on init only. I believe it has something to do with the initial configuration. Probably, if the init works it does not fail on reinit. I left it like that in order to confirm (or not) my theory. I can change it if you insist.

padenot: Ok. We have been annoying people recently with assert crashes in Nightly. Hopefully this will be useful.

Issue #359, opened 2017-09-25
Fix BMO #1399978 in cubeb_pulse.c


kinetiknz:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1399978

kinetiknz:
Closed via  #360

PR #360, opened 2017-09-25
pulse: Improve handling when there are no pulse source/sink.

djg:
Fixes issue #359

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -653,7 +653,6 @@ pulse_init(cubeb ** context, char const * context_name)
     WRAP(pa_operation_unref)(o);
   }
   WRAP(pa_threaded_mainloop_unlock)(ctx->mainloop);
-  assert(ctx->default_sink_info);
 
   *context = ctx;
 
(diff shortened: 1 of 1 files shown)


src/cubeb_pulse.c
```diff
@@ -1355,8 +1369,10 @@ pulse_server_info_cb(pa_context * c, const pa_server_info * i, void * userdata)
 
   free(list_data->default_sink_name);
   free(list_data->default_source_name);
-  list_data->default_sink_name = strdup(i->default_sink_name);
-  list_data->default_source_name = strdup(i->default_source_name);
+  if (i->default_sink_name)
```
achronop: Since default_sink/source_name expected to be `null` it would be better to initialize `list_data->default_sink/source_name = nullptr` after the free and before check.

src/cubeb_pulse.c
```diff
@@ -1236,12 +1247,12 @@ pulse_sink_info_cb(pa_context * context, const pa_sink_info * info,
   (void)context;
 
   if (eol || info == NULL)
-    return;
+    goto exit;
```
achronop: Good point. Did we deadlock without that change if there is no sink?

src/cubeb_pulse.c
djg: The if is required because strdup doesn't handle null ptr safely. ie. strdup(NULL) results in a null ptr dereference.

achronop: Indeed if check is required. I am talking about `list_data->default_sink/source_name` which is not initialized after the free so if `i->default_sink/source_name` is null we may have read after free.

djg:
I'll tighten up the logic tomorrow and post a follow-up patch.

PR #363, opened 2017-10-04
CMake: Generate and install CMake config files

sfuller:
I modified CMakeLists.txt to support generating and installing CMake config files. These config files should make using cubeb from other CMake projects much easier.


kinetiknz:
LGTM, thanks!

PR #365, opened 2017-10-12
audiounit: fix deadlock on create AD when collection cb is registered

achronop:
Deadlock is created because `create/destroy_aggregate_device()` and `device_collection_change_callabck()` are called with the `context->mutex` held. When we create or destroy an AD the callback is invoked (if registered). 

My solution makes use of a "flag" to signal that an AD is creating/destroying in order to return from the callback without doing anything (not locking the mutex).

The other solution would be to remove the collection listener before AD create/destroy and re-add it when it's done.

I found the first solution less complicated and cheaper since we avoid to call the framework API (which is full of surprises).

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -83,10 +83,7 @@ struct cubeb {
   // Current used channel layout
   std::atomic<cubeb_channel_layout> layout{ CUBEB_LAYOUT_UNDEFINED };
   // Flag to be used in device_collection_callback to avoid deadlock.
-  // This is a different use of share_ptr. The important is not the value
-  // but the ability to track all parallel instances synchronized. When the
-  // share_ptr is unique means that there is no active create or delete in any thread.
-  std::shared_ptr<const bool> add_remove_aggregate_device{ new bool(true) };
+  std::atomic<int> creating_removing_aggregate_device{ 0 };
 };
 
 static std::unique_ptr<AudioChannelLayout, decltype(&free)>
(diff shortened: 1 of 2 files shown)


padenot:
Please write a test. Simply trying to create a new stream with the device list change callback set up should be enough.

Why is this not an atomic int or something? A bit like we do for the `active_streams` but for counting aggregate devices.

achronop:
> Why is this not an atomic int or something? A bit like we do for the active_streams but for counting aggregate devices.

Just because we get the atomic counting out of box. But I understand it's not very intuitive. I'll replace it with an atomic counter.

src/cubeb_audiounit.cpp
```diff
@@ -1700,12 +1697,10 @@ static int audiounit_destroy_aggregate_device(AudioObjectID plugin_id, AudioDevi
...
 audiounit_create_aggregate_device(cubeb_stream * stm)
 {
-  int r = 0;
-  {
-    // Signal that aggregate device is creating to be used by device_collection_change_callback
-    std::shared_ptr<const bool> creating_aggregate_device = stm->context->add_remove_aggregate_device;
-    r = audiounit_create_blank_aggregate_device(&stm->plugin_id, &stm->aggregate_device_id);
-  }
+  // Signal that aggregate device is creating to be used by device_collection_change_callback
+  stm->context->creating_removing_aggregate_device += 1;
```
padenot: If we _really_ plug or unplug a device after having incremented this atomic, we'll miss the real callback. I think we need to lock here. We need atomicity for several operations.

Another approach would be to determine that we're getting a notification about an aggregate device we've just created, but we'd need the ID, and I don't know that we have it at this point.

achronop:
> If we really plug or unplug a device after having incremented this atomic, we'll miss the real callback.

Since the id is not available that's a compromise that I do not know how to avoid. 

> I think we need to lock here. We need atomicity for several operations.

Callback is called in the same thread before `audiounit_create_blank_aggregate_device()` method finish (same call stack) if we use a mutex it will deadlock when we try to lock that mutex inside the collection changed callback.

> Another approach would be to determine that we're getting a notification about an aggregate device we've just created, but we'd need the ID, and I don't know that we have it at this point.

The id is not available at that point _and_ the id that caused the callback is not provided from the system.

padenot:
Any idea here ?

padenot:
I've fixed this in a different way, PR incoming.

PR #367, opened 2017-10-17
audiounit: advance active stream counter earlier in stream init (Bug 1409216)

achronop:
Fix for crash in bug 1409216. The active stream counter is advanced right after the allocation of the new cubeb stream.


padenot:
I think this happens because `audiounit_set_device_info` fail. The patch looks like it would fix this crash, but I'd like to understand why this happens.

achronop:
I suspect that it happens when there are no media devices in the system. I asked @ChunMinChang to make a test since they have a mini mac without devices.

In any case we can go on with that and open a separate issue about the root cause.

PR #368, opened 2017-10-18
audiounit: use one instance of audiounit on duplex stream (Bug 1404969)

achronop:
AudioUnit can be configured for duplex operation when both device ids are the same for input and output. This is always the case since we configure an aggregate device on duplex stream. This patch remove the input AU and makes use of the same AU for input and output. A side effect on the positive side is that switching become cheaper.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -143,8 +143,7 @@ struct cubeb_stream {
   AudioStreamBasicDescription input_desc;
   AudioStreamBasicDescription output_desc;
   /* I/O AudioUnits */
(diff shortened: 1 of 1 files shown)


padenot:
Please have a look at the travis failure, it looks serious.

achronop:
The new push is just for testing. I'll tell when is ready.

achronop:
This is ready now. You can go on and review. When review is done please do not merge. If we are late in the release cycle at that point I would like to wait the next one. Also I would like to import it as a single commit.

padenot (approved):

PR #371, opened 2017-10-24
audiounit: reinit fail when no media devices (Bug 1409224)

achronop:
Return CUBEB_ERROR when set device info fails on reinit. This is expected when the last media device is unplugged.


kinetiknz:
Thanks!

PR #372, opened 2017-10-25
Elaborate the build instructions for Windows

SingingTree:
Add mention of visual studio 2017, add instructions on how to use CMake from within VS2017. Point out gotcha of default build dirs in VS2017.


kinetiknz:
Thanks!

PR #374, opened 2017-10-26
audiounit: change state to error on failed reinit.

achronop:
On failed reinit we go to error instead of stop state. This will align the behavior with the other backends.


padenot:
Thanks. We should really think about testing, it's insane the number of bugs we've had in device switching and other stuff.

PR #375, opened 2017-10-26
Update comments in cubeb-resampler.h

SingingTree:
Update some comments which appear to have become out of step with their respective code.


kinetiknz:
Thanks!

PR #376, opened 2017-10-30
add support for recording in sndio backend

ratchov:
This adds recording and full-duplex support for the sndio backend. Most of the changes prepare the existing code by renaming symbols and by fixing constucts that make difficult implementation of the recording bits.

### DIFF
--- a/src/cubeb_sndio.c
+++ b/src/cubeb_sndio.c
@@ -220,12 +220,12 @@ sndio_stream_init(cubeb * context,
   s = malloc(sizeof(cubeb_stream));
   if (s == NULL)
     return CUBEB_ERROR;
+  memset(s, 0, sizeof(cubeb_stream));
   s->context = context;
   s->hdl = sio_open(NULL, SIO_PLAY, 1);
   if (s->hdl == NULL) {
-    free(s);
     DPR("sndio_stream_init(), sio_open() failed\n");
-    return CUBEB_ERROR;
+    goto err;
   }
   sio_initpar(&wpar);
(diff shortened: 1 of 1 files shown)


padenot:
Amazing, @ratchov, thanks.

I'll have a look over the next few days.

landryb:
fwiw, i've tested this on OpenBSD using https://webaudiodemos.appspot.com/AudioRecorder/index.html for plain recording, and bidirectional audioconf via appear.in between two laptops running nightly+this patchset+enforcing full_duplex pref.

landryb:
Back-reference to https://bugzilla.mozilla.org/show_bug.cgi?id=1221580 for integration in gecko's copy of cubeb.

src/cubeb_sndio.c
```diff
@@ -283,9 +283,12 @@ sndio_get_max_channel_count(cubeb * ctx, uint32_t * max_channels)
...
 {
-  // XXX Not yet implemented.
-  *rate = 44100;
-
+  /*
+   * We've no device-independent prefered rate; any rate will work if
+   * sndiod is running. If it isn't, 48kHz is what is most likely to
+   * work as most (but not all) devices support it.
+   */
+  *rate = 48000;
```
padenot: In practice, this is better than 44.1kHz, because this triggers fast paths down in webrtc code (in particular, this lets us skip a number of downsampling and upsampling passes).

padenot:
This all looks fine (as far as the cubeb model is concern).

padenot:
Thanks!

PR #379, opened 2017-11-27
Replace instances of 'S' with 'RC' in the SMPTE channel layout comment.

SingingTree:
Looking at the table S appears to be undefined. I believe it is intended to
represent the same thing as 'RC': rear center. This updates the comment to
reflect this.


kinetiknz:
Thanks! Pushed as 2c25a4eb32cbdb5a3a5573acea4a66a41de1f5c8 with a small alignment fix.

PR #380, opened 2017-11-27
Update header sample code to rename several instances of output_params -> input_params.

SingingTree:
The sample code in the header looks to have a few instances where
'output_params' is used where 'input_params' looks intended. This renames
those cases.


kinetiknz:
Thanks!

PR #385, opened 2017-12-04
wasapi: return early on draining for duplex callback

achronop:
Return early from `refill_callback_duplex` when `stm->draining` is `true`

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -550,6 +550,7 @@ long
 refill(cubeb_stream * stm, void * input_buffer, long input_frames_count,
        void * output_buffer, long output_frames_needed)
 {
+  XASSERT(!stm->draining);
   /* If we need to upmix after resampling, resample into the mix buffer to
      avoid a copy. */
   void * dest = nullptr;
@@ -756,6 +757,10 @@ refill_callback_duplex(cubeb_stream * stm)
     return true;
   }
 
+  /* Wait for draining is not important on duplex. */
+  if (stm->draining) {
+    return false;
+  }
 
   ALOGV("Duplex callback: input frames: %Iu, output frames: %Iu",
         input_frames, output_frames);


src/cubeb_wasapi.cpp
```diff
@@ -750,13 +751,12 @@ refill_callback_duplex(cubeb_stream * stm)
     return rv;
   }
 
-  /* This can only happen when debugging, and having breakpoints set in the
+  /* Output frames can only be 0 when debugging, and having breakpoints set in the
    * callback in a way that it makes the stream underrun. */
-  if (output_frames == 0) {
+  if (stm->draining || output_frames == 0) {
```
padenot: Maybe we could even return false here? Draining does not really matter for the duplex case. What do you think ?

achronop: Testing the branch I get the following logs:
```
C:\Users\achronop\repos\cubeb_projest\cubeb\src\cubeb_wasapi.cpp:580: start draining.
C:\Users\achronop\repos\cubeb_projest\cubeb\src\cubeb_log.cpp:80: Duplex callback: input frames: 882, output frames: 960
C:\Users\achronop\repos\cubeb_projest\cubeb\src\cubeb_wasapi.cpp:706: Draining.
C:\Users\achronop\repos\cubeb_projest\cubeb\src\cubeb_wasapi.cpp:702: Draining finished.
stream drained
C:\Users\achronop\repos\cubeb_projest\cubeb\src\cubeb_wasapi.cpp:1190: Stop and join render thread.
C:\Users\achronop\repos\cubeb_projest\cubeb\src\cubeb_wasapi.cpp:1233: Closing thread.
C:\Users\achronop\repos\cubeb_projest\cubeb\src\cubeb_wasapi.cpp:1190: Stop and join render thread.
C:\Users\achronop\repos\cubeb_projest\cubeb\src\cubeb_wasapi.cpp:1192: No thread present.
```

if you note the sequence of the following 2 lines:
```
C:\Users\achronop\repos\cubeb_projest\cubeb\src\cubeb_wasapi.cpp:706: Draining.
C:\Users\achronop\repos\cubeb_projest\cubeb\src\cubeb_wasapi.cpp:702: Draining finished.
```
That's indicates that drain is finishing the second time that enters the `get_output_buffer()` which means it takes 2 extra iterations. If we return false we stop the render loop at 1st iteration. 

> Draining does not really matter for the duplex case

I am not sure what you mean by no matters, in Gecko we have draining on duplex stream when we do `applyConstraints()` to change the input channel count. Let me know if you mean something else.

padenot:
Draining is only useful for output only streams, to makes sure we've played everything that has been sent to the OS, so that the end of, say, a track, is not cut off.

This does not apply for duplex, because we're not playing files. What control the life-time of the stream is not a duration per se, or something that has to be played in its entirety. It's more like a user action, like, for example in Firefox, closing a tab, or calling `AudioContext.close()` or something.

achronop:
Cool, code has been updated, let me know if it's ready for merge.

src/cubeb_wasapi.cpp
```diff
@@ -756,6 +757,10 @@ refill_callback_duplex(cubeb_stream * stm)
     return true;
   }
 
+  /* Wait for drainign is not important on duplex. */
```
padenot: typo.

PR #384, opened 2017-11-29
WASAPI: handle KSAUDIO_SPEAKER_QUAD and KSAUDIO_SPEAKER_5POINT1 channel mapping

SingingTree:
Band aid WASAPI channel mappings so that KSAUDIO_SPEAKER_QUAD and KSAUDIO_SPEAKER_5POINT1 are handled rather than returning undefined. Addresses https://github.com/kinetiknz/cubeb/issues/383.

PR #387, opened 2017-12-05
Add loopback pref and an impl for the WASAPI backend

SingingTree:
Expose a new preference member on `cubeb_stream_params`. Allow callers to specify preference for a loopback device for a (input) stream. Implement loopback devices for the WASAPI backend. Add a number of tests.

Note: Locally, I had some issues with channel layout that I remedied via: https://github.com/kinetiknz/cubeb/pull/384. This PR does not require these changes, but if you encounter problems it may be worth checking if they're related.

The following are outstanding questions that should be discussed (and the PR updated):
- [x] How do we handle loopback requests for other backends (that don't support loopback at this time)? Right now the pref will be ignored, I would prefer that we return an explicit error, but wanted to discuss before I go changing other backends.
- [x] If an input only stream with a loopback device is requested, the same device will be opened in output to drive the loopback: you will get a duplex stream. Callers can interact with this output stream as if they passed in output_params requesting it (these are cloned from the input_params), but it's not intuitive to request a input stream and get a duplex. How to document this for just the WASAPI backend? Should further code be added so the output side of the stream is not exposed to the caller? (Expansion as to why an output stream is opened: loopback devices [can't be event driven](https://msdn.microsoft.com/en-us/library/windows/desktop/dd316551(v=vs.85).aspx). Driving the loopback from event from the same output device is more reliable than using a separate timer, specifically for handling periods of silence, which are [not recorded in the timer based approach](https://blogs.msdn.microsoft.com/matthew_van_eerde/2008/12/16/sample-wasapi-loopback-capture-record-what-you-hear/).
- [x] I'm ignoring AUDCLNT_BUFFERFLAGS_DATA_DISCONTINUITY in `get_input_buffer`. This was already the case, and I believe current design of driving the loop of the output device will mean that these discontinuities (nothing playing) are accounted for by pushing silence into the buffer. I think it would be worthwhile to add explicit comments to this effect, but would value feedback.
- [x] Are the comments and new tests sufficient documentation?
- [ ] Edit: Is API surface change due to addition of .prefs to cubeb_stream_params acceptable?

Please let me know if you have any queries or feedback.

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -229,7 +229,7 @@ typedef struct {
   uint32_t rate;                /**< Requested sample rate.  Valid range is [1000, 192000]. */
   uint32_t channels;            /**< Requested channel count.  Valid range is [1, 8]. */
   cubeb_channel_layout layout;  /**< Requested channel layout. This must be consistent with the provided channels. */
-  cubeb_stream_prefs prefs;      /**< Requested preferences. */
+  cubeb_stream_prefs prefs;     /**< Requested preferences. */
 } cubeb_stream_params;
 
 /** Audio device description */
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -53,16 +53,18 @@ extern "C" {
     output_params.format = CUBEB_SAMPLE_FLOAT32NE;
     output_params.rate = rate;
     output_params.channels = 2;
+    output_params.prefs = CUBEB_STREAM_PREF_NONE;
 
     cubeb_stream_params input_params;
     input_params.format = CUBEB_SAMPLE_FLOAT32NE;
     input_params.rate = rate;
     input_params.channels = 1;
+    input_params.prefs = CUBEB_STREAM_PREF_NONE;
 
     cubeb_stream * stm;
     rv = cubeb_stream_init(app_ctx, &stm, "Example Stream 1",
-                           NULL, input_params,
-                           NULL, output_params,
+                           NULL, &input_params,
+                           NULL, &output_params,
                            latency_frames,
                            data_cb, state_cb,
                            NULL);
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -213,6 +213,15 @@ typedef enum {
   CUBEB_LAYOUT_MAX
 } cubeb_channel_layout;
 
+/** Miscellaneous stream preferences. */
+typedef enum {
+  CUBEB_STREAM_PREF_NONE     = 0x00, /**< No stream preferences are requested. */
+  CUBEB_STREAM_PREF_LOOPBACK = 0x01 /**< Request a loopback stream. Should be
+                                         specified on the input params and an
+                                         output device to loopback from should
+                                         be passed in place of an input device. */
+} cubeb_stream_prefs;
+
 /** Stream format initialization parameters. */
 typedef struct {
   cubeb_sample_format format;   /**< Requested sample format.  One of
@@ -220,6 +229,7 @@ typedef struct {
   uint32_t rate;                /**< Requested sample rate.  Valid range is [1000, 192000]. */
   uint32_t channels;            /**< Requested channel count.  Valid range is [1, 8]. */
   cubeb_channel_layout layout;  /**< Requested channel layout. This must be consistent with the provided channels. */
+  cubeb_stream_prefs prefs;      /**< Requested preferences. */
 } cubeb_stream_params;
 
 /** Audio device description */
--- a/src/cubeb_opensl.c
+++ b/src/cubeb_opensl.c
@@ -1291,6 +1291,11 @@ opensl_validate_stream_param(cubeb_stream_params * stream_params)
        (stream_params->channels < 1 || stream_params->channels > 32))) {
     return CUBEB_ERROR_INVALID_FORMAT;
   }
+  if ((stream_params &&
+       (stream_params->prefs & CUBEB_STREAM_PREF_LOOPBACK))) {
+    LOG("Loopback is not supported");
+    return CUBEB_ERROR_NOT_SUPPORTED;
+  }
   return CUBEB_OK;
 }
 
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -612,22 +612,12 @@ refill(cubeb_stream * stm, void * input_buffer, long input_frames_count,
 
 /* This helper grabs all the frames available from a capture client, put them in
  * linear_input_buffer. linear_input_buffer should be cleared before the
- * callback exits. */
+ * callback exits. This helper does not work with exclusive mode streams. */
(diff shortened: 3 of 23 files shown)


SingingTree:
^ @kinetiknz, @ChunMinChang, and please let me know if you think anyone else would be suitable to provide feedback.

test/test_loopback.cpp
```diff
@@ -0,0 +1,270 @@
...
+  ASSERT_EQ(r, CUBEB_OK) << "Error initializing cubeb stream";
+
+  std::unique_ptr<cubeb_stream, decltype(&cubeb_stream_destroy)>
+    cleanup_stream_at_exit(stream, cubeb_stream_destroy);
+
+  cubeb_stream_start(stream);
+  delay(150);
+  cubeb_stream_stop(stream);
+
+  /* lock user data to be extra sure to not race any outstanding callbacks */
```
kinetiknz: If they occur after stop, it's a bug.  But this also keeps TSAN-type tools happy, so it's good to do.

test/test_loopback.cpp
```diff
@@ -0,0 +1,270 @@
...
+  std::vector<double> output_frames;
+  /* track input */
+  std::vector<double> input_frames;
+};
+
+template<typename T>
+long data_cb_loop_duplex(cubeb_stream * stream, void * user, const void * inputbuffer, void * outputbuffer, long nframes)
+{
+  struct user_state_loopback *u = (struct user_state_loopback *)user;
+  T *ib = (T *)inputbuffer;
```
kinetiknz: Pointer-decls and casts should have an extra space, e.g. `T * ib = (T *) inputbuffer`;

kinetiknz:
> How do we handle loopback requests for other backends (that don't support loopback at this time)? Right now the pref will be ignored, I would prefer that we return an explicit error, but wanted to discuss before I go changing other backends.

I think we should return CUBEB_ERROR_NOT_SUPPORTED in the other backends.  If/when they gain support for this functionality, they can remove the error and do the right thing.

> If an input only stream with a loopback device is requested, the same device will be opened in output to drive the loopback: you will get a duplex stream. Callers can interact with this output stream as if they passed in output_params requesting it (these are cloned from the input_params), but it's not intuitive to request a input stream and get a duplex. How to document this for just the WASAPI backend? Should further code be added so the output side of the stream is not exposed to the caller? (Expansion as to why an output stream is opened: loopback devices can't be event driven. Driving the loopback from event from the same output device is more reliable than using a separate timer, specifically for handling periods of silence, which are not recorded in the timer based approach.

It sounds like we should hide the dummy output stream in this case, otherwise we're exposing an implementation detail that might differ on other platforms when they grow this feature.  How much extra work do you think that would be to hide the dummy output stream?

> I'm ignoring AUDCLNT_BUFFERFLAGS_DATA_DISCONTINUITY in get_input_buffer. This was already the case, and I believe current design of driving the loop of the output device will mean that these discontinuities (nothing playing) are accounted for by pushing silence into the buffer. I think it would be worthwhile to add explicit comments to this effect, but would value feedback.

Sounds right.  It'd be good to document the current behaviour.

> Are the comments and new tests sufficient documentation?

I think what you've added is good, thank you.

test/test_loopback.cpp
SingingTree: Cool. I'll update reference-decls also.

test/test_loopback.cpp
SingingTree: Will update comment to indicate that these checks are paranoia and to appease tooling.

src/cubeb_wasapi.cpp
```diff
@@ -1520,10 +1525,20 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
         return CUBEB_ERROR;
       }
     } else {
-      hr = get_default_endpoint(device, direction);
```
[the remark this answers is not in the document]
SingingTree: Sounds good. I've revised this to still branch on is_loopback for the log on error to allow for a specialized error log. Let me know if that's okay (update PR coming shortly).

SingingTree:
PR updated with explicit commit for feedback -- figure it makes it easy to track for now. Let me know if it looks good, and if you want me to fold the changes into their respective commits, or if it's all gonna be squashed anyway or some such.

Will continue on with changes to address:
- Returning CUBEB_ERROR_NOT_SUPPORTED for other backends
- Investigation and (hopefully) impl for hiding dummy output

SingingTree:
Updated with changes to other backends.

kinetiknz:
Looks good 👍

SingingTree:
Updated with changes that hide the output from the data callback. Feedback appreciated. In particular I'd like a second pair of eyes on the interaction with the resamplers.

SingingTree:
Comment added to explain why AUDCLNT_BUFFERFLAGS_DATA_DISCONTINUITY is not explicitly checked.

include/cubeb/cubeb.h
```diff
@@ -213,13 +213,22 @@ typedef enum {
   CUBEB_LAYOUT_MAX
 } cubeb_channel_layout;
 
+typedef enum {
+  CUBEB_STREAM_PREF_NONE     = 0x00,
```
kinetiknz: Please document this flag and the struct; doxygen is warning that the struct is undocumented.

src/cubeb_opensl.c
```diff
@@ -1291,6 +1291,11 @@ opensl_validate_stream_param(cubeb_stream_params * stream_params)
        (stream_params->channels < 1 || stream_params->channels > 32))) {
     return CUBEB_ERROR_INVALID_FORMAT;
   }
+  if((stream_params &&
```
kinetiknz: Space between if and (

src/cubeb_wasapi.cpp
```diff
@@ -564,9 +568,9 @@ refill(cubeb_stream * stm, void * input_buffer, long input_frames_count,
 {
   XASSERT(!stm->draining);
   /* If we need to upmix after resampling, resample into the mix buffer to
-     avoid a copy. */
+     avoid a copy. Avoid exposing output if it is a dummy stream.*/
```
kinetiknz: Space between . and *

SingingTree:
Thanks for the feedback.
- Style and comment changes have been rolled into their respective commits (as they're small and easily recognized).
- New commit made to fix tests to make changes easily recognizable. Also fix documentation at top of cubeb.h to show prefs being set.

As evidenced by the tests, I'd been thinking incorrectly of how stream_params is initialized and had not considered that this would change the API surface. Depending on how the API is being used, callers will now need to explicitly set stream_params.prefs. Is that acceptable?

kinetiknz:
> As evidenced by the tests, I'd been thinking incorrectly of how stream_params is initialized and had not considered that this would change the API surface. Depending on how the API is being used, callers will now need to explicitly set stream_params.prefs. Is that acceptable?

Yeah, it's fine.  We don't have a stable API right now.

kinetiknz:
Great stuff, thanks!

PR #388, opened 2017-12-06
Fix issue 378: Pass through as more data as possible when upmixing

ChunMinChang:
In the current implementation, only stereo channel will have data. Those channels beyond stereo will be silenced. We should pass through as more data as possible.


kinetiknz:
Thanks!

PR #389, opened 2017-12-07
audiounit: Change AD nominal rate only when is needed

achronop:
Updating device nominal rate for aggregate device (AD) updates the sampling rate of the devices that AD is consisting of. If the device does not support the given rate the update fails. This patch does 2 things: (a) does not attempt to change the nominal rate of an AD if it is the same with output device rate (output is the master device so the rate will set to that anyway) (b) handle the update failure as non fatal failure.

Update of nominal sample rate for AD used in order to fix a dysfunction on Airpods device but the way that it works makes me wonder if we should ever attempt to change it. I push that fix in order to avoid some known failures and I will continue investigating a better way to approach the problem from the beginning.

cc @djg

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1752,8 +1752,12 @@ audiounit_create_aggregate_device(cubeb_stream * stm)
     return  CUBEB_ERROR;
   }
 
-  if (input_nominal_rate != output_nominal_rate) {
-    Float64 rate = std::min(input_nominal_rate, output_nominal_rate);
+  /* Do not attempt to change the nominal rate if it's the same with the nominal rate of
+   * output device because it's the master device and rate will set to that anyway. */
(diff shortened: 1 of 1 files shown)


djg:
Thanks for the changes. I will get to them in the morning.

src/cubeb_audiounit.cpp
```diff
@@ -1752,8 +1752,11 @@ audiounit_create_aggregate_device(cubeb_stream * stm)
     return  CUBEB_ERROR;
   }
 
-  if (input_nominal_rate != output_nominal_rate) {
-    Float64 rate = std::min(input_nominal_rate, output_nominal_rate);
+  // Ouput device is the default so AD will be configured according to that.
+  uint32_t min_rate = std::min(input_nominal_rate, output_nominal_rate);
```
padenot: This needs more info, it's weird. For example, the text you've written for this PR.

Issue #390, opened 2017-12-10
cubeb_destroy never called by firefox


zamaudio:
I realise this could be a gecko bug, but it seems essential to the audio cleanup.
The below patch forces nsLayoutStatic::Shutdown() to be called when the nsLayoutModule shuts down regardless of existing sLayoutStaticRefcnt reference counter value (which is not staying in sync).

 ```
diff --git a/layout/build/nsLayoutModule.cpp b/layout/build/nsLayoutModule.cpp
index 9e85155fdad9..db63a5036a84 100644
--- a/layout/build/nsLayoutModule.cpp
+++ b/layout/build/nsLayoutModule.cpp
@@ -335,7 +335,7 @@ Shutdown()
 
   gInitialized = false;
 
-  nsLayoutStatics::Release();
+  nsLayoutStatics::Shutdown();
 }
 
 #ifdef DEBUG
diff --git a/layout/build/nsLayoutStatics.cpp b/layout/build/nsLayoutStatics.cpp
index abc7985d2c55..6ea6bc08588b 100644
--- a/layout/build/nsLayoutStatics.cpp
+++ b/layout/build/nsLayoutStatics.cpp
@@ -138,12 +138,7 @@ nsrefcnt nsLayoutStatics::sLayoutStaticRefcnt = 0;
 nsresult
 nsLayoutStatics::Initialize()
 {
-  NS_ASSERTION(sLayoutStaticRefcnt == 0,
-               "nsLayoutStatics isn't zero!");
-
-  sLayoutStaticRefcnt = 1;
-  NS_LOG_ADDREF(&sLayoutStaticRefcnt, sLayoutStaticRefcnt,
-                "nsLayoutStatics", 1);
+  nsLayoutStatics::AddRef();
 
   nsresult rv;
 
diff --git a/layout/build/nsLayoutStatics.h b/layout/build/nsLayoutStatics.h
index f859f2d3245d..b6c8f33bacdc 100644
--- a/layout/build/nsLayoutStatics.h
+++ b/layout/build/nsLayoutStatics.h
@@ -23,6 +23,8 @@ public:
   // internally.
   static nsresult Initialize();
 
+  static void Shutdown();
+
   static void AddRef()
   {
     NS_ASSERTION(NS_IsMainThread(),
@@ -52,8 +54,6 @@ private:
   // not to be called!
   nsLayoutStatics();
 
-  static void Shutdown();
-
   static nsrefcnt sLayoutStaticRefcnt;
 };

[...]

PR #391, opened 2017-12-11
cubeb_jack: Replace all mutexes with atomics

### LINKED ISSUE #154: cubeb makes JACK stuck at Firefox's shutdown (and sometimes at runtime too)
I've been having this glitch from the time cubeb got its first JACK code to the latest code that I've shoved into Firefox 49. It gives out xruns on Firefox's shutdown and rarely makes its HTML5 player stuck even before that. The longer Firefox session and more youtube videos played during it - the worse it is. With synchronous mode (`-S`) that allows lower jackd latency with risk of one client disrupting all sound output I get things like that:
```
20:01:37.209 redirection-to-jamin: jamin:a.master.out_L -> system:playback_1 checked.
20:01:37.218 redirection-to-jamin: jamin:a.master.out_R -> system:playback_2 checked.
20:01:37.229 redirection-to-jamin: CubebUtils:AudioStream_0_out_0 -> jamin:in_L checked.
20:01:37.239 redirection-to-jamin: CubebUtils:AudioStream_0_out_1 -> jamin:in_R checked.
20:01:37.249 redirection-to-jamin: PulseAudio JACK Sink:front-left -> jamin:in_L checked.
20:01:37.258 redirection-to-jamin: PulseAudio JACK Sink:front-right -> jamin:in_R checked.
JackPosixSemaphore::TimedWait err = Connection timed out
JackFreewheelDriver::ProcessSync: SuspendRefNum error
JackAudioDriver::ProcessGraphSync: ProcessWriteSlaves error, engine may now behave abnormally!!
JackPosixProcessSync::LockedTimedWait error usec = 5120000 err = Connection timed out
JackEngine::ClientDeactivate wait error ref = 6 name = CubebUtils
JackEngine::ClientKill ref = 6 cannot be removed from the graph !!
20:01:42.271 JACK connection graph change.
20:01:42.330 JACK active patchbay scan...
20:01:42.340 redirection-to-jamin: jamin:a.master.out_L -> system:playback_1 checked.
20:01:42.350 redirection-to-jamin: jamin:a.master.out_R -> system:playback_2 checked.
20:01:42.361 redirection-to-jamin: PulseAudio JACK Sink:front-left -> jamin:in_L checked.
20:01:42.371 redirection-to-jamin: PulseAudio JACK Sink:front-right -> jamin:in_R checked.
20:01:42.604 JACK connection change.
20:01:42.815 JACK active patchbay scan...
20:01:42.825 redirection-to-jamin: jamin:a.master.out_L -> system:playback_1 checked.
20:01:42.835 redirection-to-jamin: jamin:a.master.out_R -> system:playback_2 checked.
20:01:42.846 redirection-to-jamin: PulseAudio JACK Sink:front-left -> jamin:in_L checked.
20:01:42.856 redirection-to-jamin: PulseAudio JACK Sink:front-right -> jamin:in_R checked.
JackPosixProcessSync::LockedTimedWait error usec = 1024000 err = Connection timed out
JackEngine::ClientCloseAux wait error ref = 6
JackPosixSemaphore::TimedWait err = Connection timed out
JackAudioDriver::ProcessGraphSync: SuspendRefNum error, engine may now behave abnormally!!
20:01:47.272 XRUN callback (1).
```

That makes me seriously think about disabling cubeb's native JACK support for good and use PA as a compatibility shim. It seems it doesn't close its connections properly or something.

zamaudio:
This patch removes all pthread mutexes in the jack backend, especially that ugly trylock in the realtime audio thread.  These are replaced with atomic variables.  Passes all tests, although JACK xruns when `test_sanity` is being called in the `cubeb.stream_position` test where there are massive 500ms delays.

Users note that there is still a bug with Firefox itself that causes cubeb never to shutdown (#390).  This can cause xruns on firefox quit until it is fixed.  However, this patch should see an improvement over the previous backend code.  Partially fixes #154

### DIFF
--- a/src/cubeb_jack.cpp
+++ b/src/cubeb_jack.cpp
@@ -28,6 +28,7 @@
 
 #define JACK_API_VISIT(X)                       \
   X(jack_activate)                              \
+  X(jack_deactivate)                            \
   X(jack_client_close)                          \
   X(jack_client_open)                           \
   X(jack_connect)                               \
@@ -141,10 +142,9 @@ static struct cubeb_ops const cbjack_ops = {
 struct cubeb_stream {
   cubeb * context;
 
-  /**< Mutex for each stream */
-  pthread_mutex_t mutex;
-
-  bool in_use; /**< Set to false iff the stream is free */
+  bool taken; /**< Atomic var: grabs the stream immediately */
+  bool in_use; /**< Atomic var: Set to false iff the stream is free */
+  bool active;
   bool ports_ready; /**< Set to true iff the JACK ports are ready */
 
   cubeb_data_callback data_callback;
@@ -169,9 +169,6 @@ struct cubeb {
   struct cubeb_ops const * ops;
   void * libjack;
 
-  /**< Mutex for whole context */
-  pthread_mutex_t mutex;
-
   /**< Audio buffers, converted to float */
   float in_float_interleaved_buffer[FIFO_SIZE * MAX_CHANNELS];
   float out_float_interleaved_buffer[FIFO_SIZE * MAX_CHANNELS];
@@ -188,6 +185,7 @@ struct cubeb {
   cubeb_device_collection_changed_callback collection_changed_callback;
 
   bool active;
+  bool in_use;
   unsigned int jack_sample_rate;
   unsigned int jack_latency;
   unsigned int jack_xruns;
@@ -288,11 +286,15 @@ cbjack_graph_order_callback(void * arg)
   int i;
   jack_latency_range_t latency_range;
   jack_nframes_t port_latency, max_latency = 0;
+  cubeb_stream *stm = NULL;
+
+  if (!ctx->active)
+    return 0;
 
   for (int j = 0; j < MAX_STREAMS; j++) {
-    cubeb_stream *stm = &ctx->streams[j];
+    stm = &ctx->streams[j];
 
-    if (!stm->in_use)
+    if (!stm->active)
       continue;
     if (!stm->ports_ready)
       continue;
@@ -308,7 +310,7 @@ cbjack_graph_order_callback(void * arg)
       max_latency = 128;
   }
 
-  ctx->jack_latency = max_latency;
+  ctx->jack_latency = max_latency * 1000 / ctx->jack_sample_rate;
 
   return 0;
 }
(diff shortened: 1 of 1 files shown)


jyavenard:
FWIW: making all variables atomic isn't an equivalent to the use of a single mutex.
It no longer appears to keep consistency between all group of variables.

zamaudio:
> "making all variables atomic isn't an equivalent to the use of a single mutex."
I do not make *all* variables atomic.

> "It no longer appears to keep consistency between all group of variables."
Can you please elaborate, the variables I made atomic are used in such a way that only one thread is allowed to *write* to one atomic variable and the next time the other thread executes it *reads* the value so that two threads are never writing to the same variable and getting out of sync.

kinetiknz:
What concrete problem does this change solve?  The review and maintenance burden of verifying attempts at lock free implementations likely outweighs any benefit we'll ever see from this change, in my opinion.  Sorry, but I don't think we will be merging this without an incredibly strong reason to do so.

szanni:
> What concrete problem does this change solve?

It is trying to solve a design problem. The current code is broken. Using mutexes in real time context is a big no. No audio library I am aware of does this. All use lockless buffers implemented through atomics. Some real time audio rules [here](http://www.rossbencina.com/code/real-time-audio-programming-101-time-waits-for-nothing).

The reason the current code works is because the code uses `mutex_trylock`, so it will not block. The buffer is instead filled with silence when the `mutex_trylock` fails. _How can that be considered correct behaviour?_ Especially since it can be prevented and jack is geared towards audio professionals.

That being said, the current code does seem to work on most machines without hiccups. Or I could at least not find any bug reports about audio glitches. And I understand the effort involved in verifying lockless algorithms. I have written code like this myself. It is a pain. And we do seem to get lucky most of the time as mutex contentions is low. I added some debug symbols. So far I have not been able to hit the _insert silence_ code path. I do wonder how many people do hit that code path from time to time on other hardware and less performant systems though. I am commenting here as I got concerned about the `mutex` calls myself while reading through the current code relating to another issue.

All that being said, this PR is beyond broken. Line 145 of the PR tells me the submitter has no idea how atomics work. **A bool is not an atomic**. Looking at the rest of the code: it does not seem to solve anything either. It is just the attempt at implementing a mutex through atomics and does not solve anything!? You would need a lockless ring buffer with a reader and a writer thread or something similar.

zamaudio:
@szanni You're completely right.  The current code is broken because it has trylocks in the realtime thread.  This PR was an attempt to fix it, but as you described, it uses booleans in place of a mutex instead of a lockless ring buffer mechanism.  I urge anyone who knows how to fix this to do so, and I will close this PR soon.  I don't know how to fix it properly and I also don't have time to spend on it further.

szanni:
I do in theory know how to fix this, as I have written push & callback style back ends for alsa, coreaudio, sdl, &&
But this looks like a bigger time sink. Understanding the cubeb API and then rewriting the whole jack back end entirely. Because looking at the current state of the code I have zero faith in other stuff not being broken too and making incorrect assumptions.  I mean this is how I got here in the first place. The current implementation crashes hard on xruns trying to access memory that is out of bounds. That's why I opened #607, @zamaudio you actually seem to have fixed the subtraction logic in this PR as a side effect as well.

If somebody wants to open a bug bounty, I might find the time to write a proper jack back end. I would however recommend just switching Firefox to SDL2.0 as that is battle tested, supports pretty much all platforms imaginable, and is released under the very permissive zlib license. I am not a fan of the SDL API, but at least the library is not broken and the developers seem understand how audio works.

@zamaudio maybe close this PR and open up an issue instead lamenting the mutex stuff and linking to this thread? Or I could do that too. And I hope my language wasn't too strong criticising your PR :)

PR #392, opened 2017-12-11
wasapi: increase initial silence on debug

achronop:
Increase initial silence on debug for wasapi to avoid intermittent described by bug 1393740


kinetiknz:
Thanks!

PR #396, opened 2017-12-27
pulse: Don't read pa_sink_port_info.available when runtime pulseaudio…

glandium:
… is < 2.0

The code reading that field is build-time protected, so as long as the
runtime pulseaudio version is greater than the one used at build-time,
we're good. And as of now, the build-time version for Firefox is
0.9.something, from our Centos 6 build environment.

In bugzilla bug 1399679, we're going to switch the build environment to Debian
Wheezy, which comes with pulseaudio 2.0 headers, enabling that code.
This means running the official Firefox on e.g. Ubuntu 12.04, which has
1.1, would read past the end of the pa_sink_port_info, possibly leading
to random behavior with port availability.

https://bugzilla.mozilla.org/show_bug.cgi?id=1427150

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -84,10 +84,15 @@
   X(pa_context_set_subscribe_callback)          \
   X(pa_context_subscribe)                       \
   X(pa_mainloop_api_once)                       \
+  X(pa_get_library_version)                     \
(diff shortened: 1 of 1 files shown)


djg:
Thanks @glandium

PR #397, opened 2018-01-09
audiounit: skip mixing when mixer layout is undefined

achronop:
When mixer layout is undefined we assert because the number of channel is zero. This patch skip mixing if the mixer layout is set to undefined.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -182,6 +182,7 @@ struct cubeb_stream {
   AudioObjectID plugin_id = 0;              // used to create aggregate device
   /* Mixer interface */
   std::unique_ptr<cubeb_mixer, decltype(&cubeb_mixer_destroy)> mixer;
+  cubeb_stream_params mixer_params = { CUBEB_SAMPLE_FLOAT32NE, 0, 0, CUBEB_LAYOUT_UNDEFINED };
 };
 
 bool has_input(cubeb_stream * stm)
@@ -409,19 +410,14 @@ audiounit_mix_output_buffer(cubeb_stream * stm,
                             void * output_buffer,
                             unsigned long output_buffer_length)
 {
-  cubeb_stream_params output_mixer_params = {
-    stm->output_stream_params.format,
-    stm->output_stream_params.rate,
-    CUBEB_CHANNEL_LAYOUT_MAPS[stm->context->layout].channels,
-    stm->context->layout
-  };
-
+  assert(stm->output_stream_params.layout != CUBEB_LAYOUT_UNDEFINED);
+  assert(stm->mixer_params.layout != CUBEB_LAYOUT_UNDEFINED);
   // The downmixing(from 5.1) supports in-place conversion, so we can use
   // the same buffer for both input and output of the mixer.
   cubeb_mixer_mix(stm->mixer.get(), output_frames,
                   output_buffer, output_buffer_length,
                   output_buffer, output_buffer_length,
-                  &stm->output_stream_params, &output_mixer_params);
+                  &stm->output_stream_params, &stm->mixer_params);
 }
 
 static OSStatus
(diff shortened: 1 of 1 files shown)


ChunMinChang:
Thanks for taking this issue!

We should call ```downmix_fallback``` for the device with undefined layout but its channel count is not zero. However, this won't work if we early return when ```stm->context->layout == CUBEB_LAYOUT_UNDEFINED``` is true in ```audiounit_init_mixer```. Could we check the channel count instead?

achronop:
> its channel count is not zero.
I am a little confused about which channel count you mean by that. At this point the device has been configured with the number of channels set by the user during `stream_init`. Should I query again the device and get a different number of channel? If yes do you know what is the right property to use?

ChunMinChang:
The channel count I mean is ```mixer_params->channels``` here.

The crash of [Bug 1384116](https://bugzilla.mozilla.org/show_bug.cgi?id=1384116) is
> Assertion failed: (stream_params->channels >= mixer_params->channels && mixer_params->channels > 0), function cubeb_downmix, file /Users/agaynor/projects/mozilla-central/media/libcubeb/src/cubeb_mixer.cpp, line 383

Your [analysis](https://bugzilla.mozilla.org/show_bug.cgi?id=1384116#c13) shows that the user who use an **undefined** layout will hit this the [assertion in cubeb_downmix](https://github.com/kinetiknz/cubeb/blob/bda37c26b0ed46c568e35b10728fc498262778f3/src/cubeb_mixer.cpp#L382).

> At this point the device has been configured with the number of channels set by the user during stream_init

At that time, the channel count is 0 since we use ```CUBEB_CHANNEL_LAYOUT_MAPS``` in
[audiounit_mix_output_buffer](https://github.com/kinetiknz/cubeb/blob/bda37c26b0ed46c568e35b10728fc498262778f3/src/cubeb_audiounit.cpp#L415) to guess the number of the channels. That is, some 2-channel devices configured to non-SMPTE layout(user-defined layout) such as <Right, Left> or <Left_Surround, Right_Surround> will lead to crashes since its ```mixer_params->channels``` will be 0 in [audiounit_mix_output_buffer](https://github.com/kinetiknz/cubeb/blob/bda37c26b0ed46c568e35b10728fc498262778f3/src/cubeb_audiounit.cpp#L415).

> Should I query again the device and get a different number of channel?

Yes, we need to get the actual number of channel of the device.
We should replace [```CUBEB_CHANNEL_LAYOUT_MAPS[stm->context->layout].channels```](https://github.com/kinetiknz/cubeb/blob/bda37c26b0ed46c568e35b10728fc498262778f3/src/cubeb_audiounit.cpp#L415) by ```audiounit_get_channel_count``` or ```audiounit_get_max_channel_count``` to allow the downmixing for undefined layout. The [downmix_fallback](https://github.com/kinetiknz/cubeb/blob/bda37c26b0ed46c568e35b10728fc498262778f3/src/cubeb_mixer.cpp#L352) should work in that case. It pass through as much channel data as it can, regardless of what layout is.

achronop:
> Yes, we need to get the actual number of channel of the device.
> We should replace CUBEB_CHANNEL_LAYOUT_MAPS[stm->context->layout].channels by audiounit_get_channel_count or audiounit_get_max_channel_count to allow the downmixing for undefined layout. The downmix_fallback should work in that case. It pass through as much channel data as it can, regardless of what layout is.

Then the question is what happens if the `audiounit_get_channel_count` (or max) return number of channels greater than `stream_params->channels`. In this case we do not get the downmix direction. Using directly `downmix_fallback` in that case also does not help. What can we do in that case?

ChunMinChang:
I recall that OSX will do the same thing as ```downmix_fallback``` without any setting:
https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_mixer.cpp#L226-L249
so my above concern may be ignored. I guess OSX will do the same way in upmixing, so maybe it's ok if we just do nothing when the layout is undefined.

achronop:
In that case could you review the PR?

padenot:
I've encountered something similar yesterday (see #419). Unclear if this one still applies.

achronop:
Since as we use mixer this is a valid patch. It would be nice to get review comments.

djg:
Is this PR still valid after the mixer re-write?

jyavenard:
no... this PR should be closed.

PR #398, opened 2018-01-11
cubeb_resampler_speex: don't call data callback while draining

karlt:


karlt:
@padenot, kinetik and I thought it would be best to get you to have a look at this.

padenot:
Thanks!

PR #399, opened 2018-01-12
Replace free by delete to prevent potential memory leak 

ChunMinChang:
The object allocated by ```new``` should be deallocated by ```delete``` instead of ```free```.


kinetiknz:
Nice catch!

PR #400, opened 2018-01-15
Dont enumerate cubeb's private aggregate devices on OSX

padenot:


src/cubeb_audiounit.cpp
```diff
@@ -3181,6 +3181,12 @@ audiounit_create_device_from_hwdev(cubeb_device_info * ret, AudioObjectID devid,
   return CUBEB_OK;
 }
 
+bool
+is_aggregate_device(cubeb_device_info * device_info)
+{
+  return !!strstr(device_info->friendly_name, "CubebAggregateDevice");
```
achronop: A small comment about `strstr` is that finds any occurrence of the given string. For example name like "xxxxxCubebAggregateDevicexxxx" will mach. You can guard it more with strncmp: `strncmp(str,"CubebAggregateDevice", 20) == 0`

PR #401, opened 2018-01-15
Fix deadlock and device collection changed callback issue on OSX 

padenot:
- Fix deadlock when creating a duplex stream when a device collection change callback was installed
- Stop calling this callback when creating a duplex stream (the fact that we create a device internally should not be exposed), and add a test for it.
- sed 's/std:://' since I added `using namespace std;`.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -32,8 +32,11 @@
 #include <algorithm>
 #include <atomic>
 #include <vector>
+#include <set>
 #include <sys/time.h>
 #include <string>
+
+using namespace std;
 
 #if MAC_OS_X_VERSION_MIN_REQUIRED < 101000
 typedef UInt32 AudioFormatFlags;
@@ -3321,10 +3324,39 @@ audiounit_collection_changed_callback(AudioObjectID /* inObjectID */,
       if (context->devtype_device_array == devices) {
         /* Device changed for the other scope, ignore. */
         return;
+      } else {
+        /* Also don't trigger the user callback if the new added device is private
+         * aggregate device: compute the set of new devices, and remove those
+         * with the name of our private aggregate devices. */
+        set<AudioObjectID> current_devices(devices.begin(), devices.end());
+        set<AudioObjectID> previous_devices(context->devtype_device_array.begin(),
+                                            context->devtype_device_array.end());
+        set<AudioObjectID> new_devices;
+        set_difference(current_devices.begin(), current_devices.end(),
+                       previous_devices.begin(), previous_devices.end(),
+                       inserter(new_devices, new_devices.begin()));
+
+        for (auto it = new_devices.begin(); it != new_devices.end();) {
+          CFStringRef name = get_device_name(*it);
+          if (CFStringFind(name, CFSTR("CubebAggregateDevice"), 0).location !=
+              kCFNotFound) {
+            it = new_devices.erase(it++);
+          } else {
+            it++;
+          }
+        }
+
+        // If this set of new devices is empty, it means this was triggerd
+        // solely by creating an aggregate device, no need to trigger the user
+        // callback.
+        if (new_devices.empty()) {
+          return;
+        }
       }
       /* Device on desired scope changed. */
       context->devtype_device_array = devices;
     }
+
     context->collection_changed_callback(context, context->collection_changed_user_ptr);
   });
   return noErr;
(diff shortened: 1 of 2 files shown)


achronop:
Can you please describe the scenario that leads to the deadlock solved by the second commit?

padenot:
It's in the commit message.

src/cubeb_audiounit.cpp
```diff
@@ -3301,27 +3307,58 @@ audiounit_collection_changed_callback(AudioObjectID /* inObjectID */,
...
+
+    /* Differentiate input from output changes. */
+    if (context->collection_changed_devtype == CUBEB_DEVICE_TYPE_INPUT ||
+        context->collection_changed_devtype == CUBEB_DEVICE_TYPE_OUTPUT) {
+      vector<AudioObjectID> devices = audiounit_get_devices_of_type(context->collection_changed_devtype);
+      /* When count is the same examine the devid for the case of coalescing. */
+      if (context->devtype_device_array == devices) {
+        /* Device changed for the other scope, ignore. */
+        return;
+      } else {
```
achronop: You can do the following calculation when `current_devices.size() > previous_devices.size()` in order to avoid extra computations when a device is removed.

src/cubeb_audiounit.cpp
```diff
@@ -3301,27 +3307,58 @@ audiounit_collection_changed_callback(AudioObjectID /* inObjectID */,
...
+        set<AudioObjectID> new_devices;
+        set_difference(current_devices.begin(), current_devices.end(),
+                       previous_devices.begin(), previous_devices.end(),
+                       inserter(new_devices, new_devices.begin()));
+
+        for (auto it = new_devices.begin(); it != new_devices.end();) {
+          CFStringRef name = get_device_name(*it);
+          if (CFStringFind(name, CFSTR("CubebAggregateDevice"), 0).location !=
+              kCFNotFound) {
+            it = new_devices.erase(it++);
```
achronop: Nit: since you set `it` from `erase()` you do not have to add it. I think adding was the pre-c++11 technique when `erase()` was void.

src/cubeb_audiounit.cpp
```diff
@@ -2662,8 +2665,10 @@ audiounit_stream_destroy(cubeb_stream * stm)
     LOG("(%p) Could not uninstall all device change listeners", stm);
   }
 
-  auto_lock context_lock(stm->context->mutex);
-  audiounit_stream_stop_internal(stm);
+  {
+    auto_lock context_lock(stm->context->mutex);
+    audiounit_stream_stop_internal(stm);
+  }
```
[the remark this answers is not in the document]
padenot: when arriving at the `dispatch_sync` below, without the code above, we have acquired the context mutex.

Now, in between ` audiounit_stream_stop_internal` and trying to do the `dispatch_sync`, maybe we're executing the `dispatch_async`, that tries to get the context mutex as well (but on another thread).

Since the `dispatch_sync` is, well, sync, then we can't progress, the queue is already executing a job, that is blocked on the context mutex.

We don't need to take the context mutex to delete a stream, so it's fine.

src/cubeb_audiounit.cpp
padenot: What if, during the same callback, we're removing an aggregate device and removing a physical device ? We'll miss an update.

src/cubeb_audiounit.cpp
padenot: Ha yes indeed. Old habits...

src/cubeb_audiounit.cpp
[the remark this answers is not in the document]
achronop: I see thanks.

test/test_duplex.cpp
```diff
@@ -144,7 +144,7 @@ TEST(cubeb, duplex_collection_change)
   ASSERT_EQ(r, CUBEB_OK) << "Error initializing cubeb library";
 
   r = cubeb_register_device_collection_changed(ctx,
-                                               static_cast<cubeb_device_type>(CUBEB_DEVICE_TYPE_INPUT & CUBEB_DEVICE_TYPE_INPUT),
+                                               CUBEB_DEVICE_TYPE_INPUT & CUBEB_DEVICE_TYPE_INPUT,
```
achronop: It's `input & input`. Just one `input` should be fine

src/cubeb_audiounit.cpp
achronop: Do you mean add an aggregate device and remove a physical? If that's the case I agree to your point.

PR #402, opened 2018-01-16
pulse: avoid deadlock when context subscribe fails

achronop:
When `pa_context_subscribe` fails we need to unlock the main loop to avoid infinite blocking.


kinetiknz:
Thanks!

kinetiknz:
From a quick look, this was the only place we had a missing unlock on early return.

Issue #403, opened 2018-01-17
WASAPI backend should not push silence if it fails to setup the input side


SingingTree:
If setup of the input side of a WASAPI stream [fails](https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_wasapi.cpp#L1664) cubeb still attempts to [push_silence()](https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_wasapi.cpp#L1685) on the linear_input_buffer. However, if setup fails then input_buffer_frame_count may not be set correctly. This results in attempting to push silenece to undesirable addresses (0 in my testing).

Should have a PR to address this shortly.

kinetiknz:
Closed by #404.

PR #404, opened 2018-01-17
WASAPI: Check that the input side of a stream was successfully setup before pushing silence

SingingTree:
Address #403 by checking if setup succeeded before pushing silence.


kinetiknz:
Thanks!

PR #409, opened 2018-01-26
Handle null friendly name in cubeb_audiounit.cpp

padenot:


src/cubeb_audiounit.cpp
```diff
@@ -3151,7 +3151,15 @@ audiounit_create_device_from_hwdev(cubeb_device_info * ret, AudioObjectID devid,
...
+    if (str) {
+      ret->friendly_name = audiounit_strref_to_cstr_utf8(str);
+    } else {
+      // Couldn't get a friendly_name, nor a datasource name, return a valid
+      // string of length 0.
+      char * fallback_name = new char[1];
+      fallback_name[0] = '\0';
+      ret->friendly_name = fallback_name;
+    }
     CFRelease(str);
```
achronop: You could move `CFRelease(str);` in the if branch. No point to release when str == null. I think that's the travis error report.

Issue #324, opened 2017-06-05
WASAPI backend doesn't obey latency parameter


ligfx:
I believe this a limitation of WASAPI shared mode: no matter the requested buffer size, callback events still come in at the default device period.

For example, on my computer (Windows 10, Cirrus Logic CS4208 speakers set to 24-bit 48000Hz), the default device period is 10 milliseconds / 480 frames. In `test_audio`, which requests a latency of 4096 frames, the first callback gets to fill the entire buffer (4096 frames), but each additional callback gets a number of frames equivalent to the rate-adjusted device period.

One strategy that seems to work is to ignore WASAPI refill events (e.g. don't call `IAudioRenderClient::GetBuffer`) until the desired number of frames is available to be written to. Of course, since the buffer size is equal to the desired number of frames, this causes audio glitches since the audio engine runs out of data. The default buffer size seems to be twice the period plus some margin, so I tried a buffer that's a little more than twice the requested latency, and that seemed to fix the glitches.

padenot:
Does it still work of you load the callback like crazy? Try putting a `Sleep` or something in it.

For now, we're mostly using the lowest latency all the time (because we're using this for WebRTC and other real-time stuff), but we need to implement higher-latency channels in the future (for battery consumption and CPU usage reasons), so it would be super if we can have it solved and solid.

Let me know if I can help, maybe running some tests on some other sound card/OS combination that you don't have?

ligfx:
If I `Sleep` in the callback, the number of frames available on the next call increases (makes sense to me, the audio engine has had more time to process the buffer), but the audio starts glitching. I guess that does "work," in one sense 😆 

I looked around some more, and it does seem that WASAPI doesn't offer any latency control at all before Windows 8, and only supports latency control for desktop apps starting in Windows 10.

The Chromium project looked at this in the context of reducing power usage (_[Reduce audio power usage across all platforms by using larger buffer sizes](https://bugs.chromium.org/p/chromium/issues/detail?id=362261#c35)_) and solving various audio glitches (_[AV glitches during high CPU load](https://bugs.chromium.org/p/chromium/issues/detail?id=362294#c23)_), and settled for sending large chunks to their audio thread and FIFO'ing them down to the correct WASAPI buffer size (_[Enable higher latency audio on Windows for basic playback](https://codereview.chromium.org/275943003): "The WASAPI callback frequency is fixed irrespective of the buffer size we specify, so the best we can do is reduce the number of IPC round trips."_). 

[...]

PR #415, opened 2018-02-11
[RFC] Implement CUBEB_STREAM_PREF_EXCLUSIVE

spycrab:
A bad attempt at supporting Exclusive mode.

Issues:
* I end up with segfaults when the IAudioClient should be started for some reason.
* Format availability detection is probably still off
* Formatting issues / Debugging leftovers in a few places

Some of the changes are probably not necessary (e.g. the ``CLSCTX_ALL`` stuff).

If someone had some ideas regarding what I'm doing wrong, help would be greatly appreciated!

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -220,11 +220,16 @@ enum {
 
 /** Miscellaneous stream preferences. */
 typedef enum {
-  CUBEB_STREAM_PREF_NONE     = 0x00, /**< No stream preferences are requested. */
-  CUBEB_STREAM_PREF_LOOPBACK = 0x01 /**< Request a loopback stream. Should be
-                                         specified on the input params and an
-                                         output device to loopback from should
-                                         be passed in place of an input device. */
+    CUBEB_STREAM_PREF_NONE     = 0x00, /**< No stream preferences are requested. */
+    CUBEB_STREAM_PREF_LOOPBACK = 0x01, /**< Request a loopback stream. Should be
+                                       specified on the input params and an
+                                       output device to loopback from should
+                                       be passed in place of an input device. */
+    CUBEB_STREAM_PREF_EXCLUSIVE = 0x02 /**< (Windows / WASAPI only) Request that
+                                       this stream is run in exclusive mode
+                                       which results in lower latency but
+                                       doesn't allow other applications to
+                                       use the same device */
 } cubeb_stream_prefs;
 
 /** Stream format initialization parameters. */
--- a/src/cubeb.c
+++ b/src/cubeb.c
@@ -309,6 +309,11 @@ cubeb_stream_init(cubeb * context, cubeb_stream ** stream, char const * stream_n
     return r;
   }
 
+  if ((output_stream_params && output_stream_params->prefs) & CUBEB_STREAM_PREF_EXCLUSIVE || (input_stream_params && input_stream_params->prefs & CUBEB_STREAM_PREF_EXCLUSIVE)) {
+    if (strcmp(cubeb_get_backend_id(context), "wasapi") != 0)
+      return CUBEB_ERROR_NOT_SUPPORTED;
+  }
+
   r = context->ops->stream_init(context, stream, stream_name,
                                 input_device,
                                 input_stream_params,
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -524,9 +524,20 @@ refill(cubeb_stream * stm, void * input_buffer, long input_frames_count,
 
 int wasapi_stream_reset_default_device(cubeb_stream * stm);
 
+/* Helper for making get_input_buffer work in exclusive mode */
+HRESULT get_next_packet_size(cubeb_stream * stm, PUINT32 next)
+{
+    if (stm->input_stream_params.prefs & CUBEB_STREAM_PREF_EXCLUSIVE) {
+        *next = stm->input_buffer_frame_count;
+        return S_OK;
+    } else {
+        return stm->capture_client->GetNextPacketSize(next);
+    }
+}
+
 /* This helper grabs all the frames available from a capture client, put them in
  * linear_input_buffer. linear_input_buffer should be cleared before the
- * callback exits. This helper does not work with exclusive mode streams. */
+ * callback exits. */
 bool get_input_buffer(cubeb_stream * stm)
 {
   XASSERT(has_input(stm));
@@ -543,9 +554,9 @@ bool get_input_buffer(cubeb_stream * stm)
   // single packet each time. However, if we're pulling from the stream we may
   // need to grab multiple packets worth of frames that have accumulated (so
   // need a loop).
-  for (hr = stm->capture_client->GetNextPacketSize(&next);
+  for (hr = get_next_packet_size(stm, &next);
        next > 0;
-       hr = stm->capture_client->GetNextPacketSize(&next)) {
+       hr = get_next_packet_size(stm, &next)) {
     if (hr == AUDCLNT_E_DEVICE_INVALIDATED) {
       // Application can recover from this error. More info
       // https://msdn.microsoft.com/en-us/library/windows/desktop/dd316605(v=vs.85).aspx
@@ -619,7 +630,11 @@ bool get_input_buffer(cubeb_stream * stm)
       LOG("FAILED to release intput buffer");
       return false;
     }
-    offset += input_stream_samples;
+
+	offset += input_stream_samples;
+
+    if (stm->input_stream_params.prefs & CUBEB_STREAM_PREF_EXCLUSIVE)
+      break;
   }
 
   XASSERT(stm->linear_input_buffer->length() >= offset);
@@ -1297,9 +1312,13 @@ wasapi_get_min_latency(cubeb * ctx, cubeb_stream_params params, uint32_t * laten
     return CUBEB_ERROR;
   }
 
-  /* The second parameter is for exclusive mode, that we don't use. */
   REFERENCE_TIME default_period;
-  hr = client->GetDevicePeriod(&default_period, NULL);
+
+  if (params.prefs & CUBEB_STREAM_PREF_EXCLUSIVE)
+      hr = client->GetDevicePeriod(NULL, &default_period);
+  else
+      hr = client->GetDevicePeriod(&default_period, NULL);
+
   if (FAILED(hr)) {
     LOG("Could not get device period: %lx", hr);
     return CUBEB_ERROR;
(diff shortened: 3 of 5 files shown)


spycrab:
This line also causes an access violation for some reason: https://github.com/kinetiknz/cubeb/pull/415/files#diff-2f55f506bd2dd918b7b8c6ceec8aee0dR1671

kinetiknz:
Thanks for the PR.  I think we'd add support for this as long as the changes are non-invasive and a new test is added to ensure the functionality isn't accidentally broken -- we won't use this inside Firefox, so without tests it'll end up broken fairly quickly.

I do wonder whether the user cases are better addressed via IAudioClient3 as discussed in issue #324 (although that's Windows 10 only).  To clarify, is the reason you want this feature for lower latency?

For the PR:
- The crash in LOG at 1671 is possibly due to differences in setting up the remix/resample formats in handle_channel_layout with exclusive mode; that's the piece of code I'd expect more changes around to support exclusive mode.  If you get stuck, report back and we can dig in when we have time.  (this is your point 2, possibly also point 1)
- Following on from the above: does it make sense to allow remix/resampling for exclusive streams?  Maybe cubeb should require the requested format matches the OS/hardware exactly, otherwise stream init fails.
- I'd remove the CLSCTX_ALL changes unless there's a reason for them.
- CUBEB_STREAM_PREF_WASAPI_EXCLUSIVE could be OS-agnostic and return errors from the other backends (until they support something similar, if that functionality makes sense on the platform); that'd allow removing WASAPI from the name and the ifdefs in cubeb.h.
- There's some code that assumes shared mode (e.g. https://github.com/kinetiknz/cubeb/blob/475d97f64567b292761bd9dddaecbd54749b3aa3/src/cubeb_wasapi.cpp#L623) that needs to be updated to work, or at least tested and known to fail in a sensible way.

spycrab:
@kinetiknz Thanks for your quick response!

Changes I made:
* Removed the ```CLSCTX_ALL`` bits for now
* Made the ```LOG`` line default to ``CUBEB_LAYOUT_UNKNOWN`` which fixes the segfault
* Turned the flag into ``CUBEB_STREAM_EXCLUSIVE```and added a backend check to ``cubeb.c``

I took a look at quick look at ``handle_channel_layout`` and it didn't look as if it had to be changed for exclusive mode.

Not sure if ``get_input_buffer`` plays a role at this stage though.

That's all I know at the moment :/

spycrab:
Regarding the tests: I'd suggest just adding them to ``test_audio`` via an ifdef, that's how I've tested it thus far.

kinetiknz:
> Not sure if get_input_buffer plays a role at this stage though.

It's only used on the capture side, not playback.

> Regarding the tests: I'd suggest just adding them to test_audio via an ifdef, that's how I've tested it thus far.

Sounds good.

spycrab:
Fixed ``get_input_buffer`` now. But I'm not sure where to go from here to be honest.
I've already tried overriding ``mix_format`` with a known, working one but to no avail.

kinetiknz:
The indenting is a bit weird in the latest version.

I noticed the LOG change, what are you getting in mix_params->layout?

spycrab:
``mix_params->layout`` is ``CUBEB_LAYOUT_STEREO``.
Regarding paragraph 2, no that probably doesn't make sense but I've not encountered remapping / resampling yet (according to the logs atleast).

kinetiknz:
Sorry, I meant the question about why you want this functionality vs IAudioClient3's lower latency shared mode.

I had a quick look and have a few pointers for where to look next to get this working.  There's some sample code with explananations at https://blogs.msdn.microsoft.com/matthew_van_eerde/2009/04/03/sample-wasapi-exclusive-mode-event-driven-playback-app-including-the-hd-audio-alignment-dance/ that might be useful.

Testing with test_sanity (simplest test; plays a tone - updated to pass the exclusive pref flag), IAudioClient::Initialize fails with AUDCLNT_E_BUFFER_SIZE_ERROR.  Since we're passing 0 for the hnsPeriodicity  (MSDN says this *must* be non-zero for exclusive + event streams), that's probably the first thing to fix.

From the linked blog post above, the other thing we'll likely need to handle is the "alignment dance" to calculate buffer sizing compatible with the harder.

spycrab:
Sorry, for some reason I interpreted paragraph two meaning point two (entirely my fault).
Anyway regarding your question: 
Yeah, the main reason I want exclusive mode support is for lower latency. While ``IAudioClient3`` looks interesting the Windows 10 only part of it kind of kills my interest.
(It'd also be interesting to compare latency between ``IAudioClient`` exclusive mode and ``IAudioClient3`` shared mode).

``hnsPeriodicity`` was set correctly before, I've must have messed this up while trying to fix the formatting (Not used to VS).

We should probably handle this, yes. Although I'll have to find a test case that requires that.

Thanks for your patience.

spycrab:
@kinetiknz Fixed ``hnsPeriodicity``.
Sadly ``test_sanity`` just segfaults where ``test_audio`` does :/

padenot:
We're having a look at `IAudioClient3` as well (independently, as you can imagine, we can't really use exclusive mode for Firefox). I certainly helps a lot when it's available.

spycrab:
When it's available, it's probably superior. If only for the fact that it runs in shared mode

spycrab:
Turns out they crash at different sections now.
Anyhow here's what I could find:

test_audio
========
Caused by:
```
test_audio.exe!`anonymous namespace'::stream_start_one_side(cubeb_stream * stm, `anonymous-namespace'::StreamDirection dir) Line 2081 (In cubeb_wasapi.cpp)
```
Error Message:
```
Exception thrown at 0x6A33E20E (AudioSes.dll) in test_audio.exe: 0xC0000005: Access violation reading location 0x04DF0034. occurred
```
Library Function
```
AudioSes.dll!CAudioClient::Start()
```


test_sanity
========
Caused by:
```
Test_sanity.exe!`anonymous namespace'::com_ptr<IAudioRenderClient>::release() Line 131`` (In cubeb_wasapi.cpp)

Error Message: 
```
Exception thrown at 0x6A3466C2 (AudioSes.dll) in test_sanity.exe: 0xC0000005: Access violation reading location 0x03810110. occurred
```

Library Function:
```
AudioSes.dll!CAudioRenderClient::Cleanup()
```

Hope this helps!

kinetiknz:
Right, so you're getting past the AUDCLNT_E_BUFFER_SIZE_NOT_ALIGNED issue somehow.  Did you alter the stream params in the test or are we testing different versions of the code?

Anyway, if I hardcode suitable buffer sizing to avoid AUDCLNT_E_BUFFER_SIZE_NOT_ALIGNED, I hit the same crash in stream_start_one_side.  That was tricky to track down... it turns out IAudioClient's initialized with AUDCLNT_SHAREMODE_EXCLUSIVE care if COM is still initialized (they hold a reference to some global/thread-specific COM data and use it during IAudioClient::Start).  The cubeb tests never initialize COM, so we were relying on the auto_com initialization inside cubeb_wasapi.cpp.  Since COM wasn't initialized at a higher level, auto_com completely destroys the COM global data each time... since the IAudioClient is holding a reference to this (and uses it in AUDCLNT_SHAREMODE_EXCLUSIVE ), it becomes invalid and crashes.

I note that wasapi_stream_start/stop are missing auto_com, but adding them doesn't help because the IAudioClient still holds the reference to the old, destroyed COM global data.

So I think the correct fix for this is to initialize COM in the cubeb tests, and to document that COM *must* be initialized before calling cubeb APIs.  That may also mean the auto_com stuff becomes completely obsolete.

kinetiknz:
Next issue to address: after adding a variant of auto_com to test/common.h and using it inside test_tone, the stream starts and stops as expected, but the data_callback is never called, so no tone is produced and the test's assert fails.

spycrab:
> Right, so you're getting past the AUDCLNT_E_BUFFER_SIZE_NOT_ALIGNED issue somehow. Did you alter the stream params in the test or are we testing different versions of the code?

No, I didn't. Anyway I implemented the alignment dance now, not sure if it works properly though.

> Anyway, if I hardcode suitable buffer sizing to avoid AUDCLNT_E_BUFFER_SIZE_NOT_ALIGNED, I hit the same crash in stream_start_one_side. That was tricky to track down... it turns out IAudioClient's initialized with AUDCLNT_SHAREMODE_EXCLUSIVE care if COM is still initialized (they hold a reference to some global/thread-specific COM data and use it during IAudioClient::Start). The cubeb tests never initialize COM, so we were relying on the auto_com initialization inside cubeb_wasapi.cpp. Since COM wasn't initialized at a higher level (meaning auto_com would just shuffle a ref count), auto_com completely destroys the COM global data each time... since the IAudioClient is holding a reference to this (and uses it in AUDCLNT_SHAREMODE_EXCLUSIVE mode), it becomes invalid and crashes.

Thanks for figuring this one out! I added a ``CoInitialize`` call to ``test/common.h`` temporarily.

> Next issue to address: after adding a variant of auto_com to test/common.h and using it inside test_tone, the stream starts and stops as expected, but the data_callback is never called, so no tone is produced and the test's assert fails.

Sorry but I can't reproduce this for some reason :/

kinetiknz:
> Sorry but I can't reproduce this for some reason :/

Possibly for the same reason you didn't hit the buffer alignment issues... your hardware/configuration might happen to be compatible with the configuration requested (or more flexible in accepting configurations).  So we'll need to test this with various configurations and shake out the remaining issues.

spycrab:
@kinetiknz Yeah, my audio setup is kinda weird (USB microphone with a builtin "sound card" and a headset out) :/

spycrab:
So any pointers regarding what I can still do to move this along?

kinetiknz:
Either (or both) extensive testing to shake out what else needs adjusting or reviewing all of the IAudioClient/IAudioRenderClient/IAudioCaptureClient APIs to see what restrictions they have for exclusive mode.  As mentioned in an earlier comment, GetBuffer fails with some configurations as the code is now.

spycrab:
Rebased. Also did some further testing with the devices available to me.
None of them seemed to succumb to the error mentioned here.

So I'm kinda unsure what to do next.

kinetiknz:
I need to find some time to debug the next issue, I guess.  

In the mean time:

- The indentation is different to the file in a lot of your changes (might be tabs vs spaces, hard to tell with GitHub diffs)
- CoInitialize isn't needed in common.h now
- The logging on 1551 shouldn't need to be different for exclusive mode, so that can be merged
- The channel layout comments were removed in the mixer rework, don't add them back (rebase issue?)

Adamillo:
What's the state of this PR?

padenot:
It's never been finished and it doesn't apply anymore, there's been a _lot_ of changes since then to the WASAPI backend.

I wonder how useful are exclusive streams these days.

Adamillo:
Well, exclusive WASAPI in Cubeb reduces latency by a lot, so that's why it's very useful.

PR #421, opened 2018-03-12
Cubeb mixer

djg:
Patches from @jyavenard Gecko [Bug 1432779](https://bugzil.la/1432779) and [Bug 1431221](https://bugzil.la/1431221)


djg:
I fixed up CMakeLists.txt and a problem with use of `assert(false)` in `cubeb_pulse.c`.

kinetiknz:
> Adapted from FFmpeg's libswresample, this mixer allows to convert from
any sane layout to any other and work directly on interleaved samples.

What does adapted mean?  Given the addition of Michael Niedermayer's copyright and the `av_` naming of various parts of the code, I imagine this code retains the original license.  It looks like `rematrix.c` is LGPL, and with libcubeb being ISC, that is quite a significant change.  I don't think  we can  include this in Gecko without moving libcubeb into the `lgpllibs` library, either.

kinetiknz:
> What does adapted mean? Given the addition of Michael Niedermayer's copyright and the av_ naming of various parts of the code, I imagine this code retains the original license. It looks like rematrix.c is LGPL, and with libcubeb being ISC, that is quite a significant change. I don't think we can include this in Gecko without moving libcubeb into the lgpllibs library, either.

I discussed this with @jyavenard on IRC.  The summary is that this code is the best available for our needs, so we'll deal with the licensing.  The current plan is to delineate the LGPL code clearly and provide the ability to build without it (leaving the ISC pue code having basically no remixing ability).  Within Gecko, we'll move libcubeb into lgpllibs.

Longer term (as part of the Rust rewrite work), we'll look at dropping this LGPL code and using libswresample directly for mixing, resampling, and other format conversions.  I'll file a separate issue to discuss and track that work.

kinetiknz:
This is now https://github.com/kinetiknz/cubeb/pull/426.

PR #422, opened 2018-03-12
Relax some of the checks in the loopback test to avoid false positives.

SingingTree:
- For the silent loopback test we check that we don't get any sound above an
epsilon. The value of this epsilon has been increased to accommodate a wider
range of systems.
- The delay on all tests has been increased to allow for more time to capture
looped samples. The original 150ms delay was sufficient on the machine I tested
with, however, I've seen the tests fail due to not having enough looped input.


kinetiknz:
Thanks!

PR #425, opened 2018-03-15
pulse: check pa_context_connect return value

oxij:
In `src/cubeb_pulse.c` many calls to PA API never check return codes. I discovered this while I was debugging why Firefox crashes with my "superior reimplementation" of libpulse API (https://github.com/oxij/libcardiacarrest). This is the I needed for my use case. However, if you look at PA API documentation, you'll notice that not only cubeb_pulse.c ignores some PA errors, it is not entirely clear whenever all of the uses of PA API in cubeb_pulse.c are even correct. For instance, is it ok to `drain` the `pa_context` before it successfully connected to the PA daemon? I'm not sure. cubeb_pulse.c `drain`s it anyway.

(Adapted from https://bugzilla.mozilla.org/show_bug.cgi?id=1444519.)

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -581,6 +581,8 @@ static void pulse_destroy(cubeb * ctx);
 static int
 pulse_context_init(cubeb * ctx)
 {
+  int r;
+
   if (ctx->context) {
     assert(ctx->error == 1);
     pulse_context_destroy(ctx);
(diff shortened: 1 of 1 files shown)


kinetiknz:
Thanks for the PR.  This looks fine.  We'll need to address this in the Rust port, since we ignore the result in that code too (since it's based on the C version): https://github.com/djg/cubeb-pulse-rs/blob/dev/src/backend/context.rs#L601

djg:
@kinetiknz, do you want to close this now?

PR #434, opened 2018-03-26
audiounit: Apply latency correction to stream_get_position.

kinetiknz:
The presentation latency of a stream is roughly kAudioDevicePropertyLatency +
kAudioStreamPropertyLatency + kAudioUnitProperty_Latency.  Calculate this at
device setup time and apply the correction when calculating the current stream
position for stream_get_position.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -70,6 +70,7 @@ static void audiounit_close_stream(cubeb_stream *stm);
 static int audiounit_setup_stream(cubeb_stream *stm);
 static vector<AudioObjectID>
 audiounit_get_devices_of_type(cubeb_device_type devtype);
+static UInt32 audiounit_get_device_presentation_latency(AudioObjectID devid, AudioObjectPropertyScope scope);
 
 extern cubeb_ops const audiounit_ops;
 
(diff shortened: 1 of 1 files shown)


kinetiknz:
Noting for posterity: I removed kAudioDevicePropertySafetyOffset from the latency calculation because it wasn't clear that it's required.  Unfortunately, like any time you try to understand Apple's audio API, it's not well documented and thus the correct approach is ambiguous.  Maybe we should measure it ourselves?

In favour of removing it:

- Chromium's fix (https://bugs.chromium.org/p/chromium/issues/detail?id=783282) doesn't use it.
- A recent clarification thread posted by Chromium engineers and responded to by Apple engineers (https://lists.apple.com/archives/coreaudio-api/2017/Jul/msg00034.html) doesn't mention it.
- Older coreaudio-api posts by Apple engineers don't mention it: https://lists.apple.com/archives/coreaudio-api/2009/Jun/msg00148.html and https://lists.apple.com/archives/coreaudio-api/2008/Jul/msg00077.html

In favour of keeping it:
- Another coreaudio-api post: https://lists.apple.com/archives/coreaudio-api/2010/Jan/msg00046.html
- JUCE request saying it's required to match measured latencies: https://forum.juce.com/t/audio-device-latencies-safety-offsets-on-mac/15647/3

padenot:
> . Maybe we should measure it ourselves?

I have some gear I can use to measure this properly if you want (and a bluetooth setup that has high latency when used with my macbook).

PR #436, opened 2018-03-29
wasapi: check for default playback or capture device on init

achronop:
Cubeb instance will fail to initialize when all playback devices are disabled even if there are capture devices available. This is because wasapi_init() method checks for default playback as an indirect way to verify that wasapi is enabled in the platform. In this patch if get default playback fail we check again for the default capture device.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -1153,8 +1153,12 @@ int wasapi_init(cubeb ** context, char const * context_name)
   com_ptr<IMMDevice> device;
   HRESULT hr = get_default_endpoint(device, eRender);
   if (FAILED(hr)) {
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -1153,8 +1153,11 @@ int wasapi_init(cubeb ** context, char const * context_name)
   com_ptr<IMMDevice> device;
   HRESULT hr = get_default_endpoint(device, eRender);
   if (FAILED(hr)) {
-    LOG("Could not get device: %lx", hr);
```
kinetiknz: Keep the logging so it's clear we didn't find an eRender device and have fallen back to querying eCapture.

PR #439, opened 2018-04-07
wasapi: Fix random crashes on shutdown

spycrab:
Hi there! This is a bug I discovered by accident while debugging a crash in [Dolphin](https://github.com/dolphin-emu/dolphin).

Sometimes when trying to close a cubeb stream, Dolphin will instead randomly crash because of an invalid handle. This PR fixes this.

If there's something off about this PR or this is a result of a mistake on our side, I'd be happy to assist!

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -1222,7 +1222,9 @@ bool stop_and_join_render_thread(cubeb_stream * stm)
     CloseHandle(stm->thread);
     stm->thread = NULL;
 
-    CloseHandle(stm->shutdown_event);
+	if (stm->shutdown_event != INVALID_HANDLE_VALUE) {
+      CloseHandle(stm->shutdown_event);
+	}
     stm->shutdown_event = 0;
   }
 


kinetiknz:
Thanks for the PR.  What's the value of shutdown_handle when the crash occurs?  Can you track down where the invalid handle first appears?  It should be either NULL or a valid handle.

spycrab:
> What's the value of shutdown_handle when the crash occurs?

``-1``. It shows up as ``stm->shutdown`` instead of ``stm->shutdown_event`` (doesn't exist according to VS despite the fact it's clearly in the source code) though which is strange

>  Can you track down where the invalid handle first appears? 

The handle seems to be invalid from the moment it is created but I'm not sure how to properly diagnose where it gets invaldiated.

It seems that ``SetEvent`` always returns ``true`` though.

kinetiknz:
> -1. It shows up as stm->shutdown instead of stm->shutdown_event (doesn't exist according to VS despite the fact it's clearly in the source code) though which is strange

The VS debugger gets confused and shows the layout for the cubeb_stream declared in cubeb_winmm.c.  Coincidentally winmm_stm->shutdown and wasapi_stm->shutdown_event are at the same offset in the struct.  If you edit CMakeLists.txt to exclude the WinMM backend, it'll show the correct layout for WASAPI.

> The handle seems to be invalid from the moment it is created but I'm not sure how to properly diagnose where it gets invaldiated.

So you're seeing CreateEvent return -1 [here](https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_wasapi.cpp#L1974)?  Or does it happen sometime later?

Looking at every store to stm->shutdown_event in cubeb_wasapi.cpp, it's initialized to 0, rewritten with the result of CreateEvent (which according to MSDN should be 0 on error, otherwise a valid value), and rewritten to 0 when stop_and_join_render_thread is called.

If you could put a hardware watchpoint on the location of stm->shutdown_event and work out who's storing -1 into it, that'd help understand what's going on here.

spycrab:
Sorry for the very late response but I was pretty busy.

So I did some more testing and it turns out ``CreateEvent`` actually returns ``-1`` every time it is called for some reason.

I'm currently investigating whether we are using all Cubeb API calls properly, so maybe that's the issue.

spycrab:
We were doing some really weird stuff with our audio streams, so it was on us. Mea culpa.

kinetiknz:
Thanks for following up!

PR #442, opened 2018-05-29
osx: increase active streams counter under stream lock, BMO 1462210.

achronop:
`active_streams` counter is an atomic but it is required to change under the stream lock in order to be synchronized with the other set of actions that are taking place during stream initialization. This is found in [Bug1462210](https://bugzilla.mozilla.org/show_bug.cgi?id=1462210).

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -2578,7 +2578,6 @@ audiounit_stream_init(cubeb * context,
 {
   unique_ptr<cubeb_stream, decltype(&audiounit_stream_destroy)> stm(new cubeb_stream(context),
                                                                     audiounit_stream_destroy);
-  context->active_streams += 1;
   int r;
 
   assert(context);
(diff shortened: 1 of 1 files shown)


kinetiknz:
It looks like the context mutex is held whenever audiounit_set_global_latency is called but the assert in audiounit_set_global_latency should be checking the *context* mutex is held, not the stream mutex.

- audiounit_destroy needs to take the context mutex earlier for the active_streams log/assert logic.
- audiounit_clamp_latency should assert the context mutex is held before accessing active_streams.

And to fix your original issue, you can probably get away with taking the context mutex earlier in audiounit_stream_init.

achronop:
> If an early return causes audiounit_stream_destroy to run before we reach line 2620, audiounit_stream_destroy will decrement the stream count for a stream that never incremented it.

Thanks, I should have noticed myself. I've updated the patch according to your instructions. I am also running a [try build]https://treeherder.mozilla.org/#/jobs?repo=try&revision=dcfee67b67c41ff44cec5033cdd0926611e0a06fl).

achronop:
Thank you for the review.

PR #443, opened 2018-06-01
return unique_ptr normally, without std::move

froydnj:
clang warns about using `return std::move($LOCAL)`.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -2138,7 +2138,7 @@ utf8_to_wstr(char const * str)
 
   std::unique_ptr<wchar_t []> ret(new wchar_t[size]);
   ::MultiByteToWideChar(CP_UTF8, 0, str, -1, ret.get(), size);
-  return std::move(ret);
+  return ret;
 }
 
 static com_ptr<IMMDevice>


kinetiknz:
[This kills the mingw.](https://travis-ci.org/kinetiknz/cubeb/jobs/386625571) ... which is I think is the reason we have this in the first place. 😞

froydnj:
Do we need such an old mingw build (4.8?) in the first place?  Is it possible to use a newer version of mingw?

kinetiknz:
We can use whatever the minimum mingw to build Tor Browser is, although I'm having trouble finding out what that is.

But we're also stuck with ancient (14.04) Ubuntu images on Travis, so upgrading to a newer mingw is not trivial unless there are backports available.

Another option would be to move the mingw build to AppVeyor and use the native mingw build rather than cross-compile.

danilaml:
>But we're also stuck with ancient (14.04) Ubuntu images on Travis, so upgrading to a newer mingw is not trivial unless there are backports available.

Travis had Xenial (16.04) images for a while: https://docs.travis-ci.com/user/reference/xenial/

kinetiknz:
Thanks for the heads up!  Looks like it has been available since November.  I've had a crack at switching the builds to Xenial here: https://github.com/kinetiknz/cubeb/tree/travis_xenial - there's some breakage to solve - the Android build is broken by newer CMake versions.

PR #444, opened 2018-06-09
Fix volume handling in sndio backend

landryb:
Apply volume in software as do other backends. This is necessary
because sndio volume may be controlled externally and there's no
volume getter in libcubeb to notify the caller about volume
changes.

Cf https://bugzilla.mozilla.org/show_bug.cgi?id=1467882


kinetiknz:
Thanks!

PR #447, opened 2018-06-26
wasapi: Reduce timeout threshold

Pehrsons:
See [bug 1465617 comment 20](https://bugzilla.mozilla.org/show_bug.cgi?id=1465617#c20) for the background for this change.

In short the Firefox shutdown hang monitor also times out after 5 seconds by default, so cubeb has to timeout faster or confusing almost-coherent stacks show up in crash-stats.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -871,7 +871,7 @@ wasapi_stream_render_loop(LPVOID stream)
      the timeout error handling only when the timeout_limit is reached, which is
      reset on each successful loop. */
   unsigned timeout_count = 0;
-  const unsigned timeout_limit = 5;
+  const unsigned timeout_limit = 3;
   while (is_playing) {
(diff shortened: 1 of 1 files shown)


padenot:
r+ let's land this soon to see if it makes things better.

danilaml:
Is this still relevant fix for mozilla? I found that it often reports an error and dies after waking up from sleep in a few programs that use cubeb with timeout counter < 5. Alternatively, https://github.com/mozilla/cubeb/pull/542 would probably fix that, but it seems there were no activity there for quite some time.

padenot:
I've commented in #542: rebased the patch on top of current master branch, and kicked off a couple CI runs: cubeb's own CI, that can now run Windows tests, and Firefox's CI, that's very unforgiving because it runs quite a lot of tests and quite a number of different workloads, on different OS.

I think we completely forgot about it, sorry about that, let's get a solution merged.

PR #449, opened 2018-07-09
Fix for #448

### LINKED ISSUE #448: Mono plays on left channels only on macOS < 10.12
From https://bugzilla.mozilla.org/show_bug.cgi?id=1469152

The perform proper upmix/downmix we query CoreAudio for the kAudioUnitProperty_AudioChannelLayout property which will tell us what the default output layout is.

However, kAudioUnitProperty_AudioChannelLayout only exists in macOS 10.12 and later.

So playback of mono (and other multichannels) is broken on older macOS

jyavenard:
The method kAudioUnitProperty_AudioChannelLayout used to retrieve the channel layout wasn't introduced until 10.12. So we use kAudioDevicePropertyPreferredChannelLayout instead should it fails.

Fixes #448

### DIFF
--- a/src/cubeb_mixer.cpp
+++ b/src/cubeb_mixer.cpp
@@ -533,6 +533,18 @@ struct cubeb_mixer
         output_buffer += _context._out_ch_count - _context._in_ch_count;
       }
     } else {
+      if (_context._in_ch_count == 1 && _context._out_ch_count >= 2) {
+        // Special case for upmixing mono input to stereo and more. We will
+        // duplicate the mono channel to the first two channels. On most system,
+        // the first two channels are for left and right. It is commonly
+        // expected that mono will on both left+right channels
+        for (uint32_t i = 0; i < frames; i++) {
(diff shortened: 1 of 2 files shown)


src/cubeb_mixer.cpp
```diff
@@ -533,6 +533,17 @@ struct cubeb_mixer
...
+        // Special case for upmixing mono input to stereo and more. We will
+        // duplicate the mono channel to the first two channels. On most system,
+        // the first two channels are for left and right. It is commonly
+        // expected that mono will on both left+right channels
+        for (uint32_t i = 0; i < frames; i++) {
+          output_buffer[0] = output_buffer[1] = *input_buffer;
+          output_buffer += _context._out_ch_count ;
+          input_buffer++;
+        }
+      }
```
kinetiknz: Need an else or early return here? It doesn't look like this is intended to fall through to the PodCopy loop below.

jyavenard:
Indeed, oops...

thanks for spotting that dumb mistake

PR #451, opened 2018-07-12
Simplify audiounit_add/remove_listener() by introducing property_listener

ChunMinChang:
```audiounit_add_listener(...)``` and ```audiounit_remove_listener(...)``` should use exactly same variable to register and unregister event listener. Originally, these two functions are set by different local variables with same values: ```audiounit_add_listener(X, Y, Z, ...)``` and ```audiounit_remove_listener(X', Y', Z', ...)```, where X = X', Y = Y',...etc. There are too many function parameters so it's not easy to check if we have same parameters for ```audiounit_add_listener(...)``` and ```audiounit_remove_listener(...)```.

```property_listener``` is introduced to simplify the parameter issues by using ```audiounit_add/remove_listener(property_listener L)```. A property_listener object L is created to register a event listener, and then be used to unregister the listener later. It'll be safer if ```audiounit_add_listener(...)``` and ```audiounit_remove_listener(...)``` use the exactly same object as their parameters. In addition, The readability is better since there is a **listener** be added/removed when we call ```audiounit_add/remove_listener(listener L)```.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -3308,11 +3308,10 @@ audiounit_device_collection_destroy(cubeb * /* context */,
 static vector<AudioObjectID>
 audiounit_get_devices_of_type(cubeb_device_type devtype)
 {
-  AudioObjectPropertyAddress adr = { kAudioHardwarePropertyDevices,
-                                     kAudioObjectPropertyScopeGlobal,
-                                     kAudioObjectPropertyElementMaster };
   UInt32 size = 0;
-  OSStatus ret = AudioObjectGetPropertyDataSize(kAudioObjectSystemObject, &adr, 0, NULL, &size);
+  OSStatus ret = AudioObjectGetPropertyDataSize(kAudioObjectSystemObject,
+                                                &DEVICES_PROPERTY_ADDRESS, 0,
+                                                NULL, &size);
   if (ret != noErr) {
     return vector<AudioObjectID>();
   }
@@ -3320,7 +3319,9 @@ audiounit_get_devices_of_type(cubeb_device_type devtype)
   uint32_t count = (uint32_t)(size / sizeof(AudioObjectID));
 
   vector<AudioObjectID> devices(count);
-  ret = AudioObjectGetPropertyData(kAudioObjectSystemObject, &adr, 0, NULL, &size, devices.data());
+  ret = AudioObjectGetPropertyData(kAudioObjectSystemObject,
+                                   &DEVICES_PROPERTY_ADDRESS, 0, NULL, &size,
+                                   devices.data());
   if (ret != noErr) {
     return vector<AudioObjectID>();
   }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by kinetiknz** [comment]
The changes themselves look fine, I think.

I have a more general comment about refactoring.  The longer term plan is to replace this backend with a version rewritten in Rust, like we did with the cubeb_pulseaudio.c backend.  In the meantime, I think we should only make changes to the code that fix bugs or alter the code in such a way that it makes the Rust rewrite easier to complete (e.g. if a construct is difficult to represent in Rust, we can change the C code to something more suitable first).

I'm concerned that more general clean up style refactoring is going to make the Rust rewrite and associated work (reviewing, bug tracking, etc.) more complex, because any regressions introduced by refactorings can take quite a while to be discovered (until they reach Firefox release, generally, unfortunately) and that will leave us with a situation where we are maintaining, effectively, three versions of this code: the code we shipped already (and may still need to backport emergency fixes to), the refactored code we're shipping next, and the new Rust code.  I think it's important to keep code churn to a minimum in the interim to make it easier to track bugs and regressions between versions.

**Comment by ChunMinChang** [comment]
The reason I make this change is to put the same platform-dependent C variables together, instead of creating them in each function, so it'll be easier to use them when they are transformed into Rust variables.

I plan to split the device-related code and the stream-related code in the future when it's rewritten into Rust. There is no test for device-changed listener so it's hard to make sure it works on every situation. The only way to test it is to switch or (un)plug devices manually and see what happens. Without the proper test, the risk for refactoring still exists even if it's rewritten into Rust. Maybe it's better to find a way to write a test first.

I do understand your concern about maintaining different versions of code. It's ok to postpone this change until it's rewritten into Rust if it's easier for maintaining.

File: src/cubeb_audiounit.cpp [hunk 1]
```diff
@@ -892,31 +951,23 @@ audiounit_property_listener_callback(AudioObjectID id, UInt32 address_count,
 }
 
 OSStatus
-audiounit_add_listener(cubeb_stream * stm, AudioDeviceID id, AudioObjectPropertySelector selector,
-    AudioObjectPropertyScope scope, AudioObjectPropertyListenerProc listener)
+audiounit_add_listener(property_listener * listener)
```
kinetiknz: property_listener pointer can be const here and for audiounit_remove_listener.

kinetiknz: It might make sense to move add/remove_listener to member functions of the property_listener, but it's not a big deal.

src/cubeb_audiounit.cpp
```diff
@@ -1028,17 +1095,15 @@ audiounit_uninstall_system_changed_callback(cubeb_stream * stm)
 {
   OSStatus r;
 
-  if (has_output(stm)) {
-    r = audiounit_remove_listener(stm, kAudioObjectSystemObject, kAudioHardwarePropertyDefaultOutputDevice,
-                                  kAudioObjectPropertyScopeGlobal, &audiounit_property_listener_callback);
+  if (has_output(stm) && stm->default_output_listener) {
```
kinetiknz: default_output_listener will be a nullptr if has_output was false during init, right? So we can simplify here and drop the has_output check, since it must be true.

Same for has_input below.

kinetiknz:
Needs a rebase after #453 was merged.

PR #453, opened 2018-07-13
Reduce duplicate code in audiounit_stream_get_current_device

ChunMinChang:
1. Add a simple test for ```cubeb_stream_get_current_device(...)```
2. Reduce duplicate code for input and output in ```audiounit_stream_get_current_device(...)```

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -2891,15 +2891,19 @@ int audiounit_stream_set_panning(cubeb_stream * stm, float panning)
 
 unique_ptr<char[]> convert_uint32_into_string(UInt32 data)
 {
-  /* Simply create an empty string if no data. */
-  size_t size = data == 0 ? 0 : sizeof(data);
-  auto str = unique_ptr<char[]> { new char[size + 1] };
-  // Reverse 0xWXYZ into 0xZYXW.
-  for (size_t i = 0; i < size; ++i) {
-    char * p = (char *) &data;
-    str[i] = p[size - 1 - i];
-  }
+  // Simply create an empty string if no data.
+  size_t size = data == 0 ? 0 : 4; // 4 bytes for uint32.
+  auto str = unique_ptr<char[]> { new char[size + 1] }; // + 1 for '\0'.
   str[size] = '\0';
+  if (size < 4) {
+    return str;
+  }
+
+  // Reverse 0xWXYZ into 0xZYXW.
+  str[0] = (char)(data >> 24);
+  str[1] = (char)(data >> 16);
+  str[2] = (char)(data >> 8);
+  str[3] = (char)(data);
   return str;
 }
 
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -2907,8 +2907,8 @@ unique_ptr<char[]> convert_uint32_into_string(UInt32 data)
   return str;
 }
 
-int audiounit_get_default_device_data(cubeb_device_type type,
-                                      UInt32 * data)
+int audiounit_get_default_device_datasource(cubeb_device_type type,
+                                            UInt32 * data)
 {
   AudioDeviceID id = audiounit_get_default_device_id(type);
   if (id == kAudioObjectUnknown) {
@@ -2949,7 +2949,7 @@ int audiounit_get_default_device_name(cubeb_stream * stm,
   assert(device);
 
   UInt32 data;
-  int r = audiounit_get_default_device_data(type, &data);
+  int r = audiounit_get_default_device_datasource(type, &data);
   if (r != CUBEB_OK) {
     return r;
   }
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -2908,8 +2908,7 @@ unique_ptr<char[]> convert_uint32_into_string(UInt32 data)
 }
 
 int audiounit_get_default_device_data(cubeb_device_type type,
-                                      UInt32 * data,
-                                      cubeb_stream * stm /* for log */)
+                                      UInt32 * data)
 {
   AudioDeviceID id = audiounit_get_default_device_id(type);
   if (id == kAudioObjectUnknown) {
@@ -2936,8 +2935,6 @@ int audiounit_get_default_device_data(cubeb_device_type type,
                                             &datasource_address_output,
                                           0, NULL, &size, data);
   if (r != noErr) {
-    LOG("(%p) Error when getting %s device!", stm,
-        type == CUBEB_DEVICE_TYPE_INPUT ? "input" : "output");
     *data = 0;
   }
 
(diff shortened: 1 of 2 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -2889,21 +2889,30 @@ int audiounit_stream_set_panning(cubeb_stream * stm, float panning)
   return CUBEB_OK;
 }
 
-int audiounit_stream_get_current_device(cubeb_stream * stm,
-                                        cubeb_device ** const device)
+void allocate_and_convert_uint32_into_string(char ** str, UInt32 data)
```
kinetiknz: This could return a unique_ptr<char[]> instead.

It can also have a simpler name; once it returns a unique_ptr it is obvious that it allocates from the return type.

src/cubeb_audiounit.cpp
```diff
@@ -2889,21 +2889,30 @@ int audiounit_stream_set_panning(cubeb_stream * stm, float panning)
...
-  UInt32 data;
-  char strdata[4];
-  AudioDeviceID output_device_id;
-  AudioDeviceID input_device_id;
+  /* Simply create an empty string if no data. */
+  size_t size = data == 0 ? 0 : sizeof(data);
+  *str = new char[size + 1];
+  assert(*str);
+  // Reverse 0xWXYZ into 0xZYXW.
+  for (size_t i = 0; i < size; ++i) {
```
kinetiknz: It'd be simpler to keep the old bitshifting conversion since it only ever needs to handle 4 bytes.

src/cubeb_audiounit.cpp
```diff
@@ -2889,21 +2889,30 @@ int audiounit_stream_set_panning(cubeb_stream * stm, float panning)
...
+    char * p = (char *) &data;
+    (*str)[i] = p[size - 1 - i];
+  }
+  (*str)[size] = '\0';
+}
 
-  AudioObjectPropertyAddress datasource_address = {
+int audiounit_get_default_device_data(cubeb_device_type type,
+                                      UInt32 * data,
+                                      cubeb_stream * stm /* for log */)
```
kinetiknz: Drop the comment here, it'll end up being out of date.

src/cubeb_audiounit.cpp
```diff
@@ -2915,70 +2924,66 @@ int audiounit_stream_get_current_device(cubeb_stream * stm,
...
-  (*device)->output_name = new char[size + 1];
-  if (!(*device)->output_name) {
-    return CUBEB_ERROR;
+  UInt32 data;
+  int r = audiounit_get_default_device_data(type, &data, stm);
+  if (r != CUBEB_OK) {
+    return r;
   }
+  char ** name = type == CUBEB_DEVICE_TYPE_INPUT ?
+    &(device->input_name) : &(device->output_name);
```
kinetiknz: You don't need parenthesis around device->input_name and device->output_name, but once allocate_and_convert_uint32_into_string returns a unique_ptr it doesn't matter.

test/test_devices.cpp
```diff
@@ -205,3 +205,48 @@ TEST(cubeb, enumerate_devices)
 
   cubeb_stream_destroy(stream);
 }
+
+TEST(cubeb, stream_get_current_device)
```
kinetiknz: It's not a bad thing to have a test to exercise this API, but there are not many ways the test can fail right now because it doesn't verify that the result of cubeb_stream_get_current_device is meaningful.  Can we improve this somehow?

src/cubeb_audiounit.cpp
ChunMinChang: Do you mean it should be changed into something like the following?
```cpp
unique_ptr<char[]> convert_uint32_into_string(UInt32 data)
{
  ...
}

int audiounit_get_default_device_name(...)
{
  ...
  unique_ptr<char[]> str = convert_uint32_into_string(...);
  char ** name = ...// input_name or output_name
  *name = str.release();
}
```
The cubeb_device->input/output_name is a raw pointer. ```release()``` will be called when moving the unique pointer to this raw pointer. But it will drop the advantage of ```unique_ptr```. The effect is same as returning raw pointer from ```convert_uint32_into_string```.

test/test_devices.cpp
ChunMinChang: This API is to get the names of the default devices, so it's hard to check that returned values. The default devices is platform-dependent. For now it's just used for manually check. If we can get the default devices by other way, then it's possible to compare the results.

src/cubeb_audiounit.cpp
ChunMinChang: The default device is set by system. It can be changed by system setting no matter what stream is. It's independent from streams. Could cubeb_stream be removed from this API? or we need this to debug?

src/cubeb_audiounit.cpp
ChunMinChang: ok. In that case, I'd like to specify ```size_t size``` to a fixed size since we use a fixed size for shift the bytes.

src/cubeb_audiounit.cpp
kinetiknz: Yes, but you can assign the result of convert_uint32_into_string(...).release() directly into *name without the temporary.

There are still benefits to using unique_ptr this way - it's harder for any (other) use of the function to leak and it makes it clear that the function allocates and transfers ownership to the caller.

src/cubeb_audiounit.cpp
kinetiknz: I'm not sure I follow.  If you keep the old style of:

```
  strdata[0] = (char)(data >> 24);
  strdata[1] = (char)(data >> 16);
  strdata[2] = (char)(data >> 8);
  strdata[3] = (char)(data);
```

You don't need to worry about sizes.

kinetiknz: I'm not sure I follow.  If you keep the old style of:

```
  strdata[0] = (char)(data >> 24);
  strdata[1] = (char)(data >> 16);
  strdata[2] = (char)(data >> 8);
  strdata[3] = (char)(data);
```

You don't need to worry about sizes.

src/cubeb_audiounit.cpp
kinetiknz: You could log in the failure path of audiounit_get_default_device_data's caller and avoid passing the stm in.

test/test_devices.cpp
kinetiknz: Fair enough, we can extend it later if we come up with ideas.

ChunMinChang:
Hmm... I think it's better to rename ```audiounit_stream_get_current_device``` to ```audiounit_stream_get_current_device_source```. 

We should use an explicit name for this function since the same device will return different datasources. For example, the default input device on macbook pro will return "imic" if it uses the default internal microphone or "emic" when it uses an external microphone plugged in audio jack. The name ```audiounit_stream_get_current_device``` is vague. Users will expect it returns a device-name instead of a data-source name of a device when they see it. However, it might be out-of-topic here. I can open another issue to address it later.

kinetiknz:
audiounit_stream_get_current_device isn't public, it's called via cubeb_stream_get_current_device.  We can rename the public API and all of the implementations if necessary, but it may be sufficient to add documentation alongside cubeb_stream_get_current_device or the cubeb_device struct in cubeb.h to clarify how it behaves.

PR #454, opened 2018-07-17
Remove c-style cast in cubeb_resampler.cpp

padenot:


src/cubeb_resampler.cpp
```diff
@@ -193,8 +193,10 @@ ::fill_internal_input(T * input_buffer, long * input_frames_count,
 
   /* process the input, and present exactly `output_frames_needed` in the
   * callback. */
+  size_t frames_resampled = 0;
```
achronop: nit for both: you can bring the declaration closer to the usage.

padenot:
Yeah, I've swapped the lines.

PR #461, opened 2018-10-04
 tests: check correctly the audio buffer values range (BMO 1484541). 

achronop:
* Fix the check of audio value range (BMO 1484541) for record and duplex test.
* In the duplex test, the correction revealed an error. The values in the input buffer can be outside the range of [-1, 1]. Those values are coming directly from AD. I verified that in input callback before happening any other potential processing. The input side, in that test, is configured as mono but the output as stereo. Probably this is the result of some mixing inside AD. II configured the input to stereo similar to output in order to silence the error. In general, it would be nice to start talking about how we could replace AD.
* Added macros to make it easier to enable logs in tests.


kinetiknz:
I asked @ChunMinChang to review this.

test/test_duplex.cpp
```diff
@@ -101,14 +102,14 @@ TEST(cubeb, duplex)
 
   /* typical user-case: mono input, stereo output, low latency. */
```
ChunMinChang: Should we change this comment if the input is changed to stereo?

achronop:
> Does it mean the aggregate device created inside cubeb has stereo input and stereo output even we set the input to mono?

Right, AD is stereo but we configure input AU as mono.

> Could I have more detailed information about this? Will this happen when we set the stream to mono for a physical mono-input device? or it only happens when we set the stream to mono for an stereo input device?

I do not have have the details because I am doing tries and errors. My guess is that some kind of mixing is applied that exceeds [-1, 1] range. I am testing my Plantronics headset and I do not have it. I have it with built in devices. It happens when volume is on maximum and echo is created. If I lower the volume of mic and speaker it does not happen.

Sorry I am based on observations here.

PR #463, opened 2018-10-08
audiounit: report removed devices correctly (BMO 1491152)

achronop:
Collection changed callback did not handle correctly removed devices. This is because we verify that callback has not been triggered due to adding/removing of the Aggregate Device before reporting a device changed callback .

In this fix I have removed the enumeration of the AD completely so devices comparing includes "actual" devices. Any difference on the actual device will trigger a callback.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -3396,16 +3396,25 @@ audiounit_get_devices_of_type(cubeb_device_type devtype)
   if (ret != noErr) {
     return vector<AudioObjectID>();
   }
-  /* Total number of input and output devices. */
-  uint32_t count = (uint32_t)(size / sizeof(AudioObjectID));
-
-  vector<AudioObjectID> devices(count);
+  vector<AudioObjectID> devices(size / sizeof(AudioObjectID));
   ret = AudioObjectGetPropertyData(kAudioObjectSystemObject,
                                    &DEVICES_PROPERTY_ADDRESS, 0, NULL, &size,
                                    devices.data());
   if (ret != noErr) {
     return vector<AudioObjectID>();
   }
+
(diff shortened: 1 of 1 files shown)


kinetiknz:
I asked @ChunMinChang to review this.

src/cubeb_audiounit.cpp
```diff
@@ -3511,8 +3492,8 @@ audiounit_add_device_listener(cubeb * context,
     assert(context->devtype_device_array.empty());
     /* Listener works for input and output.
      * When requested one of them we need to differentiate. */
-    if (devtype == CUBEB_DEVICE_TYPE_INPUT ||
-        devtype == CUBEB_DEVICE_TYPE_OUTPUT) {
+    if (devtype & CUBEB_DEVICE_TYPE_INPUT ||
+        devtype & CUBEB_DEVICE_TYPE_OUTPUT) {
```
ChunMinChang: Same as above. Could we add an assertion like `assert(devtype & (CUBEB_DEVICE_TYPE_INPUT | CUBEB_DEVICE_TYPE_OUTPUT));` at the beginning of this function, and remove the
```
if (devtype & CUBEB_DEVICE_TYPE_INPUT ||
    devtype & CUBEB_DEVICE_TYPE_OUTPUT) {
 ...
}
```
block? And it's clearer that when `context->collection_changed_* ` are set, the `context->devtype_device_array` is also set.

achronop:
Thank you for the comments so far. I agree with your explanation and I have updated the patches. Please let me know if you have more comments for me.

ChunMinChang:
Looks good to me! Thanks for the fix!

achronop:
@ChunMinChang thanks for the review, I am merging myself ...

PR #464, opened 2018-10-12
pulse: separate input output callbacks for collection changes (BMO 1498242)

achronop:
Use separate callback for input and output devices. This will allow to register a context with different notification callbacks for different type of devices.


achronop:
I am working on the others platforms on different branches.

For now I am taking care to allow registration for (INPUT | OUTPU). The result will be to register the same user callback and the same user pointer for INPUT and OUTPUT. However, it makes sense to restrict that case and not allow it any more. We could decide that after these patches are landed.

PR #465, opened 2018-10-15
audiounit: separate input output callbacks for collection changes (BMO 1498242)

achronop:
Use separate callback for input and output devices. This will allow to register a context with different notification callbacks for different type of devices.

See also: #464

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -132,11 +132,13 @@ struct cubeb {
   owned_critical_section mutex;
   int active_streams = 0;
   uint32_t global_latency_frames = 0;
-  cubeb_device_collection_changed_callback collection_changed_callback = nullptr;
-  void * collection_changed_user_ptr = nullptr;
-  /* Differentiate input from output devices. */
-  cubeb_device_type collection_changed_devtype = CUBEB_DEVICE_TYPE_UNKNOWN;
-  vector<AudioObjectID> devtype_device_array;
+  cubeb_device_collection_changed_callback input_collection_changed_callback = nullptr;
+  void * input_collection_changed_user_ptr = nullptr;
+  cubeb_device_collection_changed_callback output_collection_changed_callback = nullptr;
+  void * output_collection_changed_user_ptr = nullptr;
+  // Store list of devices to detect changes
+  vector<AudioObjectID> input_device_array;
+  vector<AudioObjectID> output_device_array;
   // The queue is asynchronously deallocated once all references to it are released
   dispatch_queue_t serial_queue = dispatch_queue_create(DISPATCH_QUEUE_LABEL, DISPATCH_QUEUE_SERIAL);
   // Current used channel layout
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -3450,23 +3455,27 @@ audiounit_collection_changed_callback(AudioObjectID /* inObjectID */,
   // This can be called from inside an AudioUnit function, dispatch to another queue.
   dispatch_async(context->serial_queue, ^() {
     auto_lock lock(context->mutex);
-    if (context->collection_changed_callback == NULL) {
+    if (!context->input_collection_changed_callback
+      && !context->output_collection_changed_callback) {
```
kinetiknz: `&&` on the previous line (applies to all later changes in the patch with this formatting too)

src/cubeb_audiounit.cpp
```diff
@@ -3477,60 +3486,84 @@ audiounit_add_device_listener(cubeb * context,
                               cubeb_device_collection_changed_callback collection_changed_callback,
                               void * user_ptr)
 {
+  context->mutex.assert_current_thread_owns();
+  assert(devtype & (CUBEB_DEVICE_TYPE_INPUT | CUBEB_DEVICE_TYPE_OUTPUT));
   /* Note: second register without unregister first causes 'nope' error.
    * Current implementation requires unregister before register a new cb. */
-  assert(context->collection_changed_callback == NULL);
+  assert((devtype & CUBEB_DEVICE_TYPE_INPUT) && !context->input_collection_changed_callback
+         || (devtype & CUBEB_DEVICE_TYPE_OUTPUT) && !context->output_collection_changed_callback);
```
kinetiknz: It's probably better as two asserts, one each for input and output.

src/cubeb_audiounit.cpp
```diff
@@ -3477,60 +3486,84 @@ audiounit_add_device_listener(cubeb * context,
...
 
-  OSStatus ret = AudioObjectAddPropertyListener(kAudioObjectSystemObject,
-                                                &DEVICES_PROPERTY_ADDRESS,
-                                                audiounit_collection_changed_callback,
-                                                context);
-  if (ret == noErr) {
+  if (!context->input_collection_changed_callback
+      && !context->output_collection_changed_callback) {
```
kinetiknz: Shouldn't this be `(context->input_collection_changed_callback || context->output_collection_changed_callback)`?

src/cubeb_audiounit.cpp
achronop: If you mean
```
assert((devtype & CUBEB_DEVICE_TYPE_INPUT) && !context->input_collection_changed_callback);
assert((devtype & CUBEB_DEVICE_TYPE_OUTPUT) && !context->output_collection_changed_callback);
```
Should not work, when we register for input (only) the output will fail.

src/cubeb_audiounit.cpp
achronop: Yeah, that's a tricky one. You cannot do that because `input_collection_changed_callback` or `output_collection_changed_callback` are set further bellow. They are expected to be null at that point on first pass. On any following pass, one of them may not be null and then we do not have to call register again. Actually it will fail if we register for a second time with a listener already registered.

src/cubeb_audiounit.cpp
kinetiknz: Oops, of course.

kinetiknz: Nit: there are still a couple of cases where operators are wrapped to the beginning of the next line - here and line 3540

kinetiknz:
Thanks, LGTM

PR #466, opened 2018-10-17
Build and link pulse rust backend if available.

achronop:
If the `cubeb-pulse-rs` repo is available in `cubeb/src` build the project with `USE_PULSE_RUST` and link `libcubeb_pulse` to `libcubeb`.

The author of the patch is @djg , I am only pushing it.


kinetiknz:
Note to anyone using this: it will only build cubeb-pulse-rs once and won't pick up changes to the cubeb-pulse-rs repo automatically - you need to run cargo build --release inside the cubeb-pulse-rs repo if it's ever updated.

PR #471, opened 2018-10-29
osx: check null reference before using the string (BMO 1502165).

achronop:
Check for null reference. Fixing crash:
https://crash-stats.mozilla.com/report/index/d4e665aa-1cd7-4547-95b6-877670181023


kinetiknz:
Do we know what causes the nullptr name?

achronop:
`get_device_name` method can return `null` when `AudioObjectGetPropertyData` fails to get a name. Now the question is when this could fail ... I'm afraid I don't have a straight answer for that, maybe a race with a removed device, or wired hardware ...

PR #472, opened 2018-10-30
Tests for device-collection-changed callbacks

ChunMinChang:
It would be better if we have tests for device-collection-changed callbacks. 
1. For Rust version, it needs to check if the device ralated APIs can work with C APIs.
2. We are refactoring the device-collection-changed callbacks for [BMO 1498242](https://bugzilla.mozilla.org/show_bug.cgi?id=1498242). It should be helpful to check if the code works like what we expect.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,7 +4,3 @@
 [submodule "cmake/sanitizers-cmake"]
 	path = cmake/sanitizers-cmake
 	url = https://github.com/arsenm/sanitizers-cmake
-[submodule "src/cubeb-coreaudio-rs"]
-	path = src/cubeb-coreaudio-rs
-	url = https://github.com/ChunMinChang/cubeb-coreaudio-rs.git
-	branch = build-config-in-cubeb
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,7 @@
 [submodule "cmake/sanitizers-cmake"]
 	path = cmake/sanitizers-cmake
 	url = https://github.com/arsenm/sanitizers-cmake
+[submodule "src/cubeb-coreaudio-rs"]
+	path = src/cubeb-coreaudio-rs
+	url = https://github.com/ChunMinChang/cubeb-coreaudio-rs.git
+	branch = build-config-in-cubeb
--- a/test/test_devices.cpp
+++ b/test/test_devices.cpp
@@ -258,7 +258,7 @@ void device_collection_changed_callback(cubeb * context, void * user)
 {
   fprintf(stderr, "device collection changed callback\n");
   ASSERT_TRUE(false) << "Error: device collection changed callback"
-                        " called when opening a stream";
+                        " called without changing devices";
 }
 
 TEST(cubeb, register_device_collection_change_for_unknown_type)
--- a/test/test_devices.cpp
+++ b/test/test_devices.cpp
@@ -338,7 +338,7 @@ TEST(cubeb, device_collection_change)
                                                  nullptr);
     ASSERT_EQ(r, CUBEB_OK) << "Error registering device collection changed";
 
-    // Unregister all the callbacks regardless of the scope.
+    // Unregister all the callbacks regardless of the scopes.
     r = cubeb_register_device_collection_changed(ctx,
                                                  static_cast<cubeb_device_type>(CUBEB_DEVICE_TYPE_INPUT |
                                                                                 CUBEB_DEVICE_TYPE_OUTPUT),
@@ -435,7 +435,8 @@ TEST(cubeb, register_device_collection_changed_twice)
       cubeb_register_device_collection_changed(ctx,
                                                scope,
                                                device_collection_changed_callback,
-                                               nullptr), ""
+                                               nullptr),
+      ""
     );
   }
 }
--- a/test/test_devices.cpp
+++ b/test/test_devices.cpp
@@ -431,7 +431,7 @@ TEST(cubeb, register_device_collection_changed_twice)
     ASSERT_EQ(r, CUBEB_OK) << "Error registering device collection changed";
 
     // Get an assertion fails when registering a callback within same scope twice.
-    ASSERT_DEBUG_DEATH(
+    ASSERT_DEATH(
       cubeb_register_device_collection_changed(ctx,
                                                scope,
                                                device_collection_changed_callback,
--- a/test/test_devices.cpp
+++ b/test/test_devices.cpp
@@ -274,9 +274,41 @@ TEST(cubeb, register_device_collection_change_for_unknown_type)
 
   r = cubeb_register_device_collection_changed(ctx,
                                                static_cast<cubeb_device_type>(CUBEB_DEVICE_TYPE_UNKNOWN),
+                                               nullptr,
+                                               nullptr);
+  ASSERT_EQ(r, CUBEB_ERROR_INVALID_PARAMETER) << "Error returning wrong error type";
+
+  r = cubeb_register_device_collection_changed(ctx,
+                                               static_cast<cubeb_device_type>(CUBEB_DEVICE_TYPE_UNKNOWN),
                                                device_collection_changed_callback,
                                                nullptr);
   ASSERT_EQ(r, CUBEB_ERROR_INVALID_PARAMETER) << "Error returning wrong error type";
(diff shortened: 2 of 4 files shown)


kinetiknz:
Looks like the tests are hanging on Travis Linux builds?

I'll leave review for @achronop

kinetiknz:
We only run the tests on Linux sadly - the other CI environments have no audio hardware to run against.

test/test_devices.cpp
```diff
@@ -252,4 +252,145 @@ TEST(cubeb, stream_get_current_device)
...
+  std::unique_ptr<cubeb, decltype(&cubeb_destroy)>
+    cleanup_cubeb_at_exit(ctx, cubeb_destroy);
+
+  int scopes[3] = {
+    CUBEB_DEVICE_TYPE_INPUT,
+    CUBEB_DEVICE_TYPE_OUTPUT,
+    CUBEB_DEVICE_TYPE_INPUT | CUBEB_DEVICE_TYPE_OUTPUT
+  };
+
+  for (size_t i = 0 ; i < ARRAY_LENGTH(scopes) ; ++i) {
```
achronop: same

.gitmodules
```diff
@@ -4,3 +4,7 @@
 [submodule "cmake/sanitizers-cmake"]
 	path = cmake/sanitizers-cmake
 	url = https://github.com/arsenm/sanitizers-cmake
+[submodule "src/cubeb-coreaudio-rs"]
+	path = src/cubeb-coreaudio-rs
+	url = https://github.com/ChunMinChang/cubeb-coreaudio-rs.git
+	branch = build-config-in-cubeb
```
achronop: Why do we need that change in this patch?

test/test_devices.cpp
```diff
@@ -252,4 +252,145 @@ TEST(cubeb, stream_get_current_device)
...
+    // Register a callback within the defined scoped.
+    r = cubeb_register_device_collection_changed(ctx,
+                                                 static_cast<cubeb_device_type>(scopes[i]),
+                                                 device_collection_changed_callback,
+                                                 nullptr);
+    ASSERT_EQ(r, CUBEB_OK) << "Error registering device collection changed";
+
+    // Get an assertion fails when registering a callback within same scope twice.
+    ASSERT_DEBUG_DEATH(
+      cubeb_register_device_collection_changed(ctx,
```
achronop: I do not think we assert on Linux. We do on OSX, though. If we want to keep the assert we need to add it on Linux, otherwise we remove the assert check here and we remove it on OSX as well, which means we need to make it work there.

BTW it passes locally, I don't know why ...

test/test_devices.cpp
```diff
@@ -252,4 +252,145 @@ TEST(cubeb, stream_get_current_device)
...
+  int r = CUBEB_OK;
+
+  r = common_init(&ctx, "Cubeb duplex example with collection change");
+  ASSERT_EQ(r, CUBEB_OK) << "Error initializing cubeb library";
+
+  std::unique_ptr<cubeb, decltype(&cubeb_destroy)>
+    cleanup_cubeb_at_exit(ctx, cubeb_destroy);
+
+  r = cubeb_register_device_collection_changed(ctx,
+                                               static_cast<cubeb_device_type>(CUBEB_DEVICE_TYPE_UNKNOWN),
```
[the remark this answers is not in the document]
ChunMinChang: I'll remove it, thank!

test/test_devices.cpp
```diff
@@ -252,4 +252,145 @@ TEST(cubeb, stream_get_current_device)
...
+  std::unique_ptr<cubeb, decltype(&cubeb_destroy)>
+    cleanup_cubeb_at_exit(ctx, cubeb_destroy);
+
+  int scopes[3] = {
+    CUBEB_DEVICE_TYPE_INPUT,
+    CUBEB_DEVICE_TYPE_OUTPUT,
+    CUBEB_DEVICE_TYPE_INPUT | CUBEB_DEVICE_TYPE_OUTPUT
+  };
+
+  for (size_t i = 0 ; i < ARRAY_LENGTH(scopes) ; ++i) {
```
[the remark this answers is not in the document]
ChunMinChang: will do

.gitmodules
ChunMinChang: Oops! This is only for checking if the rust implementation works with C APIs. I will remove it.

test/test_devices.cpp
ChunMinChang: I am not familiar with the use case for `cubeb_register_device_collection_changed`. We are only support to fire one callback when device collection is changed. Personally I think it's better to ask user to unregister the existing callback if they have registered a callback before. Without any warnings or assertions, when users register two callbacks, they might think both registered callbacks will be fired on device-collection changing.

achronop: I don't have strong opinion either. We can go with the assert for now. In order to achieve that in pulse c-api you go in [1] and [2] and you add an assert that the callbacks are not NULL before you set it. Be careful to cover the unset case.

I am not sure how asserts work with Rust, if you need to change it there as well you do something similar to the above in [3] and [4]

[1] https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_pulse.c#L1544
[2] https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_pulse.c#L1548
[3] https://github.com/djg/cubeb-pulse-rs/blob/dev/src/backend/context.rs#L513
[4] https://github.com/djg/cubeb-pulse-rs/blob/dev/src/backend/context.rs#L517

achronop: Can you please add a test to unregistere twice. I believe it should be allowed and return `CUBEB_OK`

ChunMinChang: Thanks for the hints! I will add the tests to check if it's ok to unregister twice.

achronop:
My understanding is that I should wait for more commits on that. Please, let me know when you are ready.

achronop:
Is it ready or I am expecting more commits? I don't see anything about comment [1]. Are you planning to do it?

[1] https://github.com/kinetiknz/cubeb/pull/472#discussion_r230000025

ChunMinChang:
I've uploaded a [commit](https://github.com/kinetiknz/cubeb/pull/472/commits/3ff6eaf7bfd7fa32f387b4d775aa9c23175448c3) for _cubeb_pulse.c_ to add assertions to make sure that the callback can only be registered when there is no existing callback.

The commits to add assertions in _cubeb-pulse-rs_ will be uploaded on its own repo. The `ASSERT_DEATH` in the tests may need to be changed since the _Rust_ assertion works differently.

test/test_devices.cpp
```diff
@@ -252,4 +252,241 @@ TEST(cubeb, stream_get_current_device)
...
+
+  // Register a callback within the defined scoped.
+  r = cubeb_register_device_collection_changed(ctx,
+                                               CUBEB_DEVICE_TYPE_OUTPUT,
+                                               device_collection_changed_callback,
+                                               nullptr);
+  ASSERT_EQ(r, CUBEB_OK) << "Error registering device collection changed";
+
+  // Get an assertion fails when registering a callback within same scope twice.
+  ASSERT_DEATH(
```
ChunMinChang: This will be removed after Rust version of the AudioUnit backend is landed. Two reasons:
1. The `panic` in Rust is [impossible to be passed to be captured from C side][gtest].
2. There is a `Rust` mutex used in `cubeb_register_device_collection_changed`. Panicking while holding a Rust mutex will poison the mutex so it should be avoided (see [D29982][D29982]).

[gtest]: 
https://users.rust-lang.org/t/how-to-write-a-death-test-in-gtest-for-rust-apis/23575/5
[D29982]: https://phabricator.services.mozilla.com/D29982

kinetiknz:

ChunMinChang:
The added tests can only be used in the C backend, so it's ok to forget it for now. However, it's important to note that the behavior of the `cubeb_register_device_collection_changed` is different between the C backend and the Rust backend. 

The C backend may git the assertions while holding a mutex. In Rust, panicking(hitting assertions) while holding a Rust mutex will poison the mutex so it should be avoided.

I'd love to see we have a test to guarantee we have consistent behaviors in different backends, but these tests are outdated now.

PR #473, opened 2018-11-01
Potential null pointer in audiounit_set_device_info

ChunMinChang:
The `device_info * info` will be `nullptr` when `audiounit_set_device_info(stm, id, static_cast<io_side>(10))` is called (see [here](https://github.com/kinetiknz/cubeb/blob/9a7a55153e7f9b9e0036ab023909c7bc4a41688b/src/cubeb_audiounit.cpp#L748-L759)). 
 
```cpp
device_info * info = nullptr;
cubeb_device_type type = CUBEB_DEVICE_TYPE_UNKNOWN;

if (side == io_side::INPUT) {
  info = &stm->input_device;
  type = CUBEB_DEVICE_TYPE_INPUT;
} else if (side == io_side::OUTPUT) {
  info = &stm->output_device;
  type = CUBEB_DEVICE_TYPE_OUTPUT;
}
memset(info, 0, sizeof(device_info));
```

On the other hand, it is unable to translate them into Rust directly. The Rust version will be something like:

```rust
let info: &mut device_info;
let mut devtype = DeviceType::UNKNOWN;

if side == io_side::INPUT {
    info = &mut stm.input_device;
    type = DeviceType::INPUT;
} else if side == io_side::OUTPUT {
    info = &mut stm.output_device;
    type = DeviceType::OUTPUT;
}
*info = unsafe { mem::zeroed() };
```

It cannot compile since there will be an error like  `'info' may be a null reference`.

Since `audiounit_set_device_info` is either called for the input side or called for the output side, we can add an assertion to make sure it only used within either input or output scope, then replace *if-else-if* by *if-else* or *ternary operator*(`?` and `:`).

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -744,6 +744,7 @@ static int
 audiounit_set_device_info(cubeb_stream * stm, AudioDeviceID id, io_side side)
 {
   assert(stm);
+  assert(side == io_side::INPUT || side == io_side::OUTPUT);
 
   device_info * info = nullptr;
   cubeb_device_type type = CUBEB_DEVICE_TYPE_UNKNOWN;
@@ -751,7 +752,7 @@ audiounit_set_device_info(cubeb_stream * stm, AudioDeviceID id, io_side side)
   if (side == io_side::INPUT) {
     info = &stm->input_device;
     type = CUBEB_DEVICE_TYPE_INPUT;
-  } else if (side == io_side::OUTPUT) {
+  } else {
     info = &stm->output_device;
     type = CUBEB_DEVICE_TYPE_OUTPUT;
   }
@@ -760,7 +761,7 @@ audiounit_set_device_info(cubeb_stream * stm, AudioDeviceID id, io_side side)
 
   if (side == io_side::INPUT) {
     info->flags |= DEV_INPUT;
-  } else if (side == io_side::OUTPUT) {
+  } else {
     info->flags |= DEV_OUTPUT;
   }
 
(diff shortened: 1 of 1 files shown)


src/cubeb_audiounit.cpp
```diff
@@ -741,33 +741,22 @@ static int audiounit_stream_get_volume(cubeb_stream * stm, float * volume);
...
-  } else if (side == io_side::OUTPUT) {
-    info->flags |= DEV_OUTPUT;
-  }
+  info->flags |= (type == CUBEB_DEVICE_TYPE_INPUT) ? DEV_INPUT : DEV_OUTPUT;
 
   AudioDeviceID default_device_id = audiounit_get_default_device_id(type);
-  if (default_device_id == kAudioObjectUnknown) {
-    return CUBEB_ERROR;
-  }
+  assert(default_device_id != kAudioObjectUnknown);
```
achronop: This will crash when there is no device. See #358 . We need to return an error.

ChunMinChang: You're right. Thanks!

kinetiknz:

ChunMinChang:
I’ve fixed this in the Rust version. If the major backend on AudioUnit will be transferred to the Rust version, maybe it’s not really necessary to merge it. The code will be more consistent between two versions but I don’t have strong opinions.

kinetiknz:
I think we need a period where any fixes to bugs that exist in both backends are applied to both backends, since we need to support cubeb_audiounit at least until we ship the Rust version in stable Firefox (and possibly a little longer, depending on how many new bugs we see).  We used the same approach for cubeb_pulse and it worked fairly well.

Then at some point we call the Rust version stable and stop maintaining cubeb_audiounit.  We should (and could already) have done that for cubeb_pulse but nobody got around to it yet.

achronop:
Regarding OSX, I believe we need to maintain both backends for a period. It is crucial to keep them both healthy unless we see many errors and we decide that it is not worth the effort.

My guess is that cubeb_pulse has been around for so long because of the low maintenance burden on Linux backend, plus the convenience of using tests/tools directly.

BTW as we are moving to rust we must do something about repo integration.

ChunMinChang:
It's just a defect of the function, so it's safe to forget about it for cubeb_audiounit as long as we call the function properly.

PR #477, opened 2018-11-28
Don't use PLII matrices when downmixing.

### LINKED ISSUE #476: Crappy audio when downmixing from 7.1 to stereo
See bugzilla bug 1509875.

The remixer is using PLII matrices, for 7.1 it sounds pretty bad and cause audible clipping.

jyavenard:
PLII mix the opposed channel with the reversed phase. It doesn't sound good when used with a common stereo headset; like the left rear channel sounds too much like it's also coming from the right. Additionally, it clips significantly causing unwanted distortion. So we default to a simpler conversion instead: left goes to left only, same for the right channel.

(Part of the fix for BMO 1509875)

Fixes #476


achronop:
I checked it this morning, I am ok with that.

PR #480, opened 2018-12-20
Add a hint for building cubeb with cubeb-pulse-rs

ChunMinChang:
I ran into a problem on building _cubeb_ with _cubeb-pulse-rs_ today. I cloned a _cubeb-pulse-rs_ repo under _cubeb/src_, turned on `BUILD_RUST_LIBS` (by `option(BUILD_RUST_LIBS "Build rust backends" ON)`) and then ran a command to build. As a result, all _libpulse_ APIs are _undefined references_ on _linking_ stage while compiling. To link to those APIs, either _libpulse_ needs to be passed to linker via `target_link_libraries(... pulse)` (something like `-lpulse` for _gcc_), or `pulse-dlopen` feature needs to be turned on (by setting `pulse-dlopen` to default in _Cargo.toml_ under _cubeb/src/cubeb-pulse-rs_). I think it's better to add some hints for the configuration in case others run into the same issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,6 +242,8 @@ if(USE_PULSE_RUST)
     LOG_BUILD ON)
   add_dependencies(cubeb cubeb_pulse_rs)
   target_compile_definitions(cubeb PRIVATE USE_PULSE_RUST)
+  # Add `pulse`at the end of target_link_libraries if pulse-dlopen feature is
+  # not turned on. Otherwise, the libpulse APIs cannot be linked.
   target_link_libraries(cubeb PRIVATE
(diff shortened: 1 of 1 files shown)


achronop:
Are you testing on Linux? I remember I did build it successfully on Linux about 2 months ago. If you are on Linux I can check again.

ChunMinChang:
If linking to _libpulse_ can solve the issue, I don't think we need to force users to clone a whole _cubeb-rs_ repo. Especially when they have no need to change the _Rust_ interface. The static library generated by _cubeb-pulse-rs_ should import the dependencies for _cubeb-rs_ it needs.

Giving messages may be a good way to work on, I'll upload other commits for that.

achronop:
I am not great on rust dependencies so I will ask @kinetiknz to review this one.

PR #481, opened 2018-12-21
Test stream register device changed callback

ChunMinChang:
Creating a test to make sure `device_changed_callback` works as what we expected and it can also be used to check Rust's `device_changed_callback` implementation.


kinetiknz:
LGTM, thanks!

PR #483, opened 2019-01-07
audiounit: stop leaking after calling audiounit_create_device_from_hw_dev. Fixes #482

### LINKED ISSUE #482: Memory leak when audiounit_workaround_for_airpod is called
In `audiounit_workaround_for_airpod`[0], we allocate memory(`char` array) for `device_id`, `friendly_name`, and `vendor_name` in `input_device_info` and `output_device_info`[1] by `new` operators[2] (in `audiounit_strref_to_cstr_utf8` called from `audiounit_create_device_from_hwdev`), but we never deallocate those memory. We should call `delete[] *.device_id/friendly_name/vendor_name` to release those memory at the end of `audiounit_workaround_for_airpod`, like what we do in `audiounit_device_collection_destroy`[3].

[0]
https://github.com/kinetiknz/cubeb/blob/9a7a55153e7f9b9e0036ab023909c7bc4a41688b/src/cubeb_audiounit.cpp#L1839-L1845

[1]
https://github.com/kinetiknz/cubeb/blob/9a7a55153e7f9b9e0036ab023909c7bc4a41688b/src/cubeb_audiounit.cpp#L3241

https://github.com/kinetiknz/cubeb/blob/9a7a55153e7f9b9e0036ab023909c7bc4a41688b/src/cubeb_audiounit.cpp#L3269

https://github.com/kinetiknz/cubeb/blob/9a7a55153e7f9b9e0036ab023909c7bc4a41688b/src/cubeb_audiounit.cpp#L3284

[2]
https://github.com/kinetiknz/cubeb/blob/9a7a55153e7f9b9e0036ab023909c7bc4a41688b/src/cubeb_audiounit.cpp#L3121

[3]
https://github.com/kinetiknz/cubeb/blob/9a7a55153e7f9b9e0036ab023909c7bc4a41688b/src/cubeb_audiounit.cpp#L3386-L3388

achronop:
When `audiounit_create_device_from_hw_dev` is called in an internal method we need to manually delete the allocated memory. When the same method is called from the external API method `audiounit_enumerate_devices` we don't need that because delete will happen on `audiounit_collection_destroy`.

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -1834,6 +1834,7 @@ static void audiounit_get_available_samplerate(AudioObjectID devid, AudioObjectP
                                    uint32_t * min, uint32_t * max, uint32_t * def);
 static int
 audiounit_create_device_from_hwdev(cubeb_device_info * dev_info, AudioObjectID devid, cubeb_device_type type);
+static void audiounit_device_destroy(cubeb_device_info * device);
 
 static void
 audiounit_workaround_for_airpod(cubeb_stream * stm)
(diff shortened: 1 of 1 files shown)


kinetiknz:
LGTM, thanks!

PR #485, opened 2019-01-15
Fix example code using nonexistent output buffer.

jamen:


achronop:
We missed that ... Thank you very much!

PR #486, opened 2019-01-21
pulse: guess default layout when it is not configured (BMO 1518106).

achronop:
In PulseAudio when channel layout is not specified a default is being used. If default layout is not set `stream_init` will fail. On PA channel I was recommended to use a channel layout instead of relying on default because it is not guaranteed to be set. This patch will check if a default layout exists, if not a layout is set. The default layouts have been taken from Gecko's default layout in playback (link bellow):

https://searchfox.org/mozilla-central/rev/6e0e603f4852b8e571e5b8ae133e772b18b6016e/dom/media/AudioConfig.cpp#75

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -85,6 +85,7 @@
   X(pa_context_subscribe)                       \
   X(pa_mainloop_api_once)                       \
   X(pa_get_library_version)                     \
+  X(pa_channel_map_init_auto)                   \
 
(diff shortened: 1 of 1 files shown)


kinetiknz:
LGTM.  We also need to make this change in https://github.com/djg/cubeb-pulse-rs/, would you mind taking care of that please?

achronop:
Thanks, I'll take care the `cubeb-pulse-rs`, I have started the patch and I'll continue today.

PR #488, opened 2019-02-04
wasapi: implement device collection changed callback

achronop:
Implement device collection changed callback for wasapi backend. Similar to the other platforms:
* It takes separated callback for input and output device changes.
* if device type is `input | output` the same callback is for input and output changes.
* A second register without an unregister first, for the same device type, creates an assert failure.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -146,9 +146,15 @@ static std::unique_ptr<wchar_t const []> utf8_to_wstr(char const * str);
 
 }
 
+class wasapi_collection_notification_client;
+
 struct cubeb {
   cubeb_ops const * ops = &wasapi_ops;
   cubeb_strings * device_ids;
+  /* Device enumerator to get notifications when the
+     device collection change. */
+  com_ptr<IMMDeviceEnumerator> device_collection_enumerator;
+  com_ptr<wasapi_collection_notification_client> collection_notification_client;
 };
 
 class wasapi_endpoint_notification_client;
@@ -268,6 +274,166 @@ struct cubeb_stream {
   /* True when we've destroyed the stream. This pointer is leaked on stream
    * destruction if we could not join the thread. */
   std::atomic<std::atomic<bool>*> emergency_bailout;
+};
+
+class wasapi_collection_notification_client : public IMMNotificationClient
+{
+public:
+  /* The implementation of MSCOM was copied from MSDN. */
+  ULONG STDMETHODCALLTYPE
+  AddRef()
+  {
+    return InterlockedIncrement(&ref_count);
+  }
+
+  ULONG STDMETHODCALLTYPE
+  Release()
+  {
+    ULONG ulRef = InterlockedDecrement(&ref_count);
+    if (0 == ulRef) {
+      delete this;
+    }
+    return ulRef;
+  }
+
+  HRESULT STDMETHODCALLTYPE
+  QueryInterface(REFIID riid, VOID **ppvInterface)
+  {
+    if (__uuidof(IUnknown) == riid) {
+      AddRef();
+      *ppvInterface = (IUnknown*)this;
+    } else if (__uuidof(IMMNotificationClient) == riid) {
+      AddRef();
+      *ppvInterface = (IMMNotificationClient*)this;
+    } else {
+      *ppvInterface = NULL;
+      return E_NOINTERFACE;
+    }
+    return S_OK;
+  }
+
+  wasapi_collection_notification_client(cubeb * context, IMMDeviceEnumerator * dev_enum)
+    : ref_count(1)
+    , cubeb_context(context)
+    , enumerator(dev_enum)
+  {
+    XASSERT(cubeb_context);
+    XASSERT(enumerator);
+  }
+
+  virtual ~wasapi_collection_notification_client()
+  { }
+
+  HRESULT STDMETHODCALLTYPE
+  OnDefaultDeviceChanged(EDataFlow flow, ERole role, LPCWSTR device_id)
+  {
+    LOG("Audio device default changed, id = %S.", device_id);
+    return S_OK;
+  }
+
+  /* The remaining methods are not implemented, they simply log when called (if
+     log is enabled), for debugging. */
+  HRESULT STDMETHODCALLTYPE OnDeviceAdded(LPCWSTR device_id)
+  {
+    LOG("Audio device added.");
+    return S_OK;
+  };
+
+  HRESULT STDMETHODCALLTYPE OnDeviceRemoved(LPCWSTR device_id)
+  {
+    LOG("Audio device removed.");
+    return S_OK;
+  }
+
+  HRESULT STDMETHODCALLTYPE
+  OnDeviceStateChanged(LPCWSTR device_id, DWORD new_state)
+  {
+    XASSERT(output_collection_changed_callback || input_collection_changed_callback);
+    LOG("Audio device state changed, id = %S, state = %lu.", device_id, new_state);
+    if (new_state == DEVICE_STATE_ACTIVE ||
+        new_state == DEVICE_STATE_NOTPRESENT ||
+        new_state == DEVICE_STATE_UNPLUGGED) {
+      EDataFlow flow;
+      HRESULT hr = GetDataFlow(device_id, &flow);
+      if (FAILED(hr)) {
+        return hr;
+      }
+      auto_lock lock(callback_lock);
+      if (flow == eCapture && input_collection_changed_callback) {
+        input_collection_changed_callback(cubeb_context,
+          input_collection_changed_user_ptr);
+      }
+      if (flow == eRender && output_collection_changed_callback) {
+        output_collection_changed_callback(cubeb_context,
+          output_collection_changed_user_ptr);
+      }
+    }
+    return S_OK;
+  }
+
+  HRESULT STDMETHODCALLTYPE
+  OnPropertyValueChanged(LPCWSTR device_id, const PROPERTYKEY key)
+  {
+    //Audio device property value changed.
+    return S_OK;
+  }
+
+  void set_input_callback(cubeb_device_collection_changed_callback callback,
+                          void * user_ptr)
+  {
+    auto_lock lock(callback_lock);
+    XASSERT(!!input_collection_changed_callback != !!callback);
+    input_collection_changed_callback = callback;
+    input_collection_changed_user_ptr = user_ptr;
+  }
+  void set_output_callback(cubeb_device_collection_changed_callback callback,
+                           void * user_ptr)
+  {
+    auto_lock lock(callback_lock);
+    XASSERT(!!output_collection_changed_callback != !!callback);
+    output_collection_changed_callback = callback;
+    output_collection_changed_user_ptr = user_ptr;
+  }
+  bool has_input_or_output()
+  {
+    auto_lock lock(callback_lock);
+    return output_collection_changed_callback ||
+             input_collection_changed_callback;
+  }
+private:
+  HRESULT GetDataFlow(LPCWSTR device_id, EDataFlow * flow)
+  {
+    com_ptr<IMMDevice> device;
+    com_ptr<IMMEndpoint> endpoint;
+
+    HRESULT hr = enumerator->GetDevice(device_id, device.receive());
+    if (FAILED(hr)) {
+      LOG("Could not get device: %lx", hr);
+      return hr;
+    }
+
+    hr = device->QueryInterface(IID_PPV_ARGS(endpoint.receive()));
+    if (FAILED(hr)) {
+      LOG("Could not get endpoint: %lx", hr);
+      return hr;
+    }
+
+    return endpoint->GetDataFlow(flow);
+  }
+
+  /* refcount for this instance, necessary to implement MSCOM semantics. */
+  LONG ref_count;
+  IMMDeviceEnumerator * enumerator = nullptr;
+
+  cubeb * cubeb_context = nullptr;
+  /* Collection changed for input (captute) devices. */
+  cubeb_device_collection_changed_callback input_collection_changed_callback = nullptr;
+  void * input_collection_changed_user_ptr = nullptr;
+  /* Collection change for output (render) devices. */
+  cubeb_device_collection_changed_callback output_collection_changed_callback = nullptr;
+  void * output_collection_changed_user_ptr = nullptr;
+  /* Protect callbacks between client thread and notification thread. */
+  owned_critical_section callback_lock;
 };
 
 class wasapi_endpoint_notification_client : public IMMNotificationClient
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -270,6 +276,166 @@ struct cubeb_stream {
...
+  }
+
+  HRESULT STDMETHODCALLTYPE
+  OnPropertyValueChanged(LPCWSTR device_id, const PROPERTYKEY key)
+  {
+    //Audio device property value changed.
+    return S_OK;
+  }
+
+  void set_input_callback(cubeb_device_collection_changed_callback callback,
```
kinetiknz: Rather than allow the callbacks to be changed on a live notification client, I think it'd be better to simply create the notification client with the appropriate callbacks and destroy/recreate it if we need to change callbacks.  That allows removal of set_input_callback, set_output_callback, has_input_or_output, and the callback_lock which I think makes the code simpler and easier to follow.

I'm a bit concerned about calling the callbacks directly from the notification client.  We went out of our way to avoid doing anything in the wasapi_endpoint_notification_client except signalling an event.  if you've checked and are sure this is safe, that's fine.  Otherwise, it might be better to use a similar approach and have the render loop do the heavy lifting - that would also simplify where the callbacks run from, since we have control over the dispatching thread.  Also simplifies ownership, since we never have to worry about wasapi_collection_notification_client having a potential dangling pointer to a cubeb context.

achronop: Thank you for the comments, both are very good.

> Rather than allow the callbacks to be changed on a live notification client, I think it'd be better to simply create the notification client with the appropriate callbacks and destroy/recreate it if we need to change callbacks. That allows removal of set_input_callback, set_output_callback, has_input_or_output, and the callback_lock which I think makes the code simpler and easier to follow.

Recrating the notification client is not enough. We need to unregister first from the enumerator and re-register after the new notification client has been created. If that's not a problem then it will be cleaner indeed. In addition to that the callback pointers can be stored in the cubeb struct, which is already passed in the notification client.

kinetiknz: > Recrating the notification client is not enough. We need to unregister first from the enumerator and re-register after the new notification client has been created. If that's not a problem then it will be cleaner indeed. In addition to that the callback pointers can be stored in the cubeb struct, which is already passed in the notification client.

Right, we'd have to undo/redo everything, basically the same model as the existing notification client for the current stream (wasapi_endpoint_notification_client).

> Sorry, I was not aware that we allow only signaling an event from the notification client. I haven't done serious stress testing on that and since you are concerned about it I don't think it can be ignored. Discussing that requirement further, the problem with render loop is that it is starting to run after a `cubeb_stream start`. On the other hand the collection changed callback can be configured at any point even when a stream is not initialized. If I understand correctly, all the benefits above, will exist, if we use the same one render thread, used today to output audio and not if we create a new thread for collection changed purpose. In this case we need to change the life time of that thread, which now lives between a stream start and a stream stop (or a stream destroy if things go wrong). Also, we risk to call the collection changed callback from the high priority audio thread when a stream is configured and running in parallel. That would not be a risk if we had a separate thread for the needs of the collection changed callbacks. Hmm, I am little confused what would be the best approach here, would you like to tell me what you think?

I'm not 100% sure we need to have such strict requirements - the problem is that I don't remember.  I'm *assuming* the wasapi_endpoint_notification_client is designed that way for a reason, and if so it's to avoid doing anything complex from the unknown context in which the notification client runs.  It might be worth reinvestigating that issue and confirming what restrictions the OS has for the code running in the notification client - alternatively, checking some other robust implementation (Chromium?) to see what they do might be sufficient.

When I made my suggestion, I didn't consider the issue of the render thread only existing when the stream is active - as you said, that won't work for the device collection notification client.  Rather than change the lifetime of the per-stream render threads, I think it'd be fine to have a single separate thread on the cubeb context (created lazily if/when a device collection callback is first registered) to run the bulk of the device collection callback code (i.e. the user registered callbacks, the GetDataFlow queries, etc.), but I'm open to other suggestions to solve this.

achronop: > It might be worth reinvestigating that issue and confirming what restrictions the OS has for the code running in the notification client

I copy here the online doc about notification client for reference. At the last paragraph it mentions that it must be nonblocking and some other restrictions.

https://docs.microsoft.com/en-us/windows/desktop/api/mmdeviceapi/nn-mmdeviceapi-immnotificationclient

achronop: This is ready now to review. I found on other projects that they use a separate thread to deliver the notification callbacks. Also, the documentation from above states that we should not do anything blocking from the notification thread. I have implemented separate thread approach. Please let me know what you think when you have a chance. Thanks!

src/cubeb_wasapi.cpp
```diff
@@ -1052,6 +1290,48 @@ HRESULT get_endpoint(com_ptr<IMMDevice> & device, LPCWSTR devid)
...
 }
 
+HRESULT register_collection_notification_client(cubeb * context)
+{
+  HRESULT hr = CoCreateInstance(__uuidof(MMDeviceEnumerator),
+                                NULL, CLSCTX_INPROC_SERVER,
+                                IID_PPV_ARGS(context->device_collection_enumerator.receive()));
+  if (FAILED(hr)) {
+    LOG("Could not get device enumerator: %lx", hr);
+    context->device_collection_enumerator.reset();
```
kinetiknz: The reset shouldn't be necessary here - we don't do this for the stream notification client.

src/cubeb_wasapi.cpp
```diff
@@ -1052,6 +1290,48 @@ HRESULT get_endpoint(com_ptr<IMMDevice> & device, LPCWSTR devid)
...
+
+  return hr;
+}
+
+HRESULT unregister_collection_notification_client(cubeb * context)
+{
+  HRESULT hr = context->device_collection_enumerator->
+    UnregisterEndpointNotificationCallback(context->collection_notification_client.get());
+  if (FAILED(hr)) {
+    return hr;
```
kinetiknz: Note that the stream notification client deliberately releases the device enumerator and turns a failure here into a success.  I think Paul added that as a workaround for an issue a while back - let's not do the same here yet, but keep it in mind if we have issues later.

src/cubeb_wasapi.cpp
```diff
@@ -2368,6 +2648,80 @@ wasapi_device_collection_destroy(cubeb * /*ctx*/, cubeb_device_collection * coll
...
+wasapi_register_device_collection_changed(cubeb * context,
+                                          cubeb_device_type devtype,
+                                          cubeb_device_collection_changed_callback collection_changed_callback,
+                                          void * user_ptr)
+{
+  if (devtype == CUBEB_DEVICE_TYPE_UNKNOWN) {
+    return CUBEB_ERROR_INVALID_PARAMETER;
+  }
+
+  HRESULT hr;
```
kinetiknz: Move the declaration down to the first use, and redeclare it again as it is needed when out of scope.  Makes it a bit clearer where it's used.

src/cubeb_wasapi.cpp
```diff
@@ -1052,6 +1290,48 @@ HRESULT get_endpoint(com_ptr<IMMDevice> & device, LPCWSTR devid)
...
+
+  hr = context->device_collection_enumerator->RegisterEndpointNotificationCallback(
+                                                context->collection_notification_client.get());
+  if (FAILED(hr)) {
+    LOG("Could not register endpoint notification callback: %lx", hr);
+    context->collection_notification_client.reset();
+    context->device_collection_enumerator.reset();
+  }
+
+  context->monitor_notifications = std::make_unique<monitor_device_notifications>(context);
```
kinetiknz: This is failing to compile on the MinGW build because it's dated and make_unique is missing.  Simplest fix for now is to avoid using make_unique until our MinGW build is newer.

Also looks like some new warnings showing up - please check/fix these.

src/cubeb_wasapi.cpp
```diff
@@ -349,14 +587,14 @@ class wasapi_endpoint_notification_client : public IMMNotificationClient
   HRESULT STDMETHODCALLTYPE
   OnDeviceStateChanged(LPCWSTR device_id, DWORD new_state)
   {
-    LOG("Audio device state changed.");
+    //LOG("Audio device state changed.");
```
kinetiknz: Revert this and the change in OnPropertyValueChanged unless there's a reason to remove them.  If they're ambiguous compared to log messages from wasapi_collection_notification_client, I'd prefix the log messages in wasapi_collection_notification_client and wasapi_endpoint_notification_client with "collection:" and "endpoint:" to disambiguate.

src/cubeb_wasapi.cpp
```diff
@@ -270,6 +284,230 @@ struct cubeb_stream {
...
+    XASSERT(enumerator);
+  }
+
+  virtual ~wasapi_collection_notification_client()
+  { }
+
+  HRESULT STDMETHODCALLTYPE
+  OnDefaultDeviceChanged(EDataFlow flow, ERole role, LPCWSTR device_id)
+  {
+    //LOG("Audio device default changed, id = %S.", device_id);
```
kinetiknz: Uncomment, this might be useful logging?

src/cubeb_wasapi.cpp
```diff
@@ -270,6 +284,230 @@ struct cubeb_stream {
...
+      }
+      cubeb_context->monitor_notifications->notify(flow);
+    }
+    return S_OK;
+  }
+
+  HRESULT STDMETHODCALLTYPE
+  OnPropertyValueChanged(LPCWSTR device_id, const PROPERTYKEY key)
+  {
+    //Audio device property value changed.
```
kinetiknz: Uncomment and add LOG(), this might be useful logging?

src/cubeb_wasapi.cpp
```diff
@@ -270,6 +284,230 @@ struct cubeb_stream {
...
+      LOG("Could not get endpoint: %lx", hr);
+      return hr;
+    }
+
+    return endpoint->GetDataFlow(flow);
+  }
+
+  /* refcount for this instance, necessary to implement MSCOM semantics. */
+  LONG ref_count;
+  IMMDeviceEnumerator * enumerator = nullptr;
```
kinetiknz: Since we don't hold our own reference here, add a comment explaining that this has the same lifetime as the context->device_collection_enumerator.

src/cubeb_wasapi.cpp
```diff
@@ -270,6 +284,230 @@ struct cubeb_stream {
...
+public:
+  monitor_device_notifications(cubeb * context)
+  : cubeb_context(context)
+  {
+    create_thread();
+  }
+  ~monitor_device_notifications()
+  {
+    SetEvent(shutdown);
+    WaitForSingleObject(thread, 2000);
```
kinetiknz: I'd make this 5s to match the timeout in stop_and_join_render_thread.

src/cubeb_wasapi.cpp
```diff
@@ -270,6 +284,230 @@ struct cubeb_stream {
...
 };
 
+class monitor_device_notifications {
+public:
+  monitor_device_notifications(cubeb * context)
+  : cubeb_context(context)
+  {
+    create_thread();
+  }
+  ~monitor_device_notifications()
```
kinetiknz: Line break between each member function please.

src/cubeb_wasapi.cpp
```diff
@@ -270,6 +284,230 @@ struct cubeb_stream {
...
+  {
+    SetEvent(shutdown);
+    WaitForSingleObject(thread, 2000);
+    CloseHandle(thread);
+
+    CloseHandle(input_changed);
+    CloseHandle(output_changed);
+    CloseHandle(shutdown);
+  }
+  bool notify(EDataFlow flow)
```
kinetiknz: I don't think we can do much useful if this fails (and the current code doesn't check the result), so make it void and maybe just log if there's an error.

src/cubeb_wasapi.cpp
```diff
@@ -270,6 +284,230 @@ struct cubeb_stream {
...
+    if (!input_changed) {
+      return CUBEB_ERROR;
+    }
+
+    shutdown = CreateEvent(nullptr, 0, 0, nullptr);
+    if (!shutdown) {
+      return CUBEB_ERROR;
+    }
+
+    thread = CreateThread(nullptr, 0, thread_proc, this, 0, nullptr);
```
kinetiknz: Use _beginthreadex here instead, and start with a 256kB stack and STACK_SIZE_PARAM_IS_A_RESERVATION.  You'll need to update the signature of thread_proc to match what _beginthreadex expects (e.g. wasapi_stream_render_loop).

src/cubeb_wasapi.cpp
```diff
@@ -270,6 +284,230 @@ struct cubeb_stream {
...
+      return CUBEB_ERROR;
+    }
+
+    return CUBEB_OK;
+  }
+
+  HANDLE thread;
+  HANDLE output_changed;
+  HANDLE input_changed;
+  HANDLE shutdown;
```
kinetiknz: These can all be initialized to INVALID_HANDLE_VALUE.

src/cubeb_wasapi.cpp
```diff
@@ -270,6 +284,230 @@ struct cubeb_stream {
...
+    } // loop
+
+    return 0;
+  }
+
+  int create_thread()
+  {
+    output_changed = CreateEvent(nullptr, 0, 0, nullptr);
+    if (!output_changed) {
+      return CUBEB_ERROR;
```
kinetiknz: Either log these errors here and return void, or log it once where create_thread is called after checking the result.

src/cubeb_wasapi.cpp
```diff
@@ -270,6 +284,230 @@ struct cubeb_stream {
...
+  notification_thread_loop()
+  {
+    HANDLE wait_array[3] = {
+      input_changed,
+      output_changed,
+      shutdown,
+    };
+
+    while (true) {
+      Sleep(200);
```
kinetiknz: This sleep shouldn't necessary?  Please either remove it or explain why it's present and we'll find a better solution.

src/cubeb_wasapi.cpp
```diff
@@ -270,6 +284,230 @@ struct cubeb_stream {
...
+
+      DWORD wait_result = WaitForMultipleObjects(ARRAY_LENGTH(wait_array),
+                                                 wait_array,
+                                                 FALSE,
+                                                 INFINITE);
+      if (wait_result == WAIT_OBJECT_0) { //input changed
```
kinetiknz: Space between // and comment.

src/cubeb_wasapi.cpp
```diff
@@ -270,6 +284,230 @@ struct cubeb_stream {
...
+  static DWORD WINAPI
+  thread_proc(LPVOID args)
+  {
+    monitor_device_notifications * that =
+      static_cast<monitor_device_notifications*>(args);
+    return that->notification_thread_loop();
+  }
+
+  int
+  notification_thread_loop()
```
kinetiknz: We *might* need to initialize COM on this thread, it depends on what happens inside the user callbacks.  If we do need to do so, we can copy the auto_com from wasapi_stream_render_loop.

src/cubeb_wasapi.cpp
achronop: I'll keep an eye on it.

src/cubeb_wasapi.cpp
achronop: I am ok reverting this one. However, OnPropertyValueChanged() creates a lot of noise. On every device change creates more 50 log lines and makes it hard to follow the interesting part of the flow. I suggest to stop logging from that method in all notification clients.

src/cubeb_wasapi.cpp
achronop: Hmm that was a leftover from the initial patch that I did not pass the whole context in the notification client. Now the context is available I could access the `device_collection_enumerator` from there and remove that class member.

src/cubeb_wasapi.cpp
achronop: This is to allow some time for the system to update it's internal list of devices after a device change event. We want to be sure that the list is updated the moment we are triggering the callback. We have something similar in Gecko [1]. Also, in a fix (Bug 1522488) recently we increased the duration from 100 to 200ms due to reports. Looking at JUCE project they do the same (longer duration) for every platform [2] for the same reason. If you have different opinion I am happy to change it.

[1] https://searchfox.org/mozilla-central/source/dom/media/MediaManager.cpp#2198
[2] https://github.com/WeAreROLI/JUCE/blob/06c3674e510f991d88f881203aedd85a60fbced9/modules/juce_events/native/juce_win32_HiddenMessageWindow.h#L107

src/cubeb_wasapi.cpp
achronop: I initialized the COM library just in case.

achronop:
Thank you for the careful review. I have resolved the applied comments. I have provided explanations on the rest of them so I've left them open. Feel free to add new questions or comments if anything else has come up.

src/cubeb_wasapi.cpp
kinetiknz: Makes sense.

PR #489, opened 2019-02-04
cubeb-test: improvements for testing collection changed callback

achronop:
Some improvements I made while I was testing device collection changed. It has been enhanced to provide options for starting/stopping a stream, registering a new callback etc.

### DIFF
--- a/tools/cubeb-test.cpp
+++ b/tools/cubeb-test.cpp
@@ -221,9 +221,8 @@ bool cubeb_client::register_device_collection_changed(
   } else if (devtype & CUBEB_DEVICE_TYPE_INPUT) {
     callback = input_device_changed_callback_s;
   }
-  int r = 0;
-  r = cubeb_register_device_collection_changed(
-      context, devtype, callback, nullptr);
+  int r = cubeb_register_device_collection_changed(
+            context, devtype, callback, nullptr);
   if (r != CUBEB_OK) {
     return false;
   }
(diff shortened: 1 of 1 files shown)


tools/cubeb-test.cpp
```diff
@@ -205,19 +211,34 @@ void cubeb_client::user_state_cb(cubeb_stream* stm, void* user,
...
   return true;
 }
 
-void cubeb_client::device_collection_changed_callback(cubeb* context,
-                                                      void* user) {
-  fprintf(stderr, "device_collection_changed_callback\n");
+bool cubeb_client::unregister_device_collection_changed(
+    cubeb_device_type devtype) const {
+  int r = 0;
+  r = cubeb_register_device_collection_changed(
```
kinetiknz: Rather than init and then immediately overwrite, just do `int r = cubeb_register_device_collection_changed(...)`.  Same above for cubeb_client::register_device_collection_changed

tools/cubeb-test.cpp
```diff
@@ -227,82 +248,166 @@ enum play_mode {
...
+void print_help() {
+  const char * msg =
+    "p: start a initialized stream\n"
+    "s: stop a started stream\n"
+    "i: change device type to input\n"
+    "o: change device type to output\n"
+    "a: change device type to input and output\n"
+    "k: change device type to unknown\n"
+    "r: register device collection changed callback for the currect device type\n"
+    "u: unregister device collection changed callback for the currect device type\n"
```
kinetiknz: currect = current for these two lines?

achronop:
Thank you for the comments

PR #496, opened 2019-03-22
Make sure data_callback and state_callback are not null

ChunMinChang:
If `state_callback` and `data_callback` are null, we will fail to call `stm->data_callback` and `stm->data_callback`. We call them without checking if they are not nulls. It seems to me that they are no options, so we should make sure they are properly set.

### DIFF
--- a/src/cubeb.c
+++ b/src/cubeb.c
@@ -303,7 +303,7 @@ cubeb_stream_init(cubeb * context, cubeb_stream ** stream, char const * stream_n
 {
   int r;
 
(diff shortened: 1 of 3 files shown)


padenot:
We could do this in `cubeb.c` so that all backend could be checked, don't you think? It seems pretty fundamental, those should be mandatory.

kinetiknz:
test/test_device_changed_callback.cpp is failing because it passes nullptr for both callbacks to cubeb_stream_init.

PR #501, opened 2019-04-10
Patch from BMO#1531833

padenot:


padenot:
Really annoying to have a pref that is gecko only here.

kinetiknz:
> Really annoying to have a pref that is gecko only here.

CUBEB_STREAM_PREF_DISABLE_DEVICE_SWITCHING?  It was originally an experiment but let's merge the patch upstream since it looks like we're keeping the behavior.

PR #504, opened 2019-04-19
test: add option to get the posotion of a stream

achronop:
Add an option in the tool to get the position of the stream.


tools/cubeb-test.cpp
```diff
@@ -151,6 +152,16 @@ bool cubeb_client::stop_stream() const {
   return true;
 }
 
+uint64_t cubeb_client::get_stream_position() const {
+  uint64_t pos = 0;
+  int rv = cubeb_stream_get_position(stream, &pos);
+  if (rv != CUBEB_OK) {
+    fprintf(stderr, "Could not stop the stream\n");
```
kinetiknz: Wrong error message

PR #510, opened 2019-06-06
sunaudio (NetBSD/illumos) backend

niacat:
This adds support for the native audio API used in NetBSD and illumos (and apparently GNU HURD, through NetBSD's rump kernel).

The tests pass and it plays audio from YouTube and records from a USB microphone through WebRTC successfully.


achronop:
Can you please also test WebRTC services like appear.in or the test page https://mozilla.github.io/webrtc-landing/gum_test.html

niacat:
@achronop 

microphone selection

"Success!"

I also get sound waves when I speak on onlinemictest.

src/cubeb_sun.c
```diff
@@ -0,0 +1,740 @@
...
+    }
+#endif
+    if (i > 0) {
+      (void)snprintf(dev_friendly, sizeof(dev_friendly), "%s %s %s (%d)",
+                     hwname.name, hwname.version, hwname.config, i - 1);
+    } else {
+      (void)snprintf(dev_friendly, sizeof(dev_friendly), "%s %s %s (default)",
+                     hwname.name, hwname.version, hwname.config);
+    }
+    device.devid = (void *)(uintptr_t)i;
```
kinetiknz: devid needs to be unique (it will always refer to the same device) and stable (the lifetime of devid must exist longer than an enumerate/collection_destroy cycle - in some backends we tie the lifetime of devid to the context).

It looks like the sunaudio backend's devid achieves the latter property by virtue of being a simple integer index.  Can you confirm it also has the former property? If an audio device is removed and a different one attached, is it possible for the device name/index to be reused, or do unique devices appear as unique /dev/audioN entries?

src/cubeb_sun.c
```diff
@@ -0,0 +1,740 @@
...
+  struct audio_info r_info; /* info for the record fd */
+  cubeb_data_callback data_cb;
+  cubeb_state_callback state_cb;
+  int16_t * play_buf;
+  int16_t * record_buf;
+  float * f_play_buf;
+  float * f_record_buf;
+  char input_name[32];
+  char output_name[32];
+  uint64_t frames_written;
```
kinetiknz: The fields shared between threads (running, volume, frames_written) should be protected by a mutex, made atomic, or at least include a comment explaining why they are safe as-is (if so).

src/cubeb_sun.c
niacat: It's possible for the same /dev/audioX node to be reused for a different device if one connected earlier is disconnected.

I'm not sure how else to satisfy this property - maybe a hash of the device name and number? If a user has two of the same sound card connected, the only difference between them will be the device number.

niacat:
@kinetiknz All tests pass except cubeb.duplex_collection_change.

AFAIK this is because it's trying to use an optional callback that isn't implemented (register_device_collection_changed).

I'm happy to take time to keep maintaining this, btw.

src/cubeb_sun.c
kinetiknz: > I'm not sure how else to satisfy this property - maybe a hash of the device name and number? If a user has two of the same sound card connected, the only difference between them will be the device number.

Sounds like that might work.  You don't have to make this change before we merge, the current behaviour won't crash but might be surprising.

> /dev/audio is the "default" device, and is really a configurable symlink that can be changed to reference another device by the user. As a NetBSD user, I'd expect all running audio applications to ignore this change until restarted - programs that only output audio generally open /dev/audio at initialization and don't close it until exit.

Firefox tends to create new streams 1:1 with media activity (new video or new call creates a new stream; even muting a call or pausing/resuming a video will result in a new stream), so if the /dev/audio symlink changes Firefox will pick it up on the next stream open.  That seems okay to me for now.  Even if the original /dev/audio was cached on the context it'd pick up the new link when a new process is created (which is frequent with multi-process Firefox).

Most of the other backends try to follow the default device when it changes - does the sunaudio API provide a device change notification or some other way to detect this, maybe even watching the symlink for changes?

kinetiknz:
> All tests pass except cubeb.duplex_collection_change.
> 
> AFAIK this is because it's trying to use an optional callback that isn't implemented (register_device_collection_changed).

It looks like that test should treat CUBEB_ERROR_NOT_SUPPORTED from cubeb_register_device_collection_changed as a passing case.

> I'm happy to take time to keep maintaining this, btw.

Awesome, thank you!  I'll go ahead and merge this now.  Thanks again for all your work.

src/cubeb_sun.c
niacat: >Most of the other backends try to follow the default device when it changes - does the sunaudio API provide a device change notification or some other way to detect this, maybe even watching the symlink for changes?

There's a way to scan for changes in the available devices - re-iterating through the available device nodes and using the AUDIO_GETDEV ioctl to identify each of them. Otherwise, there's no audio-specific API for notifying of changes, unless you count read/write suddenly failing due to disconnection of the active device.

kinetiknz:
> It looks like that test should treat CUBEB_ERROR_NOT_SUPPORTED from cubeb_register_device_collection_changed as a passing case.

Fixed by https://github.com/kinetiknz/cubeb/pull/584

PR #512, opened 2019-06-10
Fix tests warnings

ChunMinChang:
Just some warnings fix about types and unnecessary lambda captures.

### DIFF
--- a/test/test_ring_buffer.cpp
+++ b/test/test_ring_buffer.cpp
@@ -156,13 +156,13 @@ void basic_api_test(T& ring)
 }
 
 void test_reset_api() {
-	const size_t RING_BUFFER_SIZE = 128;
-	const size_t ENQUEUE_SIZE = RING_BUFFER_SIZE / 2;
+	const size_t ring_buffer_size = 128;
+	const size_t enqueue_size = ring_buffer_size / 2;
 
-	lock_free_queue<float> ring(RING_BUFFER_SIZE);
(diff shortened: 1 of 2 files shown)


tools/cubeb-test.cpp
```diff
@@ -332,7 +332,7 @@ bool choose_action(const cubeb_client& cl, operation_data * op, char c) {
     }
   } else if (c == 'c') {
     uint64_t pos = cl.get_stream_position();
-    fprintf(stderr, "stream position %lu\n", pos);
+    fprintf(stderr, "stream position %llu\n", pos);
```
kinetiknz: This will now warn on platforms where uint64_t is an unsigned long rather than unsigned long long.  I think PRIu64 works on the modern versions of Visual Studio we use now, so we should be able to use that.

test/test_ring_buffer.cpp
```diff
@@ -160,7 +160,7 @@ void test_reset_api() {
 	const size_t ENQUEUE_SIZE = RING_BUFFER_SIZE / 2;
 
 	lock_free_queue<float> ring(RING_BUFFER_SIZE);
-	std::thread t([&ring, ENQUEUE_SIZE] {
+	std::thread t([&ring] {
```
kinetiknz: The removal of ENQUEUE_SIZE here and below seems to have broken the build in MSVC: https://ci.appveyor.com/project/kinetiknz/cubeb/builds/25178084

kinetiknz:
Thanks!

PR #519, opened 2019-07-06
-Wimplicit-int-conversion warning fixes

kinetiknz:
A couple of simple fixes for -Wimplicit-int-conversion (available in Clang 8.0+).  ALSA, JACK, and AudioUnit backends were all clean.  Didn't test other build configurations.

cubeb_mixer.cpp has a bunch of warnings for this also, but it might be more involved to clean that up as it looks like it might be better to review the types used rather than just add checks and casts.

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -573,7 +573,9 @@ layout_to_channel_map(cubeb_channel_layout layout, pa_channel_map * cm)
     }
     channelMap = channelMap >> 1;
   }
-  cm->channels = cubeb_channel_layout_nb_channels(layout);
+  unsigned int channels_from_layout = cubeb_channel_layout_nb_channels(layout);
+  assert(channels_from_layout <= UINT8_MAX);
+  cm->channels = (uint8_t) channels_from_layout;
 }
 
 static void pulse_context_destroy(cubeb * ctx);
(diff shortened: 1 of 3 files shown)


src/cubeb_pulse.c
```diff
@@ -826,7 +828,9 @@ create_pa_stream(cubeb_stream * stm,
   if (ss.format == PA_SAMPLE_INVALID)
     return CUBEB_ERROR_INVALID_FORMAT;
   ss.rate = stream_params->rate;
-  ss.channels = stream_params->channels;
+  if (stream_params->channels > UINT8_MAX)
+    return CUBEB_ERROR_INVALID_FORMAT;
```
achronop: It looks good, maybe you could make that check more general, and add it in `validate_stream_params` [1] and simply assert here.

[1] https://github.com/kinetiknz/cubeb/blob/master/src/cubeb.c#L77

kinetiknz: We used to limit the channel count to 8 in validate_stream_params.  That check was completely removed in https://github.com/kinetiknz/cubeb/pull/433.  I'm fine with adding a (higher) channel limit back there, but I'll leave this check in cubeb_pulse since that's where we actually narrow the type.

PR #520, opened 2019-07-10
Add a performance tester

ChunMinChang:
Add a performance tester to count time of opening then closing the streams in parallel.
r? @achronop


kinetiknz:
What's the rationale for this particular benchmark?

ChunMinChang:
> What's the rationale for this particular benchmark?

I'd like to compare the performance between the Rust version and the C version of AudioUnit backend to make sure the Rust version can perform better.

kinetiknz:
I'm not convinced this is a realistic benchmark (how many concurrent streams are we *really* opening in a typical or even heavily loaded Gecko?) or that the test needs to be part of the main libcubeb repository at this stage.

I'm not against adding benchmarking to the repository in general, but the tests and goals need to be discussed and planned out first.  We may also need to use a set of reference devices as there may be significant variation in behaviour depending on the underlying audio (and other) hardware, OS version, etc.

Also benchmarking code should use something like https://github.com/google/benchmark rather than trying to reimplement the measurement code from scratch.

tools/cubeb-performance.cpp
```diff
@@ -0,0 +1,301 @@
...
+  for (size_t i = 0; i < number_of_streams; ++i) {
+    ids[i] = static_cast<stream_id>(i + 1);
+  }
+
+  unique_ptr<cubeb_stream* []> streams(new cubeb_stream*[number_of_streams]());
+  for (size_t i = 0; i < number_of_streams; ++i) {
+    streams[i] = nullptr;
+  }
+
+  auto start = std::chrono::high_resolution_clock::now();
```
achronop: I don't like this one here. The timer starts counting before you create the threads. Also, later you stop counting after joining all the threads. I would expect to measure closer to the cubeb_init/destroy. Probably that's why you need to truncate the max/min.

ChunMinChang:
ok, maybe it's better to move the discussion to issues instead of pull-request right now and define what it needs to be done first.

achronop:
Oh, you have closed it already, nevermind.

ChunMinChang:
> Oh, you have closed it already, nevermind.

Oh, sorry for the trouble!

kinetiknz:
> ok, maybe it's better to move the discussion to issues instead of pull-request right now and define what it needs to be done first.

This sounds good, thanks.

BTW - I had heard the Rust AudioUnit backend had quicker stream handling, that is great news.

PR #523, opened 2019-07-15
Fix output latency reporting on OSX, and add a round trip latency testing mode to the tool

padenot:
Also lower the minimum latency, we'll see in gecko whether it can work.

### DIFF
--- a/tools/cubeb-test.cpp
+++ b/tools/cubeb-test.cpp
@@ -17,7 +17,8 @@
 
 // Default values if none specified
 #define DEFAULT_RATE 44100
-#define DEFAULT_CHANNELS 2
+#define DEFAULT_OUTPUT_CHANNELS 2
+#define DEFAULT_INPUT_CHANNELS 1
 
 static const char* state_to_string(cubeb_state state) {
   switch (state) {
@@ -53,7 +54,10 @@ public:
   bool destroy_stream() const;
   bool destroy();
   bool activate_log(cubeb_log_level log_level) const;
+  void set_latency_testing(bool on);
+  void set_latency_frames(uint32_t latency_frames);
   uint64_t get_stream_position() const;
+  uint32_t get_stream_latency() const;
 
   long user_data_cb(cubeb_stream* stm, void* user, const void* input_buffer,
                     void* output_buffer, long nframes);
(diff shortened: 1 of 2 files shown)


padenot:
@ChunMinChang, if you want to have a look. I'll do a PR to the rust version as well.

padenot:
This now reports good latency figures, cross-tested by doing external measurements by feeding back an earbud playing clicks into a mic and checking the duration between the impulses in the feedback signal.

tools/cubeb-test.cpp
```diff
@@ -412,13 +488,18 @@ int main(int argc, char* argv[]) {
...
     }
   } else {
-    if (op.pm == PLAYBACK || op.pm == DUPLEX) {
+    if (op.pm == PLAYBACK || op.pm == DUPLEX || op.pm == LATENCY_TESTING) {
       cl.output_params = {CUBEB_SAMPLE_FLOAT32NE, op.rate, DEFAULT_CHANNELS,
-                          CUBEB_LAYOUT_UNDEFINED, CUBEB_STREAM_PREF_NONE};
+                          CUBEB_LAYOUT_STEREO, CUBEB_STREAM_PREF_NONE};
+    }
+    if (op.pm == RECORD || op.pm == DUPLEX || op.pm == LATENCY_TESTING) {
+      cl.input_params = {CUBEB_SAMPLE_FLOAT32NE, op.rate, 1, CUBEB_LAYOUT_UNDEFINED, CUBEB_STREAM_PREF_NONE};
```
achronop: You could make a #define with the input channels. Since it is different from the default channels we need to see that in the same place that the other channels are.

tools/cubeb-test.cpp
```diff
@@ -124,11 +130,34 @@ bool cubeb_client::init_stream() {
   _rate = has_output() ? output_params.rate : input_params.rate;
   _channels = has_output() ? output_params.channels : input_params.channels;
 
-  int rv =
+  uint32_t rate = 0;
+  int rv = cubeb_get_preferred_sample_rate(context, &rate);
```
achronop: Why do you need the preferred rate here and not the user configured rate?

padenot: Just a mistake, I've removed it.

PR #529, opened 2019-08-16
Rework how drain stop the stream

achronop:
This improves how we stop the stream when it is drained. Also, it addresses BMO 1572273. Unfortunately, I cannot verify that without the specific device mentioned there. @ChunMinChang can you please test it with your device? If the test passes we can land this one and skip #528 .

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -498,8 +498,8 @@ audiounit_input_callback(void * user_ptr,
   assert(stm->input_unit != NULL);
   assert(AU_IN_BUS == bus);
 
-  if (stm->shutdown) {
-    ALOG("(%p) input shutdown", stm);
+  if (stm->shutdown || stm->reinit_pending) {
+    ALOG("(%p) input shutdown or reinit exit callback early.", stm);
     return noErr;
   }
 
@@ -605,8 +605,8 @@ audiounit_output_callback(void * user_ptr,
   long input_frames = 0;
   void * output_buffer = NULL, * input_buffer = NULL;
 
-  if (stm->shutdown) {
-    ALOG("(%p) output shutdown.", stm);
+  if (stm->shutdown || stm->reinit_pending) {
+    ALOG("(%p) output shutdown or reinit exit callback early.", stm);
     audiounit_make_silent(&outBufferList->mBuffers[0]);
     return noErr;
   }
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -611,18 +611,6 @@ audiounit_output_callback(void * user_ptr,
     return noErr;
   }
 
-  if (stm->draining) {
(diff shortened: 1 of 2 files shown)


achronop:
Try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=9b01e8e36269b5286913585d4be02e295bafe1f8

kinetiknz:
From memory, we trigger the drain logic on the next callback to allow time for the buffer we just wrote to play out.  With this change, I think we'll stop too early.  It's probably quite hard to notice since we'd miss less than 11ms (default latency, could be lower) of audio in the typical case.  It might be worth verifying with a higher latency stream/device (Bluetooth) and a short sound effect.

ChunMinChang:
I don't understand the problem the changes would solve here. If it's a solution for a different problem from #528, the changes can be tested without my devices. If the changes are for the same problem, I'll appreciate if you can state your opinions in the review comments.

achronop:
> From memory, we trigger the drain logic on the next callback to allow time for the buffer we just wrote to play out. With this change, I think we'll stop too early.

I did an experiment that the stream was drained on the first callback. This produced a very small (in terms of duration) sound effect. I guess that was the one filled buffer. I could try to hack the latency to something larger and try again.

Can you recall if you had a specific problem that made you call the stop on the next callback or it was a "better safe than sorry" decision?

achronop:
> I don't understand the problem the changes would solve here. If it's a solution for a different problem from #528, the changes can be tested without my devices. If the changes are for the same problem, I'll appreciate if you can state your opinions in the review comments.

The changes here include a fix for the problem in #528. I thought a patch would be a faster explanation of my idea. Sorry about that I will comment on #528. Nevertheless, I would appreciate if you can test it with your device and tell me what you see. Thank you in advance.

achronop:
> I could try to hack the latency to something larger and try again.

I did an experiment similar to the above with a modified version of the backend that allowed big latency. Again, the stream drained at first callback. I verified that audio data of length equal to latency is written in the output. I started with a latency of 441 and ended with a latency of 4410 (for a rate of 44100). Further increase of the latency had no effect.

kinetiknz:
Thanks for testing! Maybe there's no issue, and I don't remember any other details now.  https://lists.apple.com/archives/coreaudio-api/2005/Dec/msg00049.html seems to suggest it works as you've observed, so I assume we waited an extra callback cycle just to be safe.

kinetiknz:
Also https://lists.apple.com/archives/coreaudio-api/2005/Dec/msg00055.html, in particular:

>  When AudioDeviceStop() is called on the IO thread, things are different. The HAL will not call the IOProc that was stopped again after the AudioDeviceStop() call completes, _but the buffer provided by that IOProc will get played_ (and nothing after that from that IOProc).

achronop:
Thanks for the links! That was a very interesting thread!

PR #530, opened 2019-08-16
WASAPI: add support for IAudioClient3 and lower latencies on Windows 10

ligfx:
Added code to get an `IAudioClient3` object, if possible, and then use `InitializeSharedAudioStream` when needed to get lower latencies.

WASAPI has some weird quirks which mean we can't always get the exact latency. It runs at a single, global (to the entire system) latency value. This means that if another client has already requested a specific latency, we have to match that. Additionally, requesting a latency value will also dynamically adjust every `IAudioClient` stream on the system to match.

All tests pass on Windows 10, and I tried it on a real-world project (Dolphin) as well. I'm not sure if there's anything special that needs to be done with duplex/loopback streams.

Quick overview:
- Tries to get an `IAudioClient3` object if possible, and uses `InitializeSharedAudioStream` to request streams with lower latencies.
- Copies the definition of `IAudioClient3` into the source, to support compiling against older SDKs. If I understand how COM works, this also allows access to `IAudioClient3` regardless of what runtime we link against. But not tested.
- `wasapi_get_min_latency` returns the exclusive-mode latency on Windows 10. This seems to always match the minimum latency reported by IAudioClient3 (a quick Google search shows others coming to the same conclusion), and simplifies the code.
- Includes a small fix for `hns_to_frames`, which can be off-by-one in certain cases.
- All tests pass on my system

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -33,6 +33,54 @@
 #include "cubeb_strings.h"
 #include "cubeb_utils.h"
 
+// Windows 10 exposes the IAudioClient3 interface to create low-latency streams.
+// Copy the interface definition from audioclient.h here to make the code simpler
+// and so that we can still access IAudioClient3 via COM if cubeb was compiled
+// against an older SDK.
+#ifndef __IAudioClient3_INTERFACE_DEFINED__
+#define __IAudioClient3_INTERFACE_DEFINED__
+MIDL_INTERFACE("7ED4EE07-8E67-4CD4-8C1A-2B7A5987AD42")
+IAudioClient3 : public IAudioClient
+{
+public:
+    virtual HRESULT STDMETHODCALLTYPE GetSharedModeEnginePeriod(
+        /* [annotation][in] */
+        _In_  const WAVEFORMATEX *pFormat,
+        /* [annotation][out] */
+        _Out_  UINT32 *pDefaultPeriodInFrames,
+        /* [annotation][out] */
+        _Out_  UINT32 *pFundamentalPeriodInFrames,
+        /* [annotation][out] */
+        _Out_  UINT32 *pMinPeriodInFrames,
+        /* [annotation][out] */
+        _Out_  UINT32 *pMaxPeriodInFrames) = 0;
+
+    virtual HRESULT STDMETHODCALLTYPE GetCurrentSharedModeEnginePeriod(
+        /* [unique][annotation][out] */
+        _Out_  WAVEFORMATEX **ppFormat,
+        /* [annotation][out] */
+        _Out_  UINT32 *pCurrentPeriodInFrames) = 0;
+
+    virtual HRESULT STDMETHODCALLTYPE InitializeSharedAudioStream(
+        /* [annotation][in] */
+        _In_  DWORD StreamFlags,
+        /* [annotation][in] */
+        _In_  UINT32 PeriodInFrames,
+        /* [annotation][in] */
+        _In_  const WAVEFORMATEX *pFormat,
+        /* [annotation][in] */
+        _In_opt_  LPCGUID AudioSessionGuid) = 0;
+};
+#ifdef __CRT_UUID_DECL
+// Required for MinGW
+__CRT_UUID_DECL(IAudioClient3, 0x7ED4EE07, 0x8E67, 0x4CD4, 0x8C, 0x1A, 0x2B, 0x7A, 0x59, 0x87, 0xAD, 0x42)
+#endif
+#endif
+// Copied from audioclient.h in the Windows 10 SDK
+#ifndef AUDCLNT_E_ENGINE_PERIODICITY_LOCKED
+#define AUDCLNT_E_ENGINE_PERIODICITY_LOCKED    AUDCLNT_ERR(0x028)
+#endif
+
 #ifndef PKEY_Device_FriendlyName
 DEFINE_PROPERTYKEY(PKEY_Device_FriendlyName,    0xa45c254e, 0xdf1c, 0x4efd, 0x80, 0x20, 0x67, 0xd1, 0x46, 0xa8, 0x50, 0xe0, 14);    // DEVPROP_TYPE_STRING
 #endif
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -1720,6 +1761,102 @@ handle_channel_layout(cubeb_stream * stm,  EDataFlow direction, com_heap_ptr<WAV
...
+  if (hr != AUDCLNT_E_ENGINE_PERIODICITY_LOCKED) {
+    LOG("Could not initialize shared stream with IAudioClient3: error: %lx", hr);
+    return false;
+  }
+
+  LOG("Got AUDCLNT_E_ENGINE_PERIODICITY_LOCKED, adjusting latency request");
+
+  uint32_t current_period = 0;
+  WAVEFORMATEX* current_format;
+  hr = audio_client3->GetCurrentSharedModeEnginePeriod(&current_format, &current_period);
```
kinetiknz: I guess you can't get away with passing a nullptr for current_format if you don't need the WAVEFORMATEX? The docs don't mention it, so I assume not.

src/cubeb_wasapi.cpp
```diff
@@ -1720,6 +1761,102 @@ handle_channel_layout(cubeb_stream * stm,  EDataFlow direction, com_heap_ptr<WAV
   }
 }
 
+static
+bool initialize_iaudioclient3(com_ptr<IAudioClient> & audio_client,
```
kinetiknz:
```
static bool
initialize_iaudioclient3
```

(I know the code isn't super consistent about this formatting, sorry)

And the indentation looks a bit off for the parameters.

src/cubeb_wasapi.cpp
```diff
@@ -1837,12 +1980,17 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
...
   }
 
-  hr = audio_client->Initialize(AUDCLNT_SHAREMODE_SHARED,
-                                flags,
-                                frames_to_hns(stm, stm->latency),
-                                0,
-                                mix_format.get(),
-                                NULL);
+  if (initialize_iaudioclient3(audio_client, stm, mix_format, flags, direction))
+  {
```
kinetiknz: Move brace to end of previous line

src/cubeb_wasapi.cpp
```diff
@@ -1837,12 +1980,17 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
...
+    LOG("Initialized with IAudioClient3");
+  } else {
+    hr = audio_client->Initialize(AUDCLNT_SHAREMODE_SHARED,
+                                  flags | AUDCLNT_STREAMFLAGS_NOPERSIST,
```
kinetiknz: Why don't we pass NOPERSIST for the IAC3 path?

src/cubeb_wasapi.cpp
```diff
@@ -1720,6 +1761,102 @@ handle_channel_layout(cubeb_stream * stm,  EDataFlow direction, com_heap_ptr<WAV
...
+  HRESULT hr;
+  uint32_t default_period = 0, fundamental_period = 0, min_period = 0, max_period = 0;
+  hr = audio_client3->GetSharedModeEnginePeriod(mix_format.get(), &default_period, &fundamental_period, &min_period, &max_period);
+  if (FAILED(hr)) {
+    LOG("Could not get shared mode engine period: error: %lx", hr);
+    return false;
+  }
+  uint32_t requested_latency = stm->latency;
+  if (requested_latency >= default_period)
+  {
```
kinetiknz: Move brace to end of previous line

src/cubeb_wasapi.cpp
```diff
@@ -1720,6 +1761,102 @@ handle_channel_layout(cubeb_stream * stm,  EDataFlow direction, com_heap_ptr<WAV
...
+  if (flags & AUDCLNT_STREAMFLAGS_LOOPBACK) {
+    // IAudioClient3 doesn't work with loopback streams, and will return error
+    // 88890021: AUDCLNT_E_INVALID_STREAM_FLAG
+    LOG("Audio stream is loopback, not using IAudioClient3");
+    return false;
+  }
+
+  // Some people have reported glitches with capture streams:
+  // http://blog.nirbheek.in/2018/03/low-latency-audio-on-windows-with.html
+  // TODO: Should those be forced to use IAudioClient?
```
kinetiknz: We'll need to test this.  It's probably best to make it bail out for capture streams for now, and we can experiment with them later.

src/cubeb_wasapi.cpp
```diff
@@ -33,6 +33,42 @@
 #include "cubeb_strings.h"
 #include "cubeb_utils.h"
 
+#ifndef _WIN32_WINNT_WIN10
```
kinetiknz: Add a comment to make it more explicit why we define this please.

src/cubeb_wasapi.cpp
ligfx: Yeah, unfortunately if you pass a nullptr you get back E_POINTER.

There's a possible situation where another IAudioClient3 stream has not only locked the latency, but also changed and locked the stream format (AUDCLNT_E_ENGINE_FORMAT_LOCKED). In that case the program needs to match both the latency and the format, which I assume is the reason for this restriction on GetCurrentSharedModeEnginePeriod. Handling this adds more complexity -- you can no longer tell if a format will work until you try initializing with it -- so I didn't implement it.

src/cubeb_wasapi.cpp
ligfx: IAudioClient3 doesn't support NOPERSIST, for whatever reason. I've changed this into a negation inside `initialize_iaudioclient3` and added a comment to make it more clear.

src/cubeb_wasapi.cpp
ligfx: Done! I assume most clients interested in low latency would use it for outputs, anyways.

src/cubeb_wasapi.cpp
ligfx: Done!

padenot:
> On the topic of testing: it occurs to me that there might be clients out there with hard-coded latency requests that would trigger the IAudioClient3 code path. If they aren't intending to get a low-latency stream but just happen to be slightly lower than WASAPI's default, they'll still lock the system at that latency. This would be 1) fine but 2) prevent other clients from accessing even lower latencies / unnecessarily increase CPU usage / power consumption. How does Firefox make latency requests to cubeb?

Firefox on Windows requests two types of latency:
- When playing back a video, it's requesting an arbitrary 100ms of latency.
- When doing things that are more real-time (WebRTC, Web Audio API, and in general anything dealing with `MediaStream` in JavaScript), it's querying the lowest available on the platform (via `cubeb_get_min_latency`).

We could therefore differentiate between scenarios where low latency is important and scenario where latency is not important (however knowing the exact latency is important, for A/V sync).

padenot:
(btw thank you very much for doing this, it's going to help a ton).

kinetiknz:
Thanks, the changes look good.

Tested locally within Firefox (and a [Try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=997d2c9d83d1208435e646a75caff86ca3689932)).  Normal media playback uses IAC as expected.  Web Audio tries to use IAC3 and succeeds, giving us 10ms:

```
cubeb_wasapi.cpp:1659: default device period: 100000, minimum device period: 30000
cubeb_wasapi.cpp:1672: Minimum latency in frames: 133
[...]
cubeb_wasapi.cpp:1990: Setup requested=[f=2 r=44100 c=2 l=0] mix=[f=2 r=44100 c=2 l=3]
cubeb_wasapi.cpp:1831: Got shared mode engine period: default=441 fundamental=441 min=441 max=441
cubeb_wasapi.cpp:1839: Requested latency 133 was adjusted to 441
cubeb_wasapi.cpp:2003: Initialized with IAudioClient3
```

Issue #527, opened 2019-08-08
Include resampler latency in cubeb_stream_get_latency?


mdsitton:
From what i can tell cubeb_resampler_latency() is not currently used anywhere in the code. For my usage of the library i need to be aware of all sources of latency, and in some cases the resampler is one of those sources.

Having the latency induced by the resampler automatically included into the value returned by cubeb_stream_get_latency() would make things much simpler since the resampler doesn't seemed to be used on all platform implementations.

achronop:
Yeah, that makes sense and it is easy to add it in `cubeb_stream_get_latency()`. Windows and OSX that need it most. The Android backend makes use of the resampler but it is usually configured in a way that it is not used unless you are doing extreme things with it.

mdsitton:
I could probably work on making a pull request myself for this if you guys don't mind?

kinetiknz:
That'd be great, thank you!

mdsitton:
After investigating this change somewhat I've come across a deeper issue with the cubeb_stream_get_latency for wasapi atleast. IAudioClient::GetStreamLatency() seems to always return 0 on windows 10?
I've not found any documentation on this anywhere, other than like a single person complaining about it on the msdn forums with no response a couple years back. I've also tested #530 thinking maybe something with IAudioClient3 would fix that but it doesn't.

Which after further digging i came across using the IAudioClock interface which is actually used already in the implementation of wasapi_stream_get_position(), with current_stream_delay()

So if no one object's here i'm planning on replacing the implementation of wasapi_stream_get_latency with something based on IAudioClock, along with adding any resampler frames. Should be able to get this up tomorrow.

ligfx:
@mdsitton yep, I can reproduce that issue with GetStreamLatency on Windows 10 (always returns 0). If IAudioClient3 is being used (anywhere on the system) then GetCurrentSharedModeEnginePeriod will return the current latency. Otherwise it just returns the default latency (about 10ms) and not the size of the stream's buffer.

PR #531, opened 2019-08-17
Compensate for resampler latency where it is utilized

mdsitton:
Work on #527

### DIFF
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -242,6 +242,7 @@ struct cubeb_stream {
   atomic<uint32_t> current_latency_frames{ 0 };
   atomic<uint32_t> total_output_latency_frames { 0 };
   unique_ptr<cubeb_resampler, decltype(&cubeb_resampler_destroy)> resampler;
+  float resample_ratio;
   /* This is true if a device change callback is currently running.  */
   atomic<bool> switching_device{ false };
   atomic<bool> buffer_size_change_state{ false };
@@ -2681,6 +2682,8 @@ audiounit_setup_stream(cubeb_stream * stm)
     input_unconverted_params.rate = stm->input_hw_rate;
   }
 
+  stm->resample_ratio = static_cast<float>(stm->output_stream_params.rate) / target_sample_rate;
+
   /* Create resampler. Output params are unchanged
    * because we do not need conversion on the output. */
   stm->resampler.reset(cubeb_resampler_create(stm,
@@ -2969,7 +2972,8 @@ static int
 audiounit_stream_get_position(cubeb_stream * stm, uint64_t * position)
 {
   assert(stm);
-  uint32_t latency_frames = stm->total_output_latency_frames + cubeb_resampler_latency(stm->resampler.get());
+  uint32_t resample_latency = static_cast<uint32_t>(ceil(cubeb_resampler_latency(stm->resampler.get()) * stm->resample_ratio));
+  uint32_t latency_frames = stm->total_output_latency_frames + resample_latency;
 
   if (latency_frames > stm->frames_played) {
     *position = 0;
@@ -2986,7 +2990,9 @@ audiounit_stream_get_latency(cubeb_stream * stm, uint32_t * latency)
   //TODO
   return CUBEB_ERROR_NOT_SUPPORTED;
 #else
-  *latency = stm->total_output_latency_frames + cubeb_resampler_latency(stm->resampler.get());
+  uint32_t resample_latency = static_cast<uint32_t>(ceil(cubeb_resampler_latency(stm->resampler.get()) * stm->resample_ratio));
+
+  *latency = stm->total_output_latency_frames + resample_latency;
   return CUBEB_OK;
 #endif
 }
--- a/src/cubeb_audiounit.cpp
+++ b/src/cubeb_audiounit.cpp
@@ -2969,10 +2969,12 @@ static int
 audiounit_stream_get_position(cubeb_stream * stm, uint64_t * position)
 {
   assert(stm);
-  if (stm->current_latency_frames > stm->frames_played) {
+  uint32_t latency_frames = stm->total_output_latency_frames + cubeb_resampler_latency(stm->resampler.get());
+
+  if (latency_frames > stm->frames_played) {
     *position = 0;
   } else {
-    *position = stm->frames_played - stm->current_latency_frames;
+    *position = stm->frames_played - latency_frames;
   }
   return CUBEB_OK;
 }
@@ -2984,7 +2986,7 @@ audiounit_stream_get_latency(cubeb_stream * stm, uint32_t * latency)
   //TODO
   return CUBEB_ERROR_NOT_SUPPORTED;
 #else
-  *latency = stm->total_output_latency_frames;
+  *latency = stm->total_output_latency_frames + cubeb_resampler_latency(stm->resampler.get());
   return CUBEB_OK;
 #endif
 }
--- a/src/cubeb_jack.cpp
+++ b/src/cubeb_jack.cpp
@@ -658,9 +658,9 @@ cbjack_get_max_channel_count(cubeb * /*ctx*/, uint32_t * max_channels)
 }
 
 static int
-cbjack_get_latency(cubeb_stream * stm, unsigned int * latency_ms)
+cbjack_get_latency(cubeb_stream * stm, unsigned int * latency)
 {
-  *latency_ms = stm->context->jack_latency;
+  *latency = stm->context->jack_latency + ceil(cubeb_resampler_latency(stm->resampler) * stm->ratio);
   return CUBEB_OK;
 }
 
@@ -955,7 +955,7 @@ cbjack_stream_stop(cubeb_stream * stream)
 static int
 cbjack_stream_get_position(cubeb_stream * stream, uint64_t * position)
 {
-  *position = stream->position;
+  *position = stream->position - ceil(cubeb_resampler_latency(stream->resampler) * stream->ratio);
   return CUBEB_OK;
 }
 
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -2596,12 +2596,15 @@ int wasapi_stream_get_position(cubeb_stream * stm, uint64_t * position)
     return CUBEB_ERROR;
   }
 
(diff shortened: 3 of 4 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -2399,28 +2399,30 @@ int wasapi_stream_get_position(cubeb_stream * stm, uint64_t * position)
   return CUBEB_OK;
 }
 
-int wasapi_stream_get_latency(cubeb_stream * stm, uint32_t * latency)
+int wasapi_stream_get_latency(cubeb_stream *stm, uint32_t *latency)
```
kinetiknz: Please retain the original formatting for all of this code.

src/cubeb_wasapi.cpp
```diff
@@ -2399,28 +2399,30 @@ int wasapi_stream_get_position(cubeb_stream * stm, uint64_t * position)
...
+  /* Verify that AudioClient has been initialized. */
+  if (!stm->output_client)
+  {
     return CUBEB_ERROR;
   }
 
-  REFERENCE_TIME latency_hns;
-  HRESULT hr = stm->output_client->GetStreamLatency(&latency_hns);
-  if (FAILED(hr)) {
+  if (!has_output(stm))
```
kinetiknz: Why is this check needed in addition to stm->output_client?

kinetiknz (commented):
It looks like the OpenSL backend needs updating too, can you check please?

The Rust version of the CoreAudio backend will need to be adjusted too: https://github.com/ChunMinChang/cubeb-coreaudio-rs, cc @ChunMinChang

kinetiknz:
We probably also want to adjust cubeb_stream_get_position to account for resampler latency.  Otherwise, cubeb_stream_get_latency will report a different latency value to the value we're using to adjust the position reported via cubeb_stream_get_position.

mdsitton:
@kinetiknz One thing for the opensl backend. There is no get_latency implementation at all currently. That's why i didn't do any changes there.

mdsitton:
Alright, so other than the changes for OpenSL, and any rust backends that should be everything.

padenot:
> I can add it, we have the right number.

#535

mdsitton:
Alright, so other than the changes for OpenSL, and any rust backends that should be everything.

achronop:
Thank you for doing that. Have you tested it? Let me know if I can help with that? Also, did you check the numbers returned by `cubeb_resampler_latency` method? are they sensible? Just curiosity since this is the first time that this method will be used.

achronop:
Also, the title of the patch is not very descriptive and indicates a windows 10 fix which is not the case.

src/cubeb_audiounit.cpp
```diff
@@ -2972,7 +2972,7 @@ audiounit_stream_get_position(cubeb_stream * stm, uint64_t * position)
   if (stm->current_latency_frames > stm->frames_played) {
     *position = 0;
   } else {
-    *position = stm->frames_played - stm->current_latency_frames;
+    *position = stm->frames_played - (stm->current_latency_frames + cubeb_resampler_latency(stm->resampler.get()));
```
kinetiknz: It's possible `frames_played` is less than the latency subtracted here, so this could wrap around.  We'll need to clamp to 0 here (and in the other backends the same way).

src/cubeb_wasapi.cpp
```diff
@@ -2403,24 +2403,21 @@ int wasapi_stream_get_latency(cubeb_stream * stm, uint32_t * latency)
...
   }
 
   auto_lock lock(stm->stream_reset_lock);
 
-  /* The GetStreamLatency method only works if the
-     AudioClient has been initialized. */
-  if (!stm->output_client) {
+  /* Verify that AudioClient has been initialized. */
+  if (!stm->output_client)
+  {
```
kinetiknz: Move the { up to the previous line here and for the has_output check above.

src/cubeb_wasapi.cpp
```diff
@@ -2403,24 +2403,21 @@ int wasapi_stream_get_latency(cubeb_stream * stm, uint32_t * latency)
...
   }
 
-  REFERENCE_TIME latency_hns;
-  HRESULT hr = stm->output_client->GetStreamLatency(&latency_hns);
-  if (FAILED(hr)) {
-    return CUBEB_ERROR;
-  }
-  *latency = hns_to_frames(stm, latency_hns);
+  double delay = current_stream_delay(stm);
+  *latency = std::ceil(delay * stm->output_mix_params.rate) + cubeb_resampler_latency(stm->resampler.get());
```
kinetiknz: Maybe we should just include the resampler delay inside the current_stream_delay implementation?

mdsitton:
I'll work on these changes when i get time. Which will probably be in a few weeks.

mdsitton:
Just an update, haven't forgotten but I've started a new job and have been in the process of moving things are starting to settle down so hopefully soon 🤞

ChunMinChang:
@mdsitton, any update?

mdsitton:
I'll try to get back on this soon, maybe this weekend.

mdsitton:
Looks like the bug with wasapi i was also fixing in this pr has already been fixed on master so i've pulled that one out of this.

I've updated everything else mostly.


One thing i need to do still is convert the audiounit code to convert the re-sampler frame latency to the stream sample freqency instead of the hw output frequency. I think there may be some issues here on the audiounit code though i need to dig through the code some more to verify.

mdsitton:
Tomorrow i'll pull out my macbook and work on the audiounit backend a bit closer. I haven't really tested the backends too much yet. Mostly just updating my previous code and reworking some parts.

mdsitton:
@kinetiknz This should be ready to review again whenever you get the chance.

#559 - AAudio PR will need to be worked on for this whenever this is merged as well as the rust version of audiounit.

mdsitton:
@kinetiknz any update on when this will be reviewed?

mdsitton:
@padenot Perhaps you could take a look at this then? been waiting on someone to review for 3 months now! I'd certainly like this to get merged if possible. Or to get further feedback on the change.

Thanks!

src/cubeb_jack.cpp
```diff
@@ -955,7 +955,7 @@ cbjack_stream_stop(cubeb_stream * stream)
 static int
 cbjack_stream_get_position(cubeb_stream * stream, uint64_t * position)
 {
-  *position = stream->position;
+  *position = stream->position - ceil(cubeb_resampler_latency(stream->resampler) * stream->ratio);
```
padenot, line 958: This will underflow on stream start, you want to branch here and return 0 if this would be negative: if the stream has just started, it's still filling the buffers in the underlying layer, and nothing is being output. Because `stream_get_position` is the time of what is being heard, then 0 is the correct answer in this situation.

src/cubeb_wasapi.cpp
```diff
@@ -2596,12 +2596,15 @@ int wasapi_stream_get_position(cubeb_stream * stm, uint64_t * position)
     return CUBEB_ERROR;
   }
 
+  double resample_ratio = stream_to_mix_samplerate_ratio(stm->output_stream_params, stm->output_mix_params);
+
   /* Calculate how far behind the current stream head the playback cursor is. */
-  uint64_t stream_delay = static_cast<uint64_t>(current_stream_delay(stm) * stm->output_stream_params.rate);
+  uint64_t resampler_delay = ceil(cubeb_resampler_latency(stm->resampler.get()) * resample_ratio);
```
padenot, line 2602: This is tricky. `cubeb_resampler_latency` returns the output latency, i.e., the number of frames of latency, expressed in output rate, so we don't need this, because `stream_get_position` returns its value in frames at the output (= stream) rate.

src/cubeb_wasapi.cpp
```diff
@@ -2637,13 +2640,18 @@ int wasapi_stream_get_latency(cubeb_stream * stm, uint32_t * latency)
   if (FAILED(hr)) {
     return CUBEB_ERROR;
   }
+
+
+  double resample_ratio = stream_to_mix_samplerate_ratio(stm->output_stream_params, stm->output_mix_params);
+  uint32_t resample_latency = ceil(cubeb_resampler_latency(stm->resampler.get()) * resample_ratio);
```
padenot, line 2646: Same logic applies here.

src/cubeb_wasapi.cpp
```diff
@@ -2664,7 +2672,11 @@ int wasapi_stream_get_input_latency(cubeb_stream * stm, uint32_t * latency)
     return CUBEB_ERROR;
   }
 
-  *latency = hns_to_frames(stm, stm->input_latency_hns);
+
+  double resample_ratio = stream_to_mix_samplerate_ratio(stm->output_stream_params, stm->output_mix_params);
```
padenot, line 2676: This is input latency, and the input mix rate is frequently different from the output rate (for example because the microphone and the output device are two separate devices, that have been configured with different rates: this is sub-optimal, but happens).

src/cubeb_wasapi.cpp
```diff
@@ -2664,7 +2672,11 @@ int wasapi_stream_get_input_latency(cubeb_stream * stm, uint32_t * latency)
...
+  uint32_t resample_latency = ceil(cubeb_resampler_latency(stm->resampler.get()) * resample_ratio);
```
padenot, line 2677: It appears that you can't write this function with the current API: `cubeb_resampler` is a bidirectional resampler, that can simultaneously resample an input stream at rate A (microphone) and an output stream at rate B (speakers), both to a rate C (that can be equal to either A or B). There is nothing in the API that surfaces the input latency, I'll add it.

In general there is some work needed on the resampler, that became very clear when working on https://github.com/kinetiknz/cubeb/pull/598

src/cubeb_audiounit.cpp
```diff
@@ -2984,7 +2990,9 @@ audiounit_stream_get_latency(cubeb_stream * stm, uint32_t * latency)
   //TODO
   return CUBEB_ERROR_NOT_SUPPORTED;
 #else
-  *latency = stm->total_output_latency_frames;
+  uint32_t resample_latency = static_cast<uint32_t>(ceil(cubeb_resampler_latency(stm->resampler.get()) * stm->resample_ratio));
```
padenot, line 2993: This is already in logical stream rate, no need to adjust.

padenot (changes requested):

PR #535, opened 2019-08-29
Implement `stream_get_latency` for the opensl backend.

padenot:


src/cubeb_opensl.c
```diff
@@ -1625,6 +1625,20 @@ opensl_stream_get_position(cubeb_stream * stm, uint64_t * position)
...
 
+static int
+opensl_stream_get_latency(cubeb_stream * stm, uint32_t * latency)
+{
+  assert(stm && latency);
+
+  uint32_t stream_latency_frames =
+    (stm->user_output_rate * stm->output_latency_ms) / 1000;
+
+  if (stm->resampler) {
```
achronop: The resampler is always there, look at [1], better assert if you want to verify.

[1] https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_opensl.c#L1379

src/cubeb_opensl.c
```diff
@@ -1625,6 +1625,20 @@ opensl_stream_get_position(cubeb_stream * stm, uint64_t * position)
   return CUBEB_OK;
 }
 
+static int
+opensl_stream_get_latency(cubeb_stream * stm, uint32_t * latency)
+{
+  assert(stm && latency);
```
achronop: I prefer to break that kind of asserts into 2 separate individual asserts since it is easier to find out which one has been hit. Not a strong comment.

src/cubeb_opensl.c
```diff
@@ -1625,6 +1625,20 @@ opensl_stream_get_position(cubeb_stream * stm, uint64_t * position)
...
+
+  uint32_t stream_latency_frames =
+    (stm->user_output_rate * stm->output_latency_ms) / 1000;
```
achronop: It does not change anything but since the `user_output_rate` is samples per seconds and the `output_latency_ms` is in ms I would prefer different parenthesis:
`stm->user_output_rate * (stm->output_latency_ms / 1000)`

PR #539, opened 2019-09-05
Add lazy bindings for ALSA and SNDIO

jbeich:
Similar to PulseAudio and Jack:
- ALSA can also be used on DragonFly, FreeBSD and NetBSD e.g., [Firefox#1021761](https://bugzilla.mozilla.org/show_bug.cgi?id=1021761)
- sndio can also be used on Linux and FreeBSD e.g., [Firefox#1351378](https://bugzilla.mozilla.org/show_bug.cgi?id=1351378), https://github.com/citra-emu/citra/issues/3862

CC @landryb, @ratchov

```diff
diff --git a/media/libcubeb/gtest/moz.build b/media/libcubeb/gtest/moz.build
index 1ff0c02545251..ca364d1ef7845 100644
--- a/media/libcubeb/gtest/moz.build
+++ b/media/libcubeb/gtest/moz.build
@@ -62,20 +62,13 @@ if CONFIG['MOZ_WIDGET_TOOLKIT'] == 'cocoa':
         '-framework AudioUnit',
         '-framework CoreAudio',
     ]
 elif CONFIG['MOZ_WIDGET_TOOLKIT'] == 'uikit':
     OS_LIBS += [
       '-framework CoreFoundation',
       '-framework AudioToolbox',
     ]
-elif CONFIG['OS_TARGET'] == 'OpenBSD':
-    OS_LIBS += [
-        'sndio',
-    ]
-else:
-    OS_LIBS += CONFIG['MOZ_ALSA_LIBS']
-    OS_LIBS += CONFIG['MOZ_PULSEAUDIO_LIBS']
 
 if CONFIG['CC_TYPE'] in ('clang', 'gcc'):
     CXXFLAGS += ['-Wno-error=shadow']
 
 FINAL_LIBRARY = 'xul-gtest'
diff --git a/toolkit/library/moz.build b/toolkit/library/moz.build
index 8ce595ff96bff..c318e96a1fdfc 100644
--- a/toolkit/library/moz.build
+++ b/toolkit/library/moz.build
@@ -226,32 +226,24 @@ if CONFIG['MOZ_SYSTEM_LIBEVENT']:
     OS_LIBS += CONFIG['MOZ_LIBEVENT_LIBS']
 
 if CONFIG['MOZ_SYSTEM_LIBVPX']:
     OS_LIBS += CONFIG['MOZ_LIBVPX_LIBS']
 
 if not CONFIG['MOZ_TREE_PIXMAN']:
[...]

PR #541, opened 2019-09-13
Don't depend on development package for ALSA and SNDIO

jbeich:
Regressed by #539

- On Debian `.so` symlink is part of separate [`libasound2-dev`](https://packages.debian.org/sid/amd64/libasound2-dev/filelist) and [`libsndio-dev`](https://packages.debian.org/sid/amd64/libsndio-dev/filelist) packages
- On OpenBSD `.so.X` symlink is not provided because OS updates often change ABI of all libraries
- On rolling distributions runtime and development bits are not separate


kinetiknz:
LGTM, thanks!

Issue #534, opened 2019-08-24
RFC: WASAPI: handle COM initialization internally again, in a thread this time


ligfx:
I'd like to suggest a new approach to COM within cubeb, where cubeb would again initialize and handle COM transparently for the user. Currently, the requirement to initialize and understand COM  falls on the user, which can be a complex and error-prone task.

For reference, cubeb used to initialize COM on behalf of the user, but that was changed in #424 to fix issues documented in #416. Cubeb was initializing COM on the stack for each API call, which could lead to crashes when objects outlasted their COM context.

The trick here is that COM objects are tied to the context on the thread they're created in— so, I'm proposing that cubeb create and manage a thread solely for creating COM objects. If the thread is a child of the main cubeb context, all COM objects will last for the lifetime of cubeb. Here's an example:

```c++
class COMApartment {
 public:
  COMApartment() {
    this->new_message_event = CreateEvent(NULL, 0, 0, NULL);
    if (!this->new_message_event) {
      LOG("Can't create new message event, error: %lx", GetLastError());
      XASSERT(false);
    }

    this->response_ready_event = CreateEvent(NULL, 0, 0, NULL);
    if (!this->response_ready_event) {
      LOG("Can't create response ready event, error: %lx", GetLastError());
      XASSERT(false);
    }

    this->thread = std::thread{[this] {
      struct auto_com {
        auto_com() {
          XASSERT(SUCCEEDED(CoInitializeEx(nullptr, COINIT_MULTITHREADED)));
        }
        ~auto_com() { CoUninitialize(); }
      } com;

      while (true) {
        WaitForSingleObject(this->new_message_event, INFINITE);
        if (this->request_message_type == MessageType::MESSAGE_QUIT) {
          return;
        }
        XASSERT(this->request_message_type ==
                MessageType::MESSAGE_CREATE_INSTANCE);
        this->result_hr =
            CoCreateInstance(this->request_rclsid, NULL, CLSCTX_INPROC_SERVER,
                             this->request_riid, this->request_ppv);
        if (!SetEvent(this->response_ready_event)) {
          LOG("SetEvent failed: %d", GetLastError());
          XASSERT(false);
        }
      }
    }};
  }
  ~COMApartment() {
    {
      std::lock_guard<owned_critical_section> write_lock(this->write_lock);
      this->request_message_type = MessageType::MESSAGE_QUIT;
      if (!SetEvent(this->new_message_event)) {
[...]

PR #542, opened 2019-09-14
[PoC] wasapi: make render thread exist for the lifetime of the stream object

ligfx:
Took a crack at this to see what it might look like. See #534 for why I think this could be a good idea.

- Converts cubeb_stream into a reference-counted object, and gets rid of
  emergency_bailout logic
- When reconfiguring, only restart a stream if it was already started
- Removes timeout inside the render thread, as we can't guarantee events
  while a stream is stopped


ligfx:
@kinetiknz 

After `stream_destroy`, the only way I see that a callback would be executed is if WaitForMultipleObjects stalls, and then returns a refill/input_available event even though a shutdown event is also set. Do we know if WaitForMultipleObjects ever stalls, or if the stalls are just inside WASAPI?

After `stream_stop` is complicated. Even after calling IAudioClient::Stop, the events can still be triggered. I'm seeing usually one more event to fill up the buffer (which executes a callback), and then events with 0 frames left to fill (which doesn't execute a callback, but might unnecessarily add CPU load/thread context switches). How strict is the requirement not to execute callbacks after `stream_stop`? Thinking through this, I think the only way to accomplish this in every case would be to move the stream start/stop logic inside the render loop thread. Not that complex to code but increases the probability of encountering a stalled thread.

kinetiknz:
> After `stream_destroy`, the only way I see that a callback would be executed is if WaitForMultipleObjects stalls, and then returns a refill/input_available event even though a shutdown event is also set. Do we know if WaitForMultipleObjects ever stalls, or if the stalls are just inside WASAPI?

Just WASAPI, so this case is fine and I'm just wrong about it.  stream_destroy is the important case, since we expect the caller to free resources after the stream is destroyed.

> After stream_stop is complicated. Even after calling IAudioClient::Stop, the events can still be triggered. I'm seeing usually one more event to fill up the buffer (which executes a callback), and then events with 0 frames left to fill (which doesn't execute a callback, but might unnecessarily add CPU load/thread context switches). How strict is the requirement not to execute callbacks after stream_stop? Thinking through this, I think the only way to accomplish this in every case would be to move the stream start/stop logic inside the render loop thread. Not that complex to code but increases the probability of encountering a stalled thread.

I think this is fine for now, since it's no different from the behaviour before your changes.  Eventually we'll want to tighten this up for all backends, I think.

ligfx:
@kinetiknz Apologies for the delay on this. Rebased to current master (taking extra care around the changes to the thread/bailout logic), and based on the comments above this is ready for review as-is.

ligfx:
@kinetiknz Are you still interested in this? Happy to rebase and fix conflicts if so.

danilaml:
@kinetiknz any reason this wasn't merged?

padenot:
Sorry about that, I think we forgot. I've rebased your initial patch, because the underlying code has changed a lot.

https://github.com/mozilla/cubeb/tree/wasapi_render_loop_lifecycle_rebase is my rebase. I'll change this branch so that it's cleaner, there are various things in a single commit at the moment and CI changes unrelated to your fix.

We now have CI for Windows (as of last week): https://github.com/mozilla/cubeb/actions/runs/4862284679/jobs/8668439986, and I'm trying the patch on Firefox's CI, running most of our media tests, we'll see.

danilaml:
@padenot I think you've made the typo. The original author of this code is @ligfx , not me. I've just discovered it when root causing the bug I was experiencing.

danilaml:
@padenot for the record, I've been running with the rebased branch for a few days so far and haven't encountered any issues (and it did indeed fix the problem I was having). Granted, that's even less coverage than mozilla CI provides, but at least I can confirm that this solves something ;P

padenot:
> Granted, that's even less coverage than mozilla CI provides

Yes, but the workloads are less synthetic, so it's useful to know, thanks for giving it a try! I'm going to re-read the patch myself and probably merge it.

padenot:
@kinetiknz I'm merging this, it's green all around on treeherder and here. I've added a couple commits to be able to have `clang-format` that has the same results locally and on the CI but it's the same as the original patch (rebased) otherwise.

PR #547, opened 2019-09-30
pulse: when default device changes update the info

achronop:
The problem was that after a default device change the cubeb_get_max_channel_count() returned the wrong number of channels; it returned the device info (including channel count) of the initial default device. In order to update the default device info dynamically, a subscription to the  event is created during context initialization and the device info is updated accordingly.

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -615,6 +615,9 @@ pulse_context_init(cubeb * ctx)
   return 0;
 }
 
+static int pulse_subscribe_notifications(cubeb * context,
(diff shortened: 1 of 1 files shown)


kinetiknz:
I assume we'll need to make the same fix to cubeb-pulse-rs too.

kinetiknz:
Sorry, I misunderstood the patch.  You've already handled the situation I'm concerned about by including PA_SUBSCRIPTION_MASK_SERVER when registering via pulse_register_device_collection_changed and never unregistering the callback.

achronop:
Thank you, I am working on the rust side.

PR #557, opened 2019-12-03
Add support for float playback on android (OpenSLES)

nyorain:
Float output is only available (and the code only be used) if the targeted android version at compile time as well as the device version at runtime is high enough (>= lollipop).
This PR does not add support for float capturing since i don't really care about that. Tested it with a >= lollipop device. 
BTW, would there be any interest in an AAudio backend for cubeb? Some of the OpenSLES stuff seems kinda messy to me and AAudio might improve performance.

### DIFF
--- a/src/cubeb_opensl.c
+++ b/src/cubeb_opensl.c
@@ -1080,25 +1080,23 @@ opensl_configure_playback(cubeb_stream * stm, cubeb_stream_params * params) {
   SLuint32* format_sample_rate = NULL;
 
 #if defined(__ANDROID__) && (__ANDROID_API__ >= ANDROID_VERSION_LOLLIPOP)
-  SLAndroidDataFormat_PCM_EX aformat;
+  SLAndroidDataFormat_PCM_EX pcm_ext_format;
   if (get_android_version() >= ANDROID_VERSION_LOLLIPOP) {
-    if (opensl_set_format_ext(&aformat, params) != CUBEB_OK) {
+    if (opensl_set_format_ext(&pcm_ext_format, params) != CUBEB_OK) {
       return CUBEB_ERROR_INVALID_FORMAT;
     }
-
-    format = &aformat;
-    format_sample_rate = &aformat.sampleRate;
+    format = &pcm_ext_format;
+    format_sample_rate = &pcm_ext_format.sampleRate;
   }
 #endif
 
-  SLDataFormat_PCM bformat;
+  SLDataFormat_PCM pcm_format;
   if(!format) {
-    if(opensl_set_format(&bformat, params) != CUBEB_OK) {
+    if(opensl_set_format(&pcm_format, params) != CUBEB_OK) {
       return CUBEB_ERROR_INVALID_FORMAT;
     }
-
-    format = &bformat;
-    format_sample_rate = &bformat.samplesPerSec;
+    format = &pcm_format;
+    format_sample_rate = &pcm_format.samplesPerSec;
   }
 
   SLDataLocator_BufferQueue loc_bufq;
--- a/src/cubeb_opensl.c
+++ b/src/cubeb_opensl.c
@@ -794,6 +794,53 @@ opensl_destroy(cubeb * ctx)
 
 static void opensl_stream_destroy(cubeb_stream * stm);
 
+#if defined(__ANDROID__) && (__ANDROID_API__ >= ANDROID_VERSION_LOLLIPOP)
+static int
+opensl_set_format_android(SLAndroidDataFormat_PCM_EX * format, cubeb_stream_params * params)
+{
+  assert(format);
+  assert(params);
+
+  format->formatType = SL_ANDROID_DATAFORMAT_PCM_EX;
+  format->numChannels = params->channels;
+  // samplesPerSec is in milliHertz
+  format->sampleRate = params->rate * 1000;
+  format->channelMask = params->channels == 1 ?
+                       SL_SPEAKER_FRONT_CENTER :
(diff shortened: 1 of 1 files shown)


achronop:
>  BTW, would there be any interest in an AAudio backend for cubeb? Some of the OpenSLES stuff seems kinda messy to me and AAudio might improve performance.

We are definitely interested in AAudio, it's a matter of priorities and time I guess. However, we know that we have to implement it at some point. An even better solution would be to add an [Oboe](https://github.com/google/oboe) backend that handles the routing between OpenSLES and AAudio internally.

nyorain:
I will fix up my AAudio backend and open a new PR, then. I considered Oboe but you already have a working OpenSLES backend and Oboe adds some stuff you don't need. In the end Oboes api is basically just AAudio so if you decide to abandon your OpenSLES backend later on and switch to Oboe, it should be very easy to convert the AAudio backend.
Or are there any reasons you would feel strongly against supporting a bare AAudio backend and in favor of using Oboe?

achronop:
>  Or are there any reasons you would feel strongly against supporting a bare AAudio backend and in favor of using Oboe?

Nope, a bare AAudio backend it's totally fine, thank you for the offering. As soon as it passes reviewing and testing we are more than happy to have.

src/cubeb_opensl.c
```diff
@@ -1020,13 +1067,34 @@ opensl_configure_playback(cubeb_stream * stm, cubeb_stream_params * params) {
...
   stm->lastCompensativePosition = -1;
 
-  SLDataFormat_PCM format;
-  int r = opensl_set_format(&format, params);
+  void* format = NULL;
+  SLuint32* format_sample_rate = NULL;
+  int r;
+#if defined(__ANDROID__) && (__ANDROID_API__ >= ANDROID_VERSION_LOLLIPOP)
+  SLAndroidDataFormat_PCM_EX aformat;
+  if(get_android_version() >= ANDROID_VERSION_LOLLIPOP) {
```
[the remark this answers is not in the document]
nyorain: I'm by no means an expert on android ndk stuff but after some research it seems to be possible. It might be a **really** bad idea to do something like this though (see https://stackoverflow.com/questions/21888052, the second answer mentions many problems). Since you might get it to work nonetheless, I'd rather leave this check here.

src/cubeb_opensl.c
```diff
@@ -1020,13 +1067,34 @@ opensl_configure_playback(cubeb_stream * stm, cubeb_stream_params * params) {
...
+    format = &aformat;
+    format_sample_rate = &aformat.sampleRate;
+    r = opensl_set_format_android(&aformat, params);
```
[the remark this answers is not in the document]
nyorain: I agree that the code can be somewhat confusing to understand, I'll move it around.

src/cubeb_opensl.c
```diff
@@ -1078,25 +1078,27 @@ opensl_configure_playback(cubeb_stream * stm, cubeb_stream_params * params) {
...
+  if (get_android_version() >= ANDROID_VERSION_LOLLIPOP) {
+    if (opensl_set_format_ext(&aformat, params) != CUBEB_OK) {
+      return CUBEB_ERROR_INVALID_FORMAT;
+    }
+
     format = &aformat;
     format_sample_rate = &aformat.sampleRate;
-    r = opensl_set_format_android(&aformat, params);
   }
 #endif
```
achronop: That's much better thank you. Now is clear that we want either one or the other. It would be nice to express that explicitly with an if-else without checking the state of the `format` pointer. Something like (easier to express it in code than describing it):
```
  void* format = NULL;
  SLuint32* format_sample_rate = NULL;
  SLDataFormat_PCM bformat;
#if defined(__ANDROID__) && (__ANDROID_API__ >= ANDROID_VERSION_LOLLIPOP)
  SLAndroidDataFormat_PCM_EX aformat;
  if (get_android_version() >= ANDROID_VERSION_LOLLIPOP) {
 ... same stuff ...
  } else {
#endif
  if(opensl_set_format(&bformat, params) != CUBEB_OK) {
  } 
  assert(!format);
  format = &bformat;
  format_sample_rate = &bformat.samplesPerSec;
#if defined(__ANDROID__) && (__ANDROID_API__ >= ANDROID_VERSION_LOLLIPOP)
  }
#endif
```
It also asserts that the `format` pointer is `null` before being set, which is not necessary, feel free to omit it if you don't like.

src/cubeb_opensl.c
```diff
@@ -1078,25 +1078,27 @@ opensl_configure_playback(cubeb_stream * stm, cubeb_stream_params * params) {
 
   void* format = NULL;
   SLuint32* format_sample_rate = NULL;
-  int r;
+
 #if defined(__ANDROID__) && (__ANDROID_API__ >= ANDROID_VERSION_LOLLIPOP)
   SLAndroidDataFormat_PCM_EX aformat;
```
achronop: Can we change the `aformat` and `bformat` names to something more representative? It can be `pcm_ext_format` and `pcm_format` or something similar.

src/cubeb_opensl.c
nyorain: I will fix the other issues but I don't think I agree on this one. I initially tried to do something like this as well and I see your point but mixing up control flow with preprocessor logic makes code harder to understand IMO. Unless you require it for merging, I would prefer to keep the code as it currently stands.

achronop: That's fine, since it is correct and makes sense it's up to you.

achronop:
Try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=9aba651779348ddccb2fe9049461770b89995fe3

Issue #556, opened 2019-11-23
Android: `use of invalid jobject` in `cubeb_jni_init()`


tido64:
On recent master (8143207), I'm getting the following error on my Nokia 8 Sirocco running Android 9 with security patch level Nov 1 2019:

```
JNI DETECTED ERROR IN APPLICATION: use of invalid jobject 0x79e5271000
"Thread-2" prio=5 tid=18 Runnable
  | group="main" sCount=0 dsCount=0 flags=0 obj=0x12c40000 self=0x79f74ea000
  | sysTid=5020 nice=0 cgrp=default sched=0/0 handle=0x79e536e4f0
  | state=R schedstat=( 50087128 34066455 67 ) utm=4 stm=1 core=0 HZ=100
  | stack=0x79e5273000-0x79e5275000 stackSize=1009KB
  | held mutexes= "mutator lock"(shared held)
  native: #00 pc 00000000003c16c8  /system/lib64/libart.so (art::DumpNativeStack(std::__1::basic_ostream<char, std::__1::char_traits<char>>&, int, BacktraceMap*, char const*, art::ArtMethod*, void*, bool)+220)
  native: #01 pc 000000000048ed34  /system/lib64/libart.so (art::Thread::DumpStack(std::__1::basic_ostream<char, std::__1::char_traits<char>>&, bool, BacktraceMap*, bool) const+352)
  native: #02 pc 00000000002e4bf8  /system/lib64/libart.so (art::JavaVMExt::JniAbort(char const*, char const*)+968)
  native: #03 pc 00000000002e50f8  /system/lib64/libart.so (art::JavaVMExt::JniAbortF(char const*, char const*, ...)+180)
  native: #04 pc 0000000000494958  /system/lib64/libart.so (art::Thread::DecodeJObject(_jobject*) const+808)
  native: #05 pc 0000000000000074  <anonymous:7a8033c000> (???)
  (no managed stack frames)
```

Where `0x79e5271000` is `jstr` in the snippet below:

```c
 // cubeb-jni.cpp
 extern "C"
 cubeb_jni *
 cubeb_jni_init()
 {
   [...]
   // Find the audio manager object and make it global to call it from another method
   jclass context_class = jni_env->FindClass("android/content/Context");
   jfieldID audio_service_field = jni_env->GetStaticFieldID(context_class, "AUDIO_SERVICE", "Ljava/lang/String;");
[...]

PR #558, opened 2019-12-03
Remove android latency assertion

nyorain:
A method for querying latency isn't strictly required, streams
work without it. But when no method could be loaded, this assertion
would be triggered when called from opensl_configure_playback.

### DIFF
--- a/src/android/cubeb-output-latency.h
+++ b/src/android/cubeb-output-latency.h
@@ -35,10 +35,7 @@ cubeb_output_latency_load_method(int version)
 bool
 cubeb_output_latency_method_is_loaded(output_latency_function * ol)
 {
-  if(!ol) {
-    return false;
-  }
-
+  assert(ol);
   if (ol->version > ANDROID_JELLY_BEAN_MR1_4_2){
     return !!ol->from_jni;
   }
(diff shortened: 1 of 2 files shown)


src/android/cubeb-output-latency.h
```diff
@@ -35,7 +35,10 @@ cubeb_output_latency_load_method(int version)
 bool
 cubeb_output_latency_method_is_loaded(output_latency_function * ol)
 {
-  assert(ol && (ol->from_jni || ol->from_lib));
+  if(!ol) {
+    return false;
+  }
+
```
tido64: I think asserting on `ol` still makes sense. We can guarantee that regardless of whether any methods were loaded. In my testing, this works:
```
assert(ol != NULL);
```
cc @achronop as we were just discussing this in #556.

nyorain: Hm yeah, you're right. But then the warning about cubeb_stream_get_position not being available in opensl_init is misplaced and will never be triggered. We'll either have to move that as well or make cubeb_output_latency_load_method return NULL on failure.

achronop: Thank you both for your input, keep doing that.

We can keep the assert as @kido64 suggested and then we can use the `cubeb_output_latency_method_is_loaded` in the warning check in `cubeb_init`, mentioned by @nyorain . Does it sound resonable?

nyorain: Yes that makes sense

achronop:
Try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=152f76b46e42b34aedd978b14c703657ce21b881

achronop:
Try is green thanks

PR #559, opened 2019-12-06
Add AAudio backend

nyorain:
This implements a first AAudio backend with support for input, output and duplex streams. It uses callbacks for best latency and is completely lock-free in the audio thread (as the aaudio docs recommend). I could only test it on one device. Some additional functions could be implemented via guesswork, jni or other APIs (see the SLES backend), I didn't do that for now and would prefer to keep this PR down to a first basic implementation. There are a couple of NOTEs in the code in places that could need improvements or at stuff I wasn't sure about. I will continue to fix out minor details but am already thankful for input, especially regarding the general backend design.
The cubeb docs don't state this so i assumed it's ok if the state callback is called from a different
thread than the audio callback and that they don't have to be mutually exclusive, i.e. can be called at the same time. Is that correct?
We furthermore currently send a callback with STATE_STOPPED some time after STATE_DRAINED since that simplifies the backend state logic and isn't strictly wrong. Does that break expected behavior? That could be changed without much work.
The backend is currently not used over SLES (i.e. never unless explicitly requested) but i guess this should be changed when it's well tested as it can be expected to provide better latency.

### DIFF
--- a/src/cubeb.c
+++ b/src/cubeb.c
@@ -63,6 +63,9 @@ int opensl_init(cubeb ** context, char const * context_name);
 #if defined(USE_OSS)
 int oss_init(cubeb ** context, char const * context_name);
 #endif
+#if defined(USE_AAUDIO)
+int aaudio_init(cubeb ** context, char const * context_name);
+#endif
 #if defined(USE_AUDIOTRACK)
 int audiotrack_init(cubeb ** context, char const * context_name);
 #endif
@@ -173,6 +176,10 @@ cubeb_init(cubeb ** context, char const * context_name, char const * backend_nam
 #if defined(USE_OSS)
       init_oneshot = oss_init;
 #endif
+    } else if (!strcmp(backend_name, "aaudio")) {
+#if defined(USE_AAUDIO)
+      init_oneshot = aaudio_init;
+#endif
     } else if (!strcmp(backend_name, "audiotrack")) {
 #if defined(USE_AUDIOTRACK)
       init_oneshot = audiotrack_init;
@@ -227,6 +234,11 @@ cubeb_init(cubeb ** context, char const * context_name, char const * backend_nam
 #endif
 #if defined(USE_OPENSL)
     opensl_init,
+#endif
+    // TODO: should probably be preferred over OpenSLES when available.
+    // Initialization will fail on old android devices.
+#if defined(USE_AAUDIO)
+    aaudio_init,
 #endif
 #if defined(USE_AUDIOTRACK)
     audiotrack_init,
--- a/src/cubeb_aaudio.c
+++ b/src/cubeb_aaudio.c
@@ -115,6 +115,10 @@ struct cubeb_stream {
   pthread_mutex_t mutex;
 
   unsigned in_frame_size;
+
+  cubeb_sample_format out_format;
+  _Atomic float volume;
+  unsigned out_channels;
   unsigned out_frame_size;
 };
 
@@ -397,6 +401,30 @@ aaudio_init(cubeb ** context, char const * context_name) {
   return CUBEB_OK;
 }
 
+static void
+apply_volume(cubeb_stream * stm, void * audio_data, uint32_t num_frames) {
+  // optimization: we don't have to change anything in this case
+  float volume = atomic_load(&stm->volume);
+  if(volume == 1.f) {
+    return;
+  }
+
+  switch(stm->out_format) {
+    case CUBEB_SAMPLE_S16NE:
+      for(uint32_t i = 0u; i < num_frames * stm->out_channels; ++i) {
+        ((int16_t*)audio_data)[i] *= volume;
+      }
+      break;
+    case CUBEB_SAMPLE_FLOAT32NE:
+      for(uint32_t i = 0u; i < num_frames * stm->out_channels; ++i) {
+        ((float*)audio_data)[i] *= volume;
+      }
+      break;
+    default:
+      assert(false && "Unreachable: invalid stream out_format");
+  }
+}
+
 // Returning AAUDIO_CALLBACK_RESULT_STOP seems to put the stream in
 // an invalid state. Seems like an AAudio bug/bad documentation.
 // We therefore only return it on error.
@@ -408,6 +436,7 @@ aaudio_duplex_data_cb(AAudioStream * astream, void * user_data,
   cubeb_stream * stm = (cubeb_stream*) user_data;
   assert(stm->ostream == astream);
   assert(stm->istream);
+  assert(num_frames >= 0);
 
   enum stream_state state = atomic_load(&stm->state);
   int istate = WRAP(AAudioStream_getState)(stm->istream);
@@ -458,12 +487,12 @@ aaudio_duplex_data_cb(AAudioStream * astream, void * user_data,
     return AAUDIO_CALLBACK_RESULT_STOP;
   } else if (done_frames < num_frames) {
     atomic_store(&stm->state, STREAM_STATE_DRAINING);
-    // RESULT_STOP would make more sense here but leads to errors.
-    // Restarting the stream would not be possible
-    return AAUDIO_CALLBACK_RESULT_CONTINUE;
+
+    char* begin = ((char*)audio_data) + done_frames * stm->out_frame_size;
+    memset(begin, 0x0, (num_frames - done_frames) * stm->out_frame_size);
   }
 
-  assert(in_num_frames == num_frames);
+  apply_volume(stm, audio_data, done_frames);
   return AAUDIO_CALLBACK_RESULT_CONTINUE;
 }
 
@@ -474,6 +503,7 @@ aaudio_output_data_cb(AAudioStream * astream, void * user_data,
   cubeb_stream * stm = (cubeb_stream*) user_data;
   assert(stm->ostream == astream);
   assert(!stm->istream);
+  assert(num_frames >= 0);
 
   enum stream_state state = atomic_load(&stm->state);
   int ostate = WRAP(AAudioStream_getState)(stm->ostream);
@@ -499,9 +529,12 @@ aaudio_output_data_cb(AAudioStream * astream, void * user_data,
     return AAUDIO_CALLBACK_RESULT_STOP;
   } else if (done_frames < num_frames) {
     atomic_store(&stm->state, STREAM_STATE_DRAINING);
-    return AAUDIO_CALLBACK_RESULT_CONTINUE;
+
+    char* begin = ((char*)audio_data) + done_frames * stm->out_frame_size;
+    memset(begin, 0x0, (num_frames - done_frames) * stm->out_frame_size);
   }
 
+  apply_volume(stm, audio_data, done_frames);
   return AAUDIO_CALLBACK_RESULT_CONTINUE;
 }
 
@@ -512,6 +545,7 @@ aaudio_input_data_cb(AAudioStream * astream, void * user_data,
   cubeb_stream * stm = (cubeb_stream*) user_data;
   assert(stm->istream == astream);
   assert(!stm->ostream);
+  assert(num_frames >= 0);
 
   enum stream_state state = atomic_load(&stm->state);
   int istate = WRAP(AAudioStream_getState)(stm->istream);
@@ -540,7 +574,6 @@ aaudio_input_data_cb(AAudioStream * astream, void * user_data,
     // stop it from the state thread. That is signaled via the
     // DRAINING state.
     atomic_store(&stm->state, STREAM_STATE_DRAINING);
-    return AAUDIO_CALLBACK_RESULT_CONTINUE;
   }
 
   return AAUDIO_CALLBACK_RESULT_CONTINUE;
@@ -629,6 +662,7 @@ aaudio_stream_destroy(cubeb_stream * stm)
     }
 
     WRAP(AAudioStream_close)(stm->ostream);
+    stm->ostream = NULL;
   }
 
   if (stm->istream) {
@@ -642,6 +676,7 @@ aaudio_stream_destroy(cubeb_stream * stm)
     }
 
     WRAP(AAudioStream_close)(stm->istream);
+    stm->istream = NULL;
   }
 
   if (stm->resampler) {
@@ -769,7 +804,11 @@ aaudio_stream_init(cubeb * ctx,
     target_sample_rate = output_stream_params->rate;
     out_params = *output_stream_params;
     out_params.rate = rate;
+
+    stm->out_channels = output_stream_params->channels;
+    stm->out_format = output_stream_params->format;
     stm->out_frame_size = frame_size;
+    atomic_store(&stm->volume, 1.f);
   }
 
   // input
@@ -1113,6 +1152,14 @@ aaudio_stream_get_position(cubeb_stream * stm, uint64_t * position)
   return CUBEB_OK;
 }
 
+static int
+aaudio_stream_set_volume(cubeb_stream * stm, float volume)
+{
+  assert(stm && stm->ostream);
+  atomic_store(&stm->volume, volume);
+  return CUBEB_OK;
+}
+
 static const struct cubeb_ops aaudio_ops = {
   .init = aaudio_init,
   .get_backend_id = aaudio_get_backend_id,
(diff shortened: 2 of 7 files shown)


achronop:
Thank you for that. It needs more iterations since it's big and I need to catch up with the AAudio specifics. For now, I am answering some of your questions

> The cubeb docs don't state this so i assumed it's ok if the state callback is called from a different
> thread than the audio callback and that they don't have to be mutually exclusive, i.e. can be called at the same time. Is that correct?

That's correct. However, we need to guarantee that the `data_callback` is not triggered after a drain or stop.  The audio thread might still be cycling but the `stm->data_callback ` *must* not be called again. That means, a client must not see another callback after a drained data_callback has returned, or must not see another callback after the `cubeb_stream_stop()` has returned. Regarding the state callback the closer to the state change event it fires the better. Just so you know on other platforms we fire the drain callback from the audio thread, not the best practice though.

> We furthermore currently send a callback with STATE_STOPPED some time after STATE_DRAINED since that simplifies the backend state logic and isn't strictly wrong. Does that break expected behavior? That could be changed without much work.

Yes. It's not a big problem but other backends do not do that so it's better to follow the same behavior here and not fire the stop event on drain. A cubeb client must call explicitly `cubeb_stream_stop()` in order to receive a stop state callback even if the stream is stopped internally due to the drain.

nyorain:
> It needs more iterations since it's big and I need to catch up with the AAudio specifics

There's no hurry. I'm still fixing up some details as well.

> However, we need to guarantee that the `data_callback` is not triggered after a drain or stop. The audio thread might still be cycling but the `stm->data_callback ` _must_ not be called again. That means, a client must not see another callback after a drained data_callback has returned, or must not see another callback after the `cubeb_stream_stop()` has returned. 

Makes sense, this is already guaranteed in the backend.

> It's not a big problem but other backends do not do that so it's better to follow the same behavior here and not fire the stop event on drain.

Okay, I changed that.

achronop:
Regarding the supported methods I would say that we need to support at least the methods that the OpenSL ES backend supports (check [here](https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_opensl.c#L1739-L1759)). From a quick look, we are missing only two.
```
  .stream_get_latency
  .stream_set_volume
```
Do you think we can add support to those easily? 

I am also adding more people to look at it.

Thank you!

nyorain:
AAudio has no volume concept but `stream_set_volume` should be easy to implement by manually multiplying each sample with the volume just like e.g. the alsa backend does it. I will add that. `stream_get_latency` is harder, we have to employ guesswork based on the buffer sizes or the jni/media library way that it's done in the OpenSLES backend.

nyorain:
For clarification: we can implement `stream_get_latency` natively once the stream is running, returning (for outputs streams) the actual latency of the last written data. But that will vary over time, e.g. grow after a callback gets triggered and then get reduced over time again. And we can't predict that before the stream is started. Is that a valid implementation of the function?

achronop:
> Is that a valid implementation of the function?

Hmm I don't think that this is what we want here. I believe the jni stuff we do on opensl es are adequate and relatively easy to import. That said, I am not sure how accurate the guessing can be. If you think that it can be good enough, please let us know, to consider it.

padenot:
@nyorain, thank you very much for doing this!

This member is for doing A/V sync, it needs to be the time it take between returning from a callback that has output data to the time one hears this buffer physically. It's a static number, for a particular stream setting and device configuration (but is expected to change if the underlying device changes).

As you can see, this is not the prettiest code, but has been determined to answer a correct number for a variety of use cases: Bluetooth device using hands-free, Bluetooth device using a2dp, normal wired headphones, speakers, external card, etc., and provide good A/V sync there (also for reporting the latency via `AudioContext.outputLatency` so people doing their own visuals can be synchronized).

We can certainly add it after merging this however, you've done enough work here :-).

The volume is simpler, and we sometimes implement it in software as you saw. On some backends, it's implemented using OS APIs, this allows a particular scenario to work better:

- Play at volume 1.0
- Pause
- Change the volume to 0.1
- Play

If implemented in the cubeb backend, and depending on the OS, etc., this can produce a burst of audio at 1.0 (because it's still buffered somewhere in the OS), and then play at 0.1 volume. Using the APIs from the OS solves this (since the volume is applied later in the pipeline).

Implementing in software is what we should do here I think.

Thanks again for working on this!

nyorain:
You're welcome, I'm using this library in my own projects and it works quite well, so thanks for writing it in the first place! 
> If implemented in the cubeb backend, and depending on the OS, etc., this can produce a burst of audio at 1.0 (because it's still buffered somewhere in the OS), and then play at 0.1 volume. Using the APIs from the OS solves this (since the volume is applied later in the pipeline).

The reasoning for including the volume api in cubeb makes sense, I added a software implementation. With AAudio, the scenario you describe shouldn't be a problem (at least not if the restarting of the stream happens after the state callback was called with STOPPED) since stopping an output stream will still first output all buffered data. 

> We can certainly add it after merging this however, you've done enough work here :-).

I don't use the latency functionality or A/V sync and wouldn't be sure how to properly test that the jni latency functions return valid values for this backend so I guess it's indeed the best to move this to a separate PR.

src/cubeb_aaudio.c
```diff
@@ -0,0 +1,1377 @@
...
+    LOG("AAudio input stream buffer capacity: %d", bcap);
+    LOG("AAudio input stream buffer size: %d", bsize);
+    LOG("AAudio input stream buffer rate: %d", rate);
+
+    stm->in_buf = malloc(bcap * frame_size);
+
+    // NOTE: not sure about this.
+    // But when this stream contains input and output stream, their
+    // sample rates have to match, don't they?
+    assert(!target_sample_rate || target_sample_rate == input_stream_params->rate);
```
padenot: Yes, this is guaranteed by the function calling this.

nyorain:
Yes, this is ready for review. Just one minor thing left (that i know of): there currently can be crashes (failed ringbuffer assertion) when restarting an app with this backend since we use ALOGV. That seems to use a static/global variable and those are not always reset when the app restarts on android (but the audio thread will change obviously, triggering the invalid thread_id assertion in the async logger ringbuffer). I'll either have to remove the calls to ALOGV or explicitly reset the thread ids in the static ringbuffer in aaudio_init somehow. If we just consider ALOGV broken on android and move on (it's equally broken for the other android backends, note how it's never used there), that should at least be documented somewhere I guess.

If you find the current threaded solution for handling stream state too complicated (using an additional notifier thread), I could revert a6422f61. See the commit message for more details (and a link to a post on the problem I did). Before this commit, we just used constant looping with short sleeping in between.

src/cubeb.c
```diff
@@ -217,6 +224,11 @@ cubeb_init(cubeb ** context, char const * context_name, char const * backend_nam
 #endif
 #if defined(USE_OPENSL)
     opensl_init,
+#endif
+    // TODO: should probably be preferred over OpenSLES when available.
+    // Initialization will fail on old android devices.
```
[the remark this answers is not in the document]
nyorain: Yes, I did that for testing and for my own cubeb builds. But should it already be done here at this stage of the backend?

achronop:
> If we just consider ALOGV broken on android and move on (it's equally broken for the other android backends, note how it's never used there), that should at least be documented somewhere I guess.

Yeah, don't spend too muh time with that. If you have a handy solution go for it otherwise just remove the call and go on. The logic is not broken but there are limitations and they are not your fault. Please, document it as you said.

achronop:
Sorry, my review is not ready. I meant to make an individual comment about the bug. The two comments above are real but more will come.

achronop:
@nyorain, I imported the backend in firefox and attempted a youtube session. It [crashed](https://crash-stats.mozilla.org/report/index/632929d3-3093-498c-b9d0-ccdd10200115) any idea what this night be? How did you test locally?

nyorain:
The backend was tested with my own dummy audio apps (that cover all cases like stop, drain, input + output or seperate streams). I could look into setting up a build environment for firefox on android but I expect that to be a lot of work, not sure when I'll have time for that tbh. Without symbols or android log I have no idea what the issue is, unfortunately.

achronop:
> The backend was tested with my own dummy audio apps (that cover all cases like stop, drain, input + output or seperate streams).

Is it something that you can send me privately and speed up my testing?

> I could look into setting up a build environment for firefox on android but I expect that to be a lot of work, not sure when I'll have time for that tbh. Without symbols or android log I have no idea what the issue is, unfortunately.

No need to spend time on that. We can go on as is for now and debug the Firefox case later.

bernhl:
Did the crash occur on a device with API level 26 (i.e. Android 8.0)? According to https://github.com/google/oboe/blob/master/docs/AndroidAudioHistory.md#80-oreo---api-26-august-21-2017 AAudio is not recommended on 8.0 as it might trigger a crash

nyorain:
Damn, completely forgot about this. Would still like to help you getting this into cubeb as I'm still using the library.
> Is it something that you can send me privately and speed up my testing?

Ok so I wrote up a "minimal" testing example. Minimal in this case still means it has dependencies since setting up logging and message loop and all that on android just requires some code (I also wanted the example to be testable on non-android system for comparison). The code is [here](https://github.com/nyorain/tokonoma/blob/master/src/playground/daudio/cubeb4.cpp). It tests almost all features of cubeb I guess, starting/stopping/draining the stream. When tapping the screen (or clicking the window on desktop systems) it toggles between recording audio and playing back the recorded audio. In my quick testing it behaved the same as the pulseaudio backend (except I'm getting a lot of `AAudioStream_read returned not enough frames` warnings, does not seem to be a problem but should probably look into it).

I uploaded the apk I tested with [here](https://github.com/nyorain/tokonoma/releases/download/dummy/daudiotest.tkn.apk). I have full understanding if you don't want to execute random APKs though. To make it work you would also have to manually enable the 'record audio' permission for the app in android settings (since it's a pure NDK app), otherwise creating the AAudio input stream will fail and the example throw an exception. In case you want to compile it yourself, I'll gladly help with issues. The android-log tag it uses is `daudiotest`.

> Did the crash occur on a device with API level 26 (i.e. Android 8.0)?

Yeah maybe that's a problem, I think I tested it with Android 9 back when I created the PR and with Android 10 today.

padenot:
I've imported this into a Firefox build, it works well, I haven't seen crashes, but I'm running Android 9 on a Sony XZ1 Compact (I hear there are issues on Android 8). I should receive a Pixel and a cheaper phone running Android 8 in the next few days.

@nyorain Thanks a lot for this. It's largely working, but some work is needed:
- Convert to C++, remove all manual lock use (prefer RAII), use smart pointers, generally rely on the type system more, use `std::atomic`, use `std::condvar` and lock, etc. This will make the code terser and safer.
- Add missing functions so that we can have lower latency (better for interactivity, and to make it on par with OpenSL). This probably requires JNI or other ways to get info from the OS. Also add stream type handling to pick the right device (front mic, camera mic, built-in aec/noise suppression/etc.). Roundtrip latency with this backend (that doesn't do anything special yet) is about 135ms (on my Sony XZ1 Compact). It's less than 50ms with our previous backend.
- Fix or remove todos as appropriate, verify other things, etc. In particular it looks like the [state change](https://github.com/kinetiknz/cubeb/pull/559/files#diff-dac779abe308bc2d429483dd37c8adceR402) is now using a regular `nanosleep` with `CLOCK_MONOTONIC` here https://android.googlesource.com/platform/frameworks/av/+/refs/heads/master/media/libaaudio/src/utility/AudioClock.h#89.
- Stress test it, with our usual array of tests: A/V sync with various bluetooth devices, device switch, interruption via notifications, run a lot of streams at once, overload the audio callback thread, compute roundtrip latency, etc.

You've done most of the work, and we need this for Firefox (OpenSL ES [is deprecated in NDK r21d](https://github.com/android/ndk/wiki/Changelog-r21#r21d), and we'd rather have a bullet-proof AAudio backend sooner than later), I'd be happy to finish this and test on more devices, with more involved test scenarios, etc.

nyorain:
I'll gladly port it to C++ (am more used to it anyways) and fix the remaining TODOs, will be looking into the clock issue. Should be able to do this until next week.
Regarding the latency functionality and tests however, I'm likely not the right person to get that right since I don't have much experience in that regard (haven't needed it for my usecases of cubeb so far) and am also not exactly eager to do JNI stuff tbh.

padenot:
>  Regarding the latency functionality and tests however, I'm likely not the right person to get that right since I don't have much experience in that regard (haven't needed it for my usecases of cubeb so far) and am also not exactly eager to do JNI stuff tbh.

For sure, it's not particularly exciting. We'll take care of this part, or any part you don't feel like working on, just let us know where  and when we can help in any way.

nyorain:
Updated the backend to C++11, let me know if it's what you intended.
Regarding the high latency I remembered that I hid the low latency aaudio mode(s) behind config macros that were disabled by default:

One config option specifies whether we pass the low latency flag to the aaudio stream builder (aaudio documentation states that this flag might have an impact on battery life). I enabled that config option by default in cmake now and would consider it sensible to leave it like this but it's up to you what should be enabled by default in cubeb in the end. In my tests, it significantly decreased the output buffer size.

The other flag (that I left disabled but maybe it's useful to you) is to use exclusive sharing mode. The aaudio documentation states that this can reduce the latency further.

padenot:
> Rebased and updated the backend to C++11, let me know if it's what you intended.

From a quick look, this is what I had in mind, thank you very much. I've received test devices, and can test this on Android 8, 9 and 11, now, on a Sony XZ1 Compact, Moto G5 and Pixel 4 devices.

> Regarding the high latency I remembered that I hid the low latency aaudio mode(s) behind config macros that were disabled by default:
> 
> One config option specifies whether we pass the low latency flag to the aaudio stream builder (aaudio documentation states that this flag might have an impact on battery life). I enabled that config option by default in cmake now and would consider it sensible to leave it like this but it's up to you what should be enabled by default in cubeb in the end. In my tests, it significantly decreased the output buffer size.

cubeb usually makes the decision, per platform, depending on the latency requested, at runtime. This allows having a battery preserving mode for playing regular media (videos, etc.), and a "lowest latency possible in non-exclusive" when doing Web Audio API, WebRTC or other interactive scenarios.

> The other flag (that I left disabled but maybe it's useful to you) is to use exclusive sharing mode. The aaudio documentation states that this can reduce the latency further.

We don't use exclusive mode in Firefox, but it might well be useful for others, or for an opt-in mode or something, thanks!

nyorain:
> cubeb usually makes the decision, per platform, depending on the latency requested, at runtime.

Makes sense, could be done here as well. Everything I could do right now is just guesswork though, this should probably be based on queried latencies or general android guarantees. So I would leave it together with the other latency stuff for you. 

Let me know how the testing goes, I'll do what I can to fix remaining errors in the code.

src/cubeb_aaudio.cpp
```diff
@@ -0,0 +1,1359 @@
...
+
+// See https://nyorain.github.io/lock-free-wakeup.html for a note
+// why this is needed. The audio thread notifies the state thread about
+// state changes and must not block. The state thread on the other hand should
+// sleep until there is work to be done. So we need a lockfree producer
+// and blocking producer. This can only be achieved safely with a new thread
+// that only serves as notifier backup (in case the notification happens
+// right between the state thread checking and going to sleep in which case
+// this thread will kick in and signal it right again).
+static void notifier_thread(cubeb* ctx)
```
padenot, line 354: That should work indeed, the cost of a single thread is acceptable.

It's still unclear to me whether notifying a condition variable is real-time safe on all operating systems that are supported by cubeb. The importance of all this is mitigated by the fact that we only need this when draining.

padenot (approved):
This is good to merge as is, I'll quickly followup with lots of little things that are needed before it reaches parity with the opensl backend for all usage, at which point we can move it to have a higher precedence than the opensl backend, that will still be used as fallback. I'm also going to have it run the complete audio-related test suite of Firefox, that usually finds lots of edge-cases.

Thank you very much for doing this and then improving it!

padenot:
Merging this now, and I'll open another PR (or maybe even multiple PRs) for improvements on latency, picking the correct buffer size, A/V sync, etc.

It is build by default when available, but not used by default (opensl is always preferred except when AAudio is requested and available).

PR #560, opened 2019-12-06
Add passthrough resampler no-copy optimization

nyorain:
For input-only streams or duplex streams on backends that always have as much input as output (this is the case for AAudio, i just split this change to a seperate PR), this eliminates a full data copy per callback.

### DIFF
--- a/.clang_complete
+++ b/.clang_complete
@@ -1,0 +1,4 @@
+-D__ANDROID__
+-D__ANDROID_API__=24
+-I./build/exports
+-I/opt/android-ndk/sysroot/usr/include
--- a/.clang_complete
+++ b/.clang_complete
@@ -1,4 +1,0 @@
--D__ANDROID__
--D__ANDROID_API__=24
--I./build/exports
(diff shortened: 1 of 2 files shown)


nyorain:
I agree with all your points, they should be fixed now.

.clang_complete
```diff
@@ -0,0 +1,4 @@
+-D__ANDROID__
+-D__ANDROID_API__=24
+-I./build/exports
+-I/opt/android-ndk/sysroot/usr/include
```
achronop: What is this new file? Why is it needed here?

achronop:
The try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=e7fd306e4c5a52c8e3fb321c438f90c897056c09

.clang_complete
nyorain: Sorry, I added that by accident. Removed it.

achronop:
This [issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1604117) is probably a regression from this patch.

PR #562, opened 2019-12-09
sndio backend fixes

Duncaen:
The first two commits fix two important memory issues.
* First commit allocates the size of `struct cubeb` instead of `struct cubeb *`, which previously would write beyond the allocated memory when assigning `libsndio` and `ops`.
* Second commit fixes writing beyond `struct pollfd[10]`, on my system 16 was required. It now uses `sio_ndfs` to avoid this from happening again.

The third commit is optional but imho improves the choice of which cubeb backend is used if libsndio.so is installed.


kinetiknz:
Thanks for the fixes!

PR #564, opened 2019-12-26
sndio: Always use SIO_DEVANY on FreeBSD/DragonFly

t6:
Forcing cubeb to snd/0 is entirely inappropriate for FreeBSD where
using the sndio backend without a running sndiod is completely ok
and does not suffer from the same drawbacks (blocking the audio
device) as ALSA on Linux.

Using SIO_DEVANY (i.e., "default") will cause libsndio to open the
default device /dev/dsp on FreeBSD, so libsndio can be used without
running sndiod or extra configuration like setting the AUDIODEVICE
environment variable.

Regressed by a7b7cdbeeb79191fbf9c8dbe3934d9d1c5531067.

### DIFF
--- a/src/cubeb_sndio.c
+++ b/src/cubeb_sndio.c
@@ -128,7 +128,7 @@ s16_to_float(void *ptr, long nsamp)
 static const char *
 sndio_get_device()
 {
-#ifndef __OpenBSD__
+#ifdef __linux__
   /*
    * On other platforms default to sndio devices,
    * so cubebs other backends can be used instead.


t6:
Genuinely confused why this was done in cubeb in the first place instead of patching sndio itself on Linux, @Duncaen ?

src/cubeb_sndio.c
```diff
@@ -128,7 +128,7 @@ s16_to_float(void *ptr, long nsamp)
 static const char *
 sndio_get_device()
 {
-#ifndef __OpenBSD__
+#if !defined(__DragonFly__) && !defined(__FreeBSD__) && !defined(__OpenBSD__)
```
jbeich: Maybe include `&& !defined(__FreeBSD_kernel__)` (for Debian GNU/kFreeBSD) as well or switch to `#ifdef __linux__`. Native ALSA only exists on Linux while `SIO_DEVANY` fails if no audio backends are enabled, so NetBSD and Solaris would still fall through to SunAudio backend.

t6: Thanks. Switched it to `#ifdef __linux__` as that makes the most sense I think.

jbeich:
> ... why this was done in cubeb in the first place ... ?

Try building sndio with `--disable-oss` (to mimic Solaris and NetBSD) and cubeb with `-DUSE_ALSA=on` (to get a fallback on FreeBSD). The following is required to make cubeb fall through to another backend. Patching sndio is unlikely to make `cubeb_stream_init` call `cubeb_init` again.
https://github.com/kinetiknz/cubeb/blob/c0a9ea4cc6e2b28af55520b16c5c9149b33cb430/src/cubeb_sndio.c#L335-L340

jbeich:
Nevermind, `SIO_DEVANY` vs. `snd/0` doesn't deny whether to call `sio_open` in `cubeb_init`.

achronop:
@Duncaen can you have a look at the question above please?

Duncaen:
I didn't expect other platforms than OpenBSD to use libsndio for direct input/output to hardware. Patching libsndio on linux seems like the better solution and keeping the old default (always `SIO_DEVANY`) makes more sense in hindsight if linux is the only platform where something like this is necessary.

achronop:
Thanks you!

PR #565, opened 2020-01-09
resampler: handle explicitly the case that in farmes less than out frames

achronop:


achronop:
Hmm are we running the tests on travis? I've just tried them locally and I am getting a failure.

achronop:
Yeah, no testing in Windows, that can be deceiving ...

src/cubeb_resampler.cpp
```diff
@@ -80,6 +80,11 @@ long passthrough_resampler<T>::fill(void * input_buffer, long * input_frames_cou
       // to the callback
       internal_input_buffer.push(static_cast<T*>(input_buffer),
                                  frames_to_samples(*input_frames_count));
+      if (internal_input_buffer.length() < frames_to_samples(output_frames)) {
+        // Grrr glitch!!!
```
kinetiknz: Better comment please :)

achronop:
try run: https://treeherder.mozilla.org/#/jobs?repo=try&revision=a7bed10d6a388605c4b47152aea21b4a6b1113fc

src/cubeb_resampler.cpp
padenot: Yeah, same as below or something.

achronop:
@nyorain, would you like to have a look at it. It's around your code so it would be nice to tell me what you think. Thanks.

src/cubeb_resampler.cpp
```diff
@@ -91,6 +97,17 @@ long passthrough_resampler<T>::fill(void * input_buffer, long * input_frames_cou
       unsigned long samples_off = frames_to_samples(output_frames);
       internal_input_buffer.push(static_cast<T*>(input_buffer) + samples_off,
                                  frames_to_samples(*input_frames_count - output_frames));
+    } else if (*input_frames_count < output_frames) {
```
nyorain: Not important, but it should be possible to combine this and the first `if` branch to remove some redundant code like this:
```c
if (internal_input_buffer.length() != 0 || *input_frames_count < output_frames)
```
The glitch case will already be correctly handled by the check you inserted into that branch I guess.

nyorain:
Otherwise it looks fine. But the conditions you handle will trigger the `assert` at the beginning of the function, you probably want to change that.

src/cubeb_resampler.cpp
achronop: Yeah, I can do that, it will look cleaner.

achronop:
Yeah, we have an extra patch in Firefox to disable that assert. I will merge it here instead of adding it on the top of an libcubeb import. Thank you for the comments.

Issue #567, opened 2020-01-13
Windows test are broken


achronop:
1. test_loopback is failing with an assert:
```
Assertion failed: out_len == output_frame_count, file z:/build/build/src/media/libcubeb/src/cubeb_resampler_internal.h, line 257
```

2. test_ring_buffer hangs

kinetiknz:
test_ring_buffer appears hung, but it's just making progress extremely slowly.  `test_ring_multi` is using `std::this_thread::sleep_for` to sleep for 10us, which I assume is intended to give the other test thread time to run.  With Windows' default timer frequency, each sleep is probably taking ~15.6ms.  Even calling `timeBeginPeriod`/`timeEndPeriod` doesn't help much; doing ~720000 ~1ms sleeps still takes a long time.  On Linux we sleep ~34us and on macOS we get 10us.

Replacing `sleep_for` with `yield`, the test passes on my Windows 10 machine in ~2.5s.  That change also drops the test duration from ~25s to ~1s on my Linux machine and ~7s to ~0.6s on macOS.

PR #569, opened 2020-01-21
test_ring_buffer: Use `yield` instead of `sleep_for(10us)`.

kinetiknz:
See issue #567 for details.

@padenot, this retains (I think) the original intention, but the tests pass without `yield` too, so I'd be just as happy to remove it entirely.


padenot:
Thanks. I'll try intentionally breaking the queue (for example putting a relaxed instead of an acquire or something), this should work (as in, correctly detect that it's now incorrect), this was one of the bugs I had when writing it, and was discovered with this sort of stress test.

I've been investigating other ways to test this kind of scenarios, because we'll need to remove a bunch of locks in gecko (and maybe cubeb, although it's not quite as problematic) this year, I'll follow up, this test kinda works but it's crude.

PR #576, opened 2020-03-23
Remove TODO, we've done all of those, or they don't apply anymore

padenot:


TODO
```diff
@@ -1,41 +0,0 @@
-TODO:
```
achronop: Since the file remains you could leave the title. You could also leave a small description like
```
TODO:
  - Describe what it is missing.

PR #580, opened 2020-04-04
sun: Convert floating-point audio to 32-bit LPCM instead of 16-bit LPCM before passing it to kernel

niacat:
The original implementation converted floats to 16-bit LPCM because that's what the sndio backend did, but it might be more desirable to leverage the NetBSD and illumos kernels' support for 32-bit LPCM instead.


kinetiknz:
Thanks!

PR #581, opened 2020-04-08
Only print the filename when logging

ChunMinChang:
This keeps the logging message consistent after djg/cubeb-rs#50 is merged, 
which fixes the [BMO 162132](https://bugzilla.mozilla.org/show_bug.cgi?id=1628132).

Instead of printing the full path to the file that is currently running,
we should print the filename only. As long as we have the filename and
the line, we know where the log comes from.

### DIFF
--- a/src/cubeb_log.h
+++ b/src/cubeb_log.h
@@ -16,7 +16,11 @@ extern "C" {
 
 #if defined(__GNUC__) || defined(__clang__)
 #define PRINTF_FORMAT(fmt, args) __attribute__((format(printf, fmt, args)))
+#if defined(__FILE_NAME__)
+#define __FILENAME__ __FILE_NAME__
+#else
 #define __FILENAME__ (__builtin_strrchr(__FILE__, '/') ? __builtin_strrchr(__FILE__, '/') + 1 : __FILE__)
+#endif
 #else
 #define PRINTF_FORMAT(fmt, args)
 #define __FILENAME__ __FILE__
--- a/src/cubeb_log.h
+++ b/src/cubeb_log.h
@@ -34,10 +34,10 @@ void cubeb_async_log_reset_threads();
 #define LOGV(msg, ...) LOG_INTERNAL(CUBEB_LOG_VERBOSE, msg, ##__VA_ARGS__)
 #define LOG(msg, ...) LOG_INTERNAL(CUBEB_LOG_NORMAL, msg, ##__VA_ARGS__)
 
-#define LOG_INTERNAL(level, fmt, ...) do {                                   \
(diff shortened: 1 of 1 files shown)


src/cubeb_log.h
```diff
@@ -16,7 +16,9 @@ extern "C" {
 
 #if defined(__GNUC__) || defined(__clang__)
 #define PRINTF_FORMAT(fmt, args) __attribute__((format(printf, fmt, args)))
+#define __FILENAME__ (__builtin_strrchr(__FILE__, '/') ? __builtin_strrchr(__FILE__, '/') + 1 : __FILE__)
```
kinetiknz, line 19: We can probably get away with using __BASE_FILE__ here.  It doesn't have exactly the same semantics as __FILE__, but it should be fine for our case.

Recent clang (>= 10?) also has __FILE_NAME__, which has the same semantics as __FILE__.

src/cubeb_log.h
```diff
@@ -16,7 +16,9 @@ extern "C" {
...
 #else
+#define __FILENAME__ __FILE__
```
kinetiknz, line 21: Move this below PRINTF_FORMAT - please keep the order of declarations the same on both sides of the conditional.

It'd be nice to have a solution for MSVC, at least, here.

src/cubeb_log.h
ChunMinChang: > We can probably get away with using **BASE_FILE** here. It doesn't have exactly the same semantics as **FILE**, but it should be fine for our case.

I've tried `__BASE_FILE__` but the output of `__BASE_FILE__` is same as `__FILE__`.

> Recent clang (>= 10?) also has **FILE_NAME**, which has the same semantics as **FILE**.

This might be a good option. I can do something like:
```c
#if defined(__FILE_NAME__)
#define __FILENAME__ __FILE_NAME__
#else
#define __FILENAME__ (__builtin_strrchr(__FILE__, '/') ? __builtin_strrchr(__FILE__, '/') + 1 : __FILE__)
#endif
```
in `#if defined(__GNUC__) || defined(__clang__)`'s block

kinetiknz: Ah sorry, my quick test of `__BASE_FILE__` was flawed.

src/cubeb_log.h
ChunMinChang: > Move this below PRINTF_FORMAT - please keep the order of declarations the same on both sides of the conditional.
Thanks for the catch!

> It'd be nice to have a solution for MSVC, at least, here.

The following could work.
```c
#include <string.h>
#define __FILENAME__ (strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__)
```

PR #583, opened 2020-04-24
Add a method to get audio input latency on a stream

padenot:
This is quite basic, only includes an implementation for windows (that I can't even compile, I'm getting a VM and will update this, the general idea is there).

An implementation for macOS is upcoming on `cubeb-coreaudio-rs`, and an update to `cubeb-rs` and associated crates is also coming since this adds a public method.

We need to take the resampling latency into account, which is not done at the moment, this will come after, I want to rework bits of this to have less latency, more control, and a more accurate reporting.

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -559,6 +559,16 @@ CUBEB_EXPORT int cubeb_stream_get_position(cubeb_stream * stream, uint64_t * pos
     @retval CUBEB_ERROR */
 CUBEB_EXPORT int cubeb_stream_get_latency(cubeb_stream * stream, uint32_t * latency);
 
+/** Get the input latency for this stream, in frames. This is the number of
+    frames between the time the audio input devices records the data, and they
+    are available in the data callback.
+    This returns CUBEB_ERROR when the stream is output-only.
+    @param stream
+    @param latency Current approximate stream latency in frames.
+    @retval CUBEB_OK
+    @retval CUBEB_ERROR_NOT_SUPPORTED
+    @retval CUBEB_ERROR */
+CUBEB_EXPORT int cubeb_stream_get_input_latency(cubeb_stream * stream, uint32_t * latency);
 /** Set the volume for a stream.
     @param stream the stream for which to adjust the volume.
     @param volume a float between 0.0 (muted) and 1.0 (maximum volume)
--- a/src/cubeb_opensl.c
+++ b/src/cubeb_opensl.c
@@ -1752,6 +1752,7 @@ static struct cubeb_ops const opensl_ops = {
   .stream_reset_default_device = NULL,
   .stream_get_position = opensl_stream_get_position,
   .stream_get_latency = opensl_stream_get_latency,
+  .stream_get_input_latency = NULL,
   .stream_set_volume = opensl_stream_set_volume,
   .stream_get_current_device = NULL,
   .stream_device_destroy = NULL,
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -89,6 +89,9 @@ DEFINE_PROPERTYKEY(PKEY_Device_InstanceId,      0x78c34fc8, 0x104a, 0x4aca, 0x9e
 #endif
 
 namespace {
+
+const int64_t LATENCY_NOT_AVAILABLE_YET = -1;
+
 struct com_heap_ptr_deleter {
   void operator()(void * ptr) const noexcept {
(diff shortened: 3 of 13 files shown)


src/cubeb_opensl.c
```diff
@@ -1705,6 +1705,18 @@ opensl_stream_get_latency(cubeb_stream * stm, uint32_t * latency)
...
+{
+  assert(stm);
+  assert(latency);
+
+  uint32_t stream_latency_frames =
+    stm->user_output_rate * (stm->output_latency_ms / 1000);
+
+  return stream_latency_frames + cubeb_resampler_latency(stm->resampler);
+}
+
```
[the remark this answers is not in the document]
padenot: Wrong version of the branch indeed.

include/cubeb/cubeb.h
```diff
@@ -559,6 +559,16 @@ CUBEB_EXPORT int cubeb_stream_get_position(cubeb_stream * stream, uint64_t * pos
...
+/** Get the input latency for this stream, in frames. This is the number of
+    frames between the time the audio input devices records the data, and they
+    are available in the data callback.
+    This returns CUBEB_ERROR when the stream is output-only.
+    @param stream
+    @param latency Current approximate stream latency in frames.
+    @retval CUBEB_OK
+    @retval CUBEB_ERROR_NOT_SUPPORTED
+    @retval CUBEB_ERROR */
+CUBEB_EXPORT int cubeb_stream_get_input_latency(cubeb_stream * stream, uint32_t * latency);
```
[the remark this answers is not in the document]
padenot: I don't know if I want to break ABI. I don't find other users of this particular method, we could change it.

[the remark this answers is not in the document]
kinetiknz: I'm fine with the current naming.  It's not ideal, but renaming the existing function to `cubeb_stream_get_output_latency`, aside from breaking API compat, makes *that* inconsistent with the remaining stream operations (`cubeb_stream_get_position`, et al.) that only apply the the output stream.

If we wanted to do something about consistency, adding new stream operations that take a `cubeb_device_type` to indicate which stream device/direction to operate on might be better than adding both input and output variants of each one.

src/cubeb_wasapi.cpp
```diff
@@ -1527,6 +1539,11 @@ int wasapi_init(cubeb ** context, char const * context_name)
     return CUBEB_ERROR;
   }
 
+  if (QueryPerformanceFrequency(&ctx->performance_counter_frequency)) {
+    LOG("Failed getting performance counter frequency, latency reporting will be innacurate");
```
kinetiknz, line 1543: nit: typo

src/cubeb_wasapi.cpp
```diff
@@ -2614,6 +2631,22 @@ int wasapi_stream_get_latency(cubeb_stream * stm, uint32_t * latency)
...
+{
+  XASSERT(stm && latency);
+
+  if (!has_input(stm)) {
+    return CUBEB_ERROR;
+  }
+
+  auto_lock lock(stm->stream_reset_lock);
+
+  *latency = hns_to_frames(stm, stm->input_latency_hns);
```
kinetiknz, line 2645: If `stm->input_latency_hns` has never been computed before this call, we'll return CUBEB_OK with a latency of 0.  It seems like it'd be better for this function to fail or somehow indicate the latency value is unreliable.

padenot: Agreed, I'm changing this before merging.

padenot:
Sorry for all the pushes, etc., I don't have a windows machine and I'm locked at home so I used the CI to compile. This is now green, but I can't test, I'm getting a VM.

src/cubeb_wasapi.cpp
```diff
@@ -1527,6 +1547,16 @@ int wasapi_init(cubeb ** context, char const * context_name)
     return CUBEB_ERROR;
   }
 
+
+
```
kinetiknz, line 1551: nit: extra newlines snuck in here

ChunMinChang:
> Sorry for all the pushes, etc., I don't have a windows machine and I'm locked at home so I used the CI to compile. This is now green, but I can't test, I'm getting a VM.

I just got a Windows but I haven't setup it. I can help you to test something if you need.

padenot:
> I just got a Windows but I haven't setup it. I can help you to test something if you need.

It turns out that it works in a VM. The audio latency is insane (roundtrip is around 500ms), but at least it works. Thanks!

PR #585, opened 2020-04-30
Use the same workaround in get_latency than in get_position, in cubeb_wasapi.cpp

padenot:


padenot:
I see this with the default setup on my XPS 15 (I don't have the machine to check the driver, sorry), and the default on the [Microsoft-provided Windows 10 virtual machines](https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/), with an audio device added in VMWare Fusion and no other changes.

PR #586, opened 2020-05-07
Allow setting the voice flag for input OR output, depending on the param struct that has the flag

padenot:
Some GeckoView users don't want to output routing to be changed when an input stream of type voice is requested but would still like to use the correct mic. With this patch, we can request normal audio routing for output (`MUSIC` the default), while still telling Android that the input is a voice stream.

This helps fixing [BMO#1626081](https://bugzilla.mozilla.org/show_bug.cgi?id=1626081).


achronop:
With the minor build error solved.

PR #590, opened 2020-05-28
Add input stream latency test

Koalab99:
It calls the stream ops and ask for input latency if the device
has type INPUT.

### DIFF
--- a/tools/cubeb-test.cpp
+++ b/tools/cubeb-test.cpp
@@ -57,8 +57,8 @@ public:
   void set_latency_testing(bool on);
   void set_latency_frames(uint32_t latency_frames);
   uint64_t get_stream_position() const;
-  uint32_t get_stream_latency() const;
-  uint32_t get_input_stream_latency() const;
+  uint32_t get_stream_output_latency() const;
+  uint32_t get_stream_input_latency() const;
   uint32_t get_max_channel_count() const;
 
   long user_data_cb(cubeb_stream* stm, void* user, const void* input_buffer,
@@ -195,7 +195,7 @@ uint64_t cubeb_client::get_stream_position() const {
   return pos;
 }
 
-uint32_t cubeb_client::get_stream_latency() const {
+uint32_t cubeb_client::get_stream_output_latency() const {
   uint32_t latency = 0;
   int rv = cubeb_stream_get_latency(stream, &latency);
   if (rv != CUBEB_OK) {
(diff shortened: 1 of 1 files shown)


padenot:
@achronop if you can have a look.

tools/cubeb-test.cpp
```diff
@@ -439,6 +450,10 @@ bool choose_action(cubeb_client& cl, operation_data * op, int c) {
     uint64_t pos = cl.get_stream_position();
     uint64_t latency = cl.get_stream_latency();
     fprintf(stderr, "stream position %" PRIu64 " (latency %" PRIu64 ")\n", pos, latency);
+    if(op->collection_device_type & CUBEB_DEVICE_TYPE_INPUT) {
+      latency = cl.get_input_stream_latency();
+      fprintf(stderr, "input stream latency %" PRIu64 ")\n", latency);
```
achronop, line 455: I see a parenthesis close `)` but not a parenthesis open in the message.

That said can you please make it in an away that it will print in one line. Routhly something like:
```
get possition;
fprintf("position message"); // note no endl '\n'
if (DUBLEX || PLAYBACK)
  get output latency
  fprintf("output latency"); // again no endl 
if (DUBLEX || RECORD)
  get input latency
  fprintf("input latency"); // again no endl
fprinf("\n");

File: tools/cubeb-test.cpp [hunk 2]
```diff
@@ -58,6 +58,7 @@ class cubeb_client final {
   void set_latency_frames(uint32_t latency_frames);
   uint64_t get_stream_position() const;
   uint32_t get_stream_latency() const;
+  uint32_t get_input_stream_latency() const;
```
achronop, line 61: Can you change the `get_input_stream_latency()` to `get_stream_input_latency`.

Can you please also change the `get_stream_latency()` to `get_stream_output_latency` for clarity. This is a wrapper not cubeb itself so we can change the ABI without tears.

tools/cubeb-test.cpp
```diff
@@ -439,6 +450,10 @@ bool choose_action(cubeb_client& cl, operation_data * op, int c) {
     uint64_t pos = cl.get_stream_position();
     uint64_t latency = cl.get_stream_latency();
```
achronop, line 451: `get_stream_*_latency()` return `uint32_t` but we store it in a `uint64_t`. It's not your fault that code was there. Do you mind if you change it to use `uint32_t` and replace the format specifiers in the print messages?

Koalab99:
I think it fixed the issues, let me know if something is wrong with it

PR #591, opened 2020-06-09
Clamp stream_delay calculation to zero

padenot:
For the same reasons we do it in other methods: it's possible that we've
not written enough frames to be greater than the latency.


padenot:
https://bugzilla.mozilla.org/show_bug.cgi?id=1644233

PR #593, opened 2020-06-29
Fixed small memory leak

Filoppi:
Reference:
https://docs.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-isformatsupported

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -1777,10 +1777,11 @@ handle_channel_layout(cubeb_stream * stm,  EDataFlow direction, com_heap_ptr<WAV
   waveformatex_update_derived_properties(mix_format.get());
 
   /* Check if wasapi will accept our channel layout request. */
-  WAVEFORMATEX * closest = nullptr;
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -1802,6 +1802,7 @@ handle_channel_layout(cubeb_stream * stm,  EDataFlow direction, com_heap_ptr<WAV
   } else {
     LOG("IsFormatSupported unhandled error: %lx", hr);
   }
+  CoTaskMemFree(closest);
```
kinetiknz, line 1805: Rather than this, if we declare a new `com_heap_ptr` to take ownership of `closest` directly after the call to `IsFormatSupported` we can rely on RAII for this.

Filoppi: I've done that

kinetiknz: Thank you, merged in https://github.com/kinetiknz/cubeb/commit/ea39471a989d35224cc5905f9467da5a36917827 with a small build fix.

PR #594, opened 2020-06-29
Remove AUDCLNT_STREAMFLAGS_NOPERSIST

Filoppi:
Kind of subjective but I'm not sure why you'd want Windows to not remember the volume and mute setting after restarting the application.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -1827,7 +1827,6 @@ initialize_iaudioclient3(com_ptr<IAudioClient> & audio_client,
 
   // IAudioClient3 doesn't support AUDCLNT_STREAMFLAGS_NOPERSIST, and will return
   // AUDCLNT_E_INVALID_STREAM_FLAG. This is undocumented.
-  flags = flags ^ AUDCLNT_STREAMFLAGS_NOPERSIST;
 
   // Some people have reported glitches with capture streams:
   // http://blog.nirbheek.in/2018/03/low-latency-audio-on-windows-with.html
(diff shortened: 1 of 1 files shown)


Filoppi:
For example, the Dolphin supports multiple audio backends on Windows, one of which is cubeb. If you change the volume settings from Windows while using another backend, it will remember it, but if you use cubeb it won't, which I find weird and undesirable.
But we could add a preference, or we could just patch Dolphin to ignore that flag.

Filoppi:
But why did you put AUDCLNT_STREAMFLAGS_NOPERSIST in the first place? Do you need it for Firefox? Firefox seems to remember the volume it was set to in the Windows Mixer after closing and reopening it, so I don't understand how that could be if it uses cubeb.
And what about other platforms? Do they also not save volume changes between application restarts?

src/cubeb_wasapi.cpp
```diff
@@ -1827,7 +1827,6 @@ initialize_iaudioclient3(com_ptr<IAudioClient> & audio_client,
 
   // IAudioClient3 doesn't support AUDCLNT_STREAMFLAGS_NOPERSIST, and will return
```
Filoppi, line 1828: This is actually document (at least now):
https://docs.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient3-initializesharedaudiostream
it says which flags are supported, and that's not in the list. Nor is AUDCLNT_STREAMFLAGS_LOOPBACK

Filoppi: Which is another good reason to take out the AUDCLNT_STREAMFLAGS_NOPERSIST flag now, as when the IAudioClient3 implementation is enabled again, that flag won't be supported.

padenot:
> But why did you put AUDCLNT_STREAMFLAGS_NOPERSIST in the first place? Do you need it for Firefox? Firefox seems to remember the volume it was set to in the Windows Mixer after closing and reopening it, so I don't understand how that could be if it uses cubeb.

I'm very confused about this. I just tested, and indeed, the volume is preserved accross restart. Either I don't understand this flag, or it's broken.

In any case, the patch in #599 looks good.

padenot:
Closing this in favor of #599, where I've asked a question.

PR #598, opened 2020-07-27
Some windows fixes

padenot:
A series of loosely related fixes that improve duplex audio stream, in particular for bluetooth devices that support both A2DP and handsfree, to stop picking the A2DP device when using the mic, to have lower latency and more importantly because it seems like it doesn't work to have the mic in handsfree and the headphones themselves in A2DP (that would be weird anyways).

I'll add commits, because there is a problem with the resampling code (apparently a bad rounding that leads to premature draining of the stream), and I'm not sure yet about the default device matching policy. I might make it opt-in except when we know it makes sense (we might be able to detect that it's a bluetooth device, etc.). I'll also look at what other communication software do.

The review can start now, though, the remaining commits will be small adjustments.

### DIFF
--- a/src/cubeb_resampler_internal.h
+++ b/src/cubeb_resampler_internal.h
@@ -35,6 +35,7 @@ MOZ_END_STD_NAMESPACE
 #include "cubeb_utils.h"
 #include "cubeb-speex-resampler.h"
 #include "cubeb_resampler.h"
+#include "cubeb_log.h"
 #include <stdio.h>
 
 /* This header file contains the internal C++ API of the resamplers, for testing. */
@@ -528,6 +529,7 @@ cubeb_resampler_create_internal(cubeb_stream * stream,
       (output_params && output_params->rate == target_rate)) ||
       (input_params && !output_params && (input_params->rate == target_rate)) ||
       (output_params && !input_params && (output_params->rate == target_rate))) {
+    LOG("Input and output sample-rate match, target rate of %dHz", target_rate);
     return new passthrough_resampler<T>(stream, callback,
                                         user_ptr,
                                         input_params ? input_params->channels : 0,
@@ -578,6 +580,7 @@ cubeb_resampler_create_internal(cubeb_stream * stream,
   }
 
   if (input_resampler && output_resampler) {
+    LOG("Resampling input (%d) and output (%d) to target rate of %dHz", input_params->rate, output_params->rate, target_rate);
     return new cubeb_resampler_speex<T,
                                      cubeb_resampler_speex_one_way<T>,
                                      cubeb_resampler_speex_one_way<T>>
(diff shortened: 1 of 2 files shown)


padenot:
With those patches on top, a duplex stream works with a Sony WH1000 xm3. It appears that there is a drift between input and output (about 1% on the windows 10 laptop I'm using with this headset), so this underruns somewhat on the input, but it's certainly usable. I can reproduce those underruns with other software, so it must be the device or a driver.

Increasing the input latency reduces quite a lot the input underruns. https://github.com/kinetiknz/cubeb/pull/598/commits/76a1bd39cfd1d6884617e10d6d3030f6659caa45 implements a patch that is in Gecko in a cleaner way (as it can clearly happens when there is a drift between input and output).

https://github.com/kinetiknz/cubeb/pull/598/commits/36db54248094ff892e942baa1ae35c0fa2ba8aeb adds a couple things so that this behaviour change in picking the input device and increasing latency, etc., only happens when using an input device that is using the handsfree protocol (all the other devices I've tested work with lower latency and don't need to match the output device).

The right solution to remove the remaining glitches caused by the input underrun is to implement a drift compensator, that we also need for other configurations (macOS, depending on the input or output device choice).

src/cubeb_wasapi.cpp
```diff
@@ -2030,14 +2030,41 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
...
+    return CUBEB_ERROR;
+  }
+
+  cubeb_device_info device_info;
+  int rv = wasapi_create_device(stm->context, device_info, stm->device_enumerator.get(), device.get());
+  if (rv != CUBEB_OK) {
+    LOG("Could not get cubeb_device_info.");
+  }
+
+  REFERENCE_TIME latency = stm->latency;
```
kinetiknz, line 2052: `stm->latency` is in frames, needs conversion?

src/cubeb_wasapi.cpp
```diff
@@ -2030,14 +2030,41 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
...
+  REFERENCE_TIME minimum_period;
+  REFERENCE_TIME default_period;
+  hr = audio_client->GetDevicePeriod(&default_period, &minimum_period);
+  if (FAILED(hr)) {
+    LOG("Could not get device period: %lx", hr);
+    return CUBEB_ERROR;
+  }
+
+  cubeb_device_info device_info;
+  int rv = wasapi_create_device(stm->context, device_info, stm->device_enumerator.get(), device.get());
```
kinetiknz, line 2047: Need to clean up `device_info` to avoid potential leaks.

src/cubeb_resampler_internal.h
```diff
@@ -192,6 +192,21 @@ class cubeb_resampler_speex_one_way : public processor {
...
     assert(r == RESAMPLER_ERR_SUCCESS && "resampler allocation failure");
+
+    uint32_t ratio_num, ratio_den;
+    speex_resampler_get_ratio(speex_resampler, &ratio_num, &ratio_den);
+    uint32_t input_latency = speex_resampler_get_input_latency(speex_resampler);
+    int64_t skip_frac_num = input_latency * ratio_den;
+    T input_buffer[8192] = {};
+    T output_buffer[8192] = {};
+    uint32_t input_frame_count, output_frame_count;
+    input_frame_count = input_latency;
```
kinetiknz, line 203: Assert this is no larger than 8192?

Also initialize these two vars when declared.

kinetiknz (approved):

kinetiknz:
LGTM, thanks!

PR #599, opened 2020-07-29
Add stream preference to request a persistent stream session and implement this in WASAPI.

AdmiralCurtiss:
Like #594 but as an optional thing.

This adds a stream preference flag to request a persistent audio session, so that volume and mute settings remain the same even after closing and reopening a program on Windows.

See https://docs.microsoft.com/en-us/windows/win32/coreaudio/audclnt-streamflags-xxx-constants for details.

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -236,8 +236,12 @@ typedef enum {
                                        selected, as well as the quality of the stream,
                                        for example to accomodate bluetooth SCO modes on
                                        bluetooth devices. */
-  CUBEB_STREAM_PREF_RAW = 0x08  /**< Windows only. Bypass all signal processing
+  CUBEB_STREAM_PREF_RAW = 0x08, /**< Windows only. Bypass all signal processing
                                      except for always on APO, driver and hardware. */
+  CUBEB_STREAM_PREF_PERSIST = 0x10, /**< Request that the volume and mute settings
+                                         should persist across restarts of the stream
+                                         and/or application. May not be honored for
+                                         all backends and platforms. */
 } cubeb_stream_prefs;
 
 /** Stream format initialization parameters. */
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -1857,7 +1857,7 @@ initialize_iaudioclient3(com_ptr<IAudioClient> & audio_client,
 
   // IAudioClient3 doesn't support AUDCLNT_STREAMFLAGS_NOPERSIST, and will return
   // AUDCLNT_E_INVALID_STREAM_FLAG. This is undocumented.
-  flags = flags ^ AUDCLNT_STREAMFLAGS_NOPERSIST;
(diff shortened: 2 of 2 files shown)


padenot:
It seems like this doesn't do anything. Is that your experience as well? I assume you see a change when passing the flag or not passing the flag, otherwise you wouldn't have sent a patch, but I'm trying with Firefox, and we're passing this flag, and the volume I set in `App volume and device preferences` is kept when restarting Firefox.

Filoppi:
> It seems like this doesn't do anything. Is that your experience as well? I assume you see a change when passing the flag or not passing the flag, otherwise you wouldn't have sent a patch, but I'm trying with Firefox, and we're passing this flag, and the volume I set in `App volume and device preferences` is kept when restarting Firefox.

Having or not this flag worked as expected when I tried it on dolphin, though it might not be supported on some kind of WASAPI sessions? Also make sure the Firefox process was completely closed.

padenot:
I'll test harder, thanks for confirming quickly. I'm pretty sure it was closed completely, it was a restart because of a Nightly build update. We do things related to WASAPI sessions, that must be it, i'll try with and without.

AdmiralCurtiss:
Yeah I remember testing this via Dolphin when I made the PR and the behavior was what I expected: Without the `CUBEB_STREAM_PREF_PERSIST` flag, which internally sets the WASAPI `AUDCLNT_STREAMFLAGS_NOPERSIST` flag, the volume settings would reset every time I closed and reopened Dolphin. With the cubeb flag, which clears the WASAPI one, it was kept.

I'll rebase and retest this, just to be sure.

AdmiralCurtiss:
Yeah, still behaves like this. `CUBEB_STREAM_PREF_NONE` resets the volume every time, `CUBEB_STREAM_PREF_PERSIST` keeps it between runs.

I'll also note that apparently cubeb currently doesn't compile with `/std:c++latest` ([see here](https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=vs-2019)), had to switch to c++17 which worked fine.

AdmiralCurtiss:
Oh and for reference, this is how we initialize cubeb: https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/AudioCommon/CubebStream.cpp

padenot:
Thanks for rebasing on top of my recent changes.

padenot:
> Oh and for reference, this is how we initialize cubeb:
> 
> https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/AudioCommon/CubebUtils.cpp#L66
> https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/AudioCommon/CubebStream.cpp#L36

I assume you don't have a control over the sampling-rate, that is dictated by the environment? On Android it can make quite a difference, but I don't know if Dolphin runs or is often used there. On other platforms, we resample (either ourselves or via OS facilities), but on Android, it's critical to use the native rate to get a latency and scheduler fast-path.

PR #601, opened 2020-08-05
sun: building fix in sun_stream_get_latency

khng300:
That's all to let it compile on a running OI-hipster 20200504.


kinetiknz:
Thanks!

PR #603, opened 2020-08-11
jack: Output stereo audio if the input source is mono

spinnylights:
This brings the JACK backend into parity with e.g. the Pulse backend, which [has this behavior already](https://github.com/kinetiknz/cubeb/blob/master/src/cubeb_pulse.c#L580). A typical Firefox user [expects this behavior](https://bugzilla.mozilla.org/show_bug.cgi?id=979705), and because Firefox creates a new unique stream at every audio playback init it's difficult to achieve this behavior currently as a JACK user even if you know what you're doing. Other application users (of e.g. music or video players) will overwhelmingly have the same expectation. There may be esoteric cases in which strict 1 input <-> 1 output is desirable even if the source is mono, but if that case should be supported I think it should enabled via some sort of config rather than the default behavior.

I've tested this patch against the `test_audio` test and confirmed that it results in the desired behavior, whereas the current code does not. If y'all agree that mono source -> stereo out should be the expected behavior for all backends, I could look into adding backend-agnostic testing for this.

Since the patch in the main commit resulted in some obvious code duplication, I included a second commit that cleans it up by adding two new functions. Depending on y'all's tastes and long-term plans, y'all may not want that refactoring, so if y'all would like only the first commit I'm happy to toss the second.

### DIFF
--- a/src/cubeb_jack.cpp
+++ b/src/cubeb_jack.cpp
@@ -238,6 +238,22 @@ load_jack_lib(cubeb * context)
   return CUBEB_OK;
 }
 
+static void
+cbjack_connect_port_out (cubeb_stream * stream, const size_t out_port, const char * const phys_in_port)
+{
+  const char *src_port = api_jack_port_name (stream->output_ports[out_port]);
+
+  api_jack_connect (stream->context->jack_client, src_port, phys_in_port);
+}
+
+static void
+cbjack_connect_port_in (cubeb_stream * stream, const char * const phys_out_port, size_t in_port)
+{
+  const char *src_port = api_jack_port_name (stream->input_ports[in_port]);
+
+  api_jack_connect (stream->context->jack_client, phys_out_port, src_port);
+}
+
 static int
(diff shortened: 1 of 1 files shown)


spinnylights:
(Closing and reopening to trigger Travis rebuild per https://github.com/kinetiknz/cubeb/issues/606.)

spinnylights:
Hmm, it still isn't getting run.

spinnylights:
Still no luck, even after a rebase.

padenot:
I don't understand why only one appveyor ran here. I think this patch makes sense, and I can merge it, however I don't have a jack setup so I can't double check.

spinnylights:
Ah, that's nice to hear—thanks for taking a look! If you'd like a hand with that in some way I'd be glad to help, I know jack can be kind of a weird maze. :stuck_out_tongue: Otherwise I'm available if there's any odds-n-ends-type work that needs to happen here.

padenot (approved):

PR #604, opened 2020-08-17
Fix CMake rust path and only use rust deps if needed

geekuillaume:
This change fixes two things:
- Use `PROJECT_SOURCE_DIR` instead of `CMAKE_SOURCE_DIR` to use the `cubeb` directory instead of the root directory when the cmake project is included by another cmake project.
- Only build the rust version of pulse and audiounit backend if the C version of the backend was also going to be built.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,7 +231,7 @@ if(USE_KAI)
   target_link_libraries(cubeb PRIVATE kai)
 endif()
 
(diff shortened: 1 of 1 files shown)


padenot:
I think this makes sense, but I don't know enough about CMake. @kinetiknz, can you have a look ?

padenot (approved):

PR #607, opened 2020-09-03
Fix integer underflow/unsigned wrapping.

szanni:
Got pointed here from the Mozilla bug tracker to send in the patch upstream:

This is an attempt at a fix for [firefox bug 1614278](https://bugzilla.mozilla.org/show_bug.cgi?id=1614278) where firefox crashes on a regular on xruns. This has been a long standing issue for me (older than the linked bug report). With this patch applied firefox does run stable again with the jack backend.
Looking at the subtraction logic in the xrun handing I am somewhat surprised this does not crash more often. Might also be related to #154 

--

The number of xrun fragments to skip is added once to ctx->jack_xruns but
then subtracted multiple times (once for each stream) from the same variable.
This causes the unsigned int to "underflow" and wrap around.

- Use unsigned int in all calculations instead of mixing un/signed.
- Remove erroneous and useless subtraction logic
- Replace for loop by simple multiplication
- Remove unneeded casts to float

### DIFF
--- a/src/cubeb_jack.cpp
+++ b/src/cubeb_jack.cpp
@@ -290,9 +290,9 @@ cbjack_xrun_callback(void * arg)
   cubeb * ctx = (cubeb *)arg;
 
   float delay = api_jack_get_xrun_delayed_usecs(ctx->jack_client);
-  int fragments = (int)ceilf( ((delay / 1000000.0) * ctx->jack_sample_rate )
-                             / (float)(ctx->jack_buffer_size) );
-  ctx->jack_xruns += fragments;
+  float fragments = ceilf(((delay / 1000000.0) * ctx->jack_sample_rate) / ctx->jack_buffer_size);
+
+  ctx->jack_xruns += (unsigned int)fragments;
   return 0;
 }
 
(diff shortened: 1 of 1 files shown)


szanni:
Any chance to get a review on this one? @padenot @kinetiknz

padenot (approved):

padenot:
Looks good, sorry for the delay, feel free to ping us directly to get a faster review if we don't get to it quickly enough, especially for things like this that make your life easier.

I'm a bit surprised you're under-running that much, that said. Do you have a setup in which this is expected for some reason?

szanni:
Delays do happen. I know so myself. All good, thanks for looking at it now.

And regarding the xruns: it is because I usually push my hardware to the limits. Meaning I often run my setup like in a live environment: bitwig and/or traktor open + the buffer size at the lowest setting that does _not_ give me xruns. If I then open Firefox additionally - to watch some tutorial or hunt for some new music - it tends to push my laptop beyond its limits. Hence the occasional under run.

padenot:
I see, sounds like a normal behaviour then, considering what is being discussed for the jack backend [in other issues](https://github.com/kinetiknz/cubeb/issues/608).

PR #610, opened 2020-09-15
Reset device when reconfiguring stream

padenot:


padenot:
@kinetiknz I'm merging this now because I'd like to land the full series in Firefox sooner than later (before uplift). I'm adding a smoke test for device reconfigure, only useful for Windows, but it's not the first time we've had issues around this.

PR #611, opened 2020-09-15
Add a test case for WASAPI reconfigure event

padenot:
This would have caught the mistake fixed by https://github.com/kinetiknz/cubeb/commit/a45f6d072f6e17e82a34769ef527f3041dca0259.

### DIFF
--- a/test/test_sanity.cpp
+++ b/test/test_sanity.cpp
@@ -639,6 +639,62 @@ TEST(cubeb, drain)
   do_drain = 0;
 }
 
+TEST(cubeb, device_reset)
+{
+  int r;
+  cubeb * ctx;
+  cubeb_stream * stream;
+  cubeb_stream_params params;
+  uint64_t position;
+
+  r = common_init(&ctx, "test_sanity");
+  ASSERT_EQ(r, CUBEB_OK);
+  ASSERT_NE(ctx, nullptr);
+
+  if (strcmp(cubeb_get_backend_id(ctx), "wasapi")) {
+    // cubeb_stream_reset_default_device is only useful and implemented in the
+    // WASAPI backend.
+    return;
+  }
+
+  params.format = STREAM_FORMAT;
+  params.rate = STREAM_RATE;
+  params.channels = STREAM_CHANNELS;
+  params.layout = STREAM_LAYOUT;
+  params.prefs = CUBEB_STREAM_PREF_NONE;
+
+  r = cubeb_stream_init(ctx, &stream, "test", NULL, NULL, NULL, &params, STREAM_LATENCY,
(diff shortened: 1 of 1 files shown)


test/test_sanity.cpp
```diff
@@ -639,6 +639,62 @@ TEST(cubeb, drain)
...
+  cubeb * ctx;
+  cubeb_stream * stream;
+  cubeb_stream_params params;
+  uint64_t position;
+
+  r = common_init(&ctx, "test_sanity");
+  ASSERT_EQ(r, CUBEB_OK);
+  ASSERT_NE(ctx, nullptr);
+
+  if (strcmp(cubeb_get_backend_id(ctx), "wasapi")) {
```
achronop, line 654: Optional: you could wrap the test around an `ifdef windows`

test/test_sanity.cpp
```diff
@@ -639,6 +639,62 @@ TEST(cubeb, drain)
...
+  r = cubeb_stream_start(stream);
+  ASSERT_EQ(r, CUBEB_OK);
+
+  uint32_t iterations = 5;
+  uint64_t previous_position = 0;
+  while (iterations--) {
+    r = cubeb_stream_get_position(stream, &position);
+    ASSERT_EQ(r, CUBEB_OK);
+    ASSERT_GE(position, previous_position);
+    delay(100);
```
achronop, line 680: `previous_position` remains `0` all the way through, I would expect to set it at the end of the iteration after the `ASSERT_GE` and reset it to `0` after the loop (in order to use it on the next loop)

test/test_sanity.cpp
```diff
@@ -639,6 +639,62 @@ TEST(cubeb, drain)
...
+  ASSERT_EQ(r, CUBEB_OK);
+
+  iterations = 5;
+  while (iterations--) {
+    r = cubeb_stream_get_position(stream, &position);
+    ASSERT_EQ(r, CUBEB_OK);
+    ASSERT_GE(position, previous_position);
+    delay(100);
+  }
+
```
achronop, line 693: `cubeb_stream_stop` before destroy ??

achronop (approved):
Looks good to me with the nonoptional comments set.

test/test_sanity.cpp
padenot: Not resetting it in between the two loops allows checking that it's still increasing after a device change, which is something we really want to make sure we do.

achronop: Right, I missed that!

achronop: You could do everything in one while loop then to avoid repetition (optional).

padenot: This will do, thanks!

PR #613, opened 2020-09-20
Add an option to not automatically connect ports on jack.

derekdreery:
This patch adds a perf to disable auto jack port connection. This is useful if the user of the lib wants to coordinate connection management themselves, for example linking into a mixing app rather than straight to audio_out.

### DIFF
--- a/src/cubeb_jack.cpp
+++ b/src/cubeb_jack.cpp
@@ -907,10 +907,12 @@ cbjack_stream_init(cubeb * context, cubeb_stream ** stream, char const * stream_
     }
   }
 
-  if (cbjack_connect_ports(stm) != CUBEB_OK) {
-    pthread_mutex_unlock(&stm->mutex);
-    cbjack_stream_destroy(stm);
-    return CUBEB_ERROR;
+  if (!input_stream_params->prefs & CUBEB_STREAM_PREF_JACK_NO_AUTO_CONNECT) {
+    if (cbjack_connect_ports(stm) != CUBEB_OK) {
+      pthread_mutex_unlock(&stm->mutex);
+      cbjack_stream_destroy(stm);
+      return CUBEB_ERROR;
+    }
   }
 
   *stream = stm;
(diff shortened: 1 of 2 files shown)


src/cubeb_jack.cpp
```diff
@@ -888,10 +888,12 @@ cbjack_stream_init(cubeb * context, cubeb_stream ** stream, char const * stream_
     }
   }
 
-  if (cbjack_connect_ports(stm) != CUBEB_OK) {
-    pthread_mutex_unlock(&stm->mutex);
-    cbjack_stream_destroy(stm);
-    return CUBEB_ERROR;
+  if (!input_stream_params->prefs & CUBEB_STREAM_PREF_JACK_NO_AUTO_CONNECT) {
```
derekdreery, line 891: Double-negative is unfortunate, but necessary to ensure that the default prefs (zeroed mem) are backwards compatible.

padenot:
Can you rebase over recent changes? I'll do it and merge this on monday otherwise.

derekdreery:
I'm not on my linux computer RN, but will try to get on it this weekend if I can.

derekdreery:
rebased.

padenot (approved):

padenot:
Thanks!

szanni:
How would I re-enable auto connect? Is this a Firefox or Jack thing?

This patch seems to breaks audio for a few users, having to manually connect the ports via a patch bay. (See [No sound with Jack Audio sound server since FF84 upgrade](https://bugs.archlinux.org/task/69014) and [No sound in Firefox 84 using Jack](https://bbs.archlinux.org/viewtopic.php?id=261755))
I personally do not use any fancy jack management software, but plain old `jack_control` to start the jack server. Any pointers?

derekdreery:
@szanni this patch left the default to auto-connect. You have to explicitaly set the no auto connect config setting to get the behavior you are describing. If this isn't the case, it's a bug.

szanni:
@derekdreery thatks for confirming. I just wanted to make sure I was not missing something before filing a report. Report is now filed at [https://bugzilla.mozilla.org/show_bug.cgi?id=1698285](https://bugzilla.mozilla.org/show_bug.cgi?id=1698285)

PR #614, opened 2020-09-21
 Clarify and fix expectations around wasapi_create_device, fix crashes

padenot:
Two issues, explained with comments in the code.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -711,8 +711,9 @@ intern_device_id(cubeb * ctx, wchar_t const * id)
   XASSERT(id);
 
   char const * tmp = wstr_to_utf8(id);
-  if (!tmp)
+  if (!tmp) {
     return nullptr;
+  }
 
   char const * interned = cubeb_strings_intern(ctx->device_ids, tmp);
 
@@ -2875,6 +2876,8 @@ wasapi_is_default_device(EDataFlow flow, ERole role, LPCWSTR device_id,
   return ret;
 }
 
+/* `ret` must be deallocated with `wasapi_destroy_device`, iff the return value
+ * of this function is `CUBEB_OK`. */
 int
 wasapi_create_device(cubeb * ctx, cubeb_device_info& ret, IMMDeviceEnumerator * enumerator, IMMDevice * dev)
 {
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -2050,34 +2051,37 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
...
-    LOG("Input is a not bluetooth handsfree, latency %s to %u frames (minimum %u)", latency_frames < minimum_period_frames ? "increased" : "set", latency_frames, minimum_period_frames);
-    latency_frames = std::max(latency_frames, minimum_period_frames);
     stm->input_bluetooth_handsfree = false;
+    latency_hns = frames_to_hns(mix_params->rate, latency_frames);
+    LOG("Could not get cubeb_device_info.");
   }
 
-  REFERENCE_TIME latency_hns = frames_to_hns(device_info.default_rate, latency_frames);
-
-  wasapi_destroy_device(&device_info);
```
padenot, line 2079: This can free an invalid pointer, if the call above to `wasapi_create_device` had failed.

src/cubeb_wasapi.cpp
```diff
@@ -2896,43 +2906,50 @@ wasapi_create_device(cubeb * ctx, cubeb_device_info& ret, IMMDeviceEnumerator *
...
   ret.device_id = device_id_intern;
   ret.devid = reinterpret_cast<cubeb_devid>(ret.device_id);
   prop_variant namevar;
   hr = propstore->GetValue(PKEY_Device_FriendlyName, &namevar);
-  if (SUCCEEDED(hr)) {
+  if (SUCCEEDED(hr) && namevar.vt == VT_LPWSTR) {
     ret.friendly_name = wstr_to_utf8(namevar.pwszVal);
-  } else {
+  }
+  if (!ret.friendly_name) {
```
padenot, line 2952: We want to have a valid pointer to a 0-length string regardless of the sucess of the `GetValue` call, if we have a `nullptr` here (`wstr_to_utf8` can return `nullptr`).

padenot:
Really all this should be better, but I don't really want to rewrite all this in a more modern style if we're moving to rust.

kinetiknz (approved):

PR #615, opened 2020-10-01
Add method to change stream name

mymindstorm:
This PR adds a simple method called `cubeb_stream_set_name` to change a created stream's name so that bug 1120222 can be fixed. I'm new to C/CXX development, so there may be some mistakes.


kinetiknz (approved):
Thanks for this, looks good to me!

PR #616, opened 2020-10-05
 Pass in AUDCLNT_STREAMOPTIONS_RAW when possible when initializing an AudioClient

padenot:
This is night and day in terms of quality on the XPS15 that I'm using, and I'm using the default setup, with the WAVES MaxxAudio app (required on this machine, otherwise, for example, plugging in a pair of analog headphones won't switch the audio output to those headphones). This app can be manually disabled, but otherwise inserts an APO that does noise reduction and echo cancellation, etc., but more importantly doubles ou triples the roundtrip latency.

Additionally, stacking echo cancellers and noise reduction is kind of wasteful.

I suppose we could make this a flag, to then make it controllable externally.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -2102,9 +2102,12 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
     LOG("Could not get cubeb_device_info.");
   }
 
-  if (initialize_iaudioclient2(audio_client) != CUBEB_OK) {
-    LOG("Can't create the reconfigure event, error: %lx", GetLastError());
-    // This is not fatal.
+  if (stm->output_stream_params.prefs & CUBEB_STREAM_PREF_RAW ||
+      stm->input_stream_params.prefs & CUBEB_STREAM_PREF_RAW) {
+    if (initialize_iaudioclient2(audio_client) != CUBEB_OK) {
+      LOG("Can't initialize an IAudioClient2, error: %lx", GetLastError());
+      // This is not fatal.
+    }
   }
 
 #if 0 // See https://bugzilla.mozilla.org/show_bug.cgi?id=1590902
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -1828,7 +1828,7 @@ initialize_iaudioclient2(com_ptr<IAudioClient> & audio_client)
 #endif
   HRESULT hr = audio_client2->SetClientProperties(&properties);
   if (FAILED(hr)) {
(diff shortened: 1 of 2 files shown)


padenot:
Also I'm bumping the required SDK version. Requiring an SDK that is about 7 years old sound OK. I'll also make this a pref.

kinetiknz:
Thanks, this will be great. 

> I'll also make this a pref.

Cheers, I can imagine situations where the APO processing may be necessary or desired.

src/cubeb_wasapi.cpp
```diff
@@ -1812,6 +1812,26 @@ handle_channel_layout(cubeb_stream * stm,  EDataFlow direction, com_heap_ptr<WAV
...
+  if (!audio_client2) {
+    LOG("Could not get IAudioClient2 interface, not setting AUDCLNT_STREAMOPTIONS_RAW.");
+    return CUBEB_OK;
+  }
+  AudioClientProperties properties = { 0 };
+  properties.cbSize = sizeof(AudioClientProperties);
+  properties.Options |= AUDCLNT_STREAMOPTIONS_RAW;
+  HRESULT hr = audio_client2->SetClientProperties(&properties);
+  if (FAILED(hr)) {
+    LOG("Can't create the reconfigure event, error: %lx", GetLastError());
```
kinetiknz, line 1831: This log message should mention SetClientProperties failing.

src/cubeb_wasapi.cpp
```diff
@@ -2082,6 +2102,11 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
     LOG("Could not get cubeb_device_info.");
   }
 
+  if (initialize_iaudioclient2(audio_client) != CUBEB_OK) {
+    LOG("Can't create the reconfigure event, error: %lx", GetLastError());
```
kinetiknz: This log message should mention IAC2 setup failing and falling back.

kinetiknz (approved):

src/cubeb_wasapi.cpp
```diff
@@ -2102,9 +2102,12 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
     LOG("Could not get cubeb_device_info.");
   }
 
-  if (initialize_iaudioclient2(audio_client) != CUBEB_OK) {
-    LOG("Can't create the reconfigure event, error: %lx", GetLastError());
-    // This is not fatal.
+  if (stm->output_stream_params.prefs & CUBEB_STREAM_PREF_RAW ||
+      stm->input_stream_params.prefs & CUBEB_STREAM_PREF_RAW) {
```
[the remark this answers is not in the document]
padenot: Better indeed.

PR #600, opened 2020-08-05
OSS (FreeBSD/illumos) backend

khng300:
This PR implements the OSS backend mainly targeting FreeBSD. The code was based on https://github.com/kinetiknz/cubeb/pull/510.
It has been tested on Illumos as well. (I didn't test the Illumos build with Firefox).
Real tests are done with FreeBSD Firefox build.
Playback is tested by playing YouTube musics.
Recording is tested with https://test.webrtc.org/ and the following project: https://higuma.github.io/web-audio-recorder-js/.

All tests passed except test/test_duplex.cpp.

```
      Start  1: sanity
 1/13 Test  #1: sanity ...........................   Passed   16.67 sec
      Start  2: tone
 2/13 Test  #2: tone .............................   Passed    0.59 sec
      Start  3: audio
 3/13 Test  #3: audio ............................   Passed   17.38 sec
      Start  4: record
 4/13 Test  #4: record ...........................   Passed    0.52 sec
      Start  5: devices
 5/13 Test  #5: devices ..........................   Passed    0.01 sec
      Start  6: callback_ret
 6/13 Test  #6: callback_ret .....................   Passed    1.12 sec
      Start  7: resampler
 7/13 Test  #7: resampler ........................   Passed    1.58 sec
      Start  8: duplex
 8/13 Test  #8: duplex ...........................***Failed    0.52 sec
      Start  9: latency
 9/13 Test  #9: latency ..........................   Passed    0.01 sec
      Start 10: ring_array
10/13 Test #10: ring_array .......................   Passed    0.01 sec
      Start 11: utils
11/13 Test #11: utils ............................   Passed    0.01 sec
      Start 12: ring_buffer
12/13 Test #12: ring_buffer ......................   Passed    2.15 sec
      Start 13: device_changed_callback
13/13 Test #13: device_changed_callback ..........   Passed    0.00 sec

92% tests passed, 1 tests failed out of 13

Total Test time (real) =  40.57 sec

The following tests FAILED:
          8 - duplex (Failed)
Errors while running CTest
```

Thanks @mekanix for the testings with virtual_oss set up as well!

PR #618, opened 2020-10-10
oss: Initialize cubeb_stream::state properly

khng300:
This fixes https://github.com/kinetiknz/cubeb/pull/600#issuecomment-706457745 .


hselasky:
Looks good to me!

khng300:
@padenot We should merge this also, as assert-enabled build would be broken otherwise.

padenot (approved):

PR #619, opened 2020-10-14
Revert "Don't account for unresampled data when computing how much input is needed."

vitor-k:
This reverts commit d0127cdb668fe7908397ba4eaca2dc1e8600fc09.
From testing this caused the audio to get out of sync/delayed in Citra (testing based on citra-emu/citra#5577).
The 3ds' native sample rate is 32728 Hz, which is what Citra uses.

While the drift is small, it is very noticeable over anything but very short play sessions.
From some loose testing, in 10 minutes there's a couple seconds of added delay to e.g. a confirmation sound.

From testing, reverting this commit fixed the issue. I checked the original PR (#598), but I don't see an
explanation for this specific change.

### DIFF
--- a/src/cubeb_resampler_internal.h
+++ b/src/cubeb_resampler_internal.h
@@ -307,9 +307,11 @@ public:
   uint32_t input_needed_for_output(int32_t output_frame_count) const
   {
     assert(output_frame_count >= 0); // Check overflow
+    int32_t unresampled_frames_left = samples_to_frames(resampling_in_buffer.length());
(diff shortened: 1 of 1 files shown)


padenot:
Yep, this is wrong. I was papering over something else. I should have had gotten to the bottom of this.

padenot:
I'm closing this, this patch is now also in #620, and there is an explanation there. Thanks very much for the PR, and sorry about the bug.

PR #620, opened 2020-10-19
Revert "Don't account for unresampled data when computing how much input is needed." and fix the real problem

padenot:
cc @vitor-k

In #619, a patch is being rightfully reverted. I made this because it made cubeb stream with a user rate of 48000 and device rate of 44100 drain instantly, and it made sense at the time somehow, but I was fixing the symptoms and not the cause. This patch causes a drift, as reported in #619.

I'm adding another patch here on top of #619 to fix the initial problem while keeping the revert to unbreak the use-case discussed there. The issue in my particular use-case was that `input_needed_for_output` was asking the user callback for a number of audio frames that was too low, because it's subtracting the frames already buffered. But this is the beginning of the stream, so the number of already buffered frames should be zero, but is not, because we're adding latency frames at construction in the duplex case when rates are not identical, but this makes no sense because streams are in opposite directions.

With this removed, `input_needed_for_output` calls the user callback with a sensible number of frames requested, at the user rate, and resampling back to the device rate before handing the frames to the OS works.

I'm doing a separate PR because I believe I can't put a patch on top of #619, but this is based on the same branch.

### DIFF
--- a/src/cubeb_resampler.cpp
+++ b/src/cubeb_resampler.cpp
@@ -138,17 +138,6 @@ cubeb_resampler_speex<T, InputProcessor, OutputProcessor>
   , user_ptr(ptr)
 {
   if (input_processor && output_processor) {
-    // Add some delay on the processor that has the lowest delay so that the
-    // streams are synchronized.
-    uint32_t in_latency = input_processor->latency();
-    uint32_t out_latency = output_processor->latency();
-    if (in_latency > out_latency) {
-      uint32_t latency_diff = in_latency - out_latency;
-      output_processor->add_latency(latency_diff);
-    } else if (in_latency < out_latency) {
-      uint32_t latency_diff = out_latency - in_latency;
-      input_processor->add_latency(latency_diff);
-    }
     fill_internal = &cubeb_resampler_speex::fill_internal_duplex;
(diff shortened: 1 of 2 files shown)


padenot:
I wonder how I can write a test for this.

src/cubeb_resampler.cpp
```diff
@@ -138,17 +138,6 @@ cubeb_resampler_speex<T, InputProcessor, OutputProcessor>
   , user_ptr(ptr)
 {
   if (input_processor && output_processor) {
-    // Add some delay on the processor that has the lowest delay so that the
-    // streams are synchronized.
-    uint32_t in_latency = input_processor->latency();
-    uint32_t out_latency = output_processor->latency();
-    if (in_latency > out_latency) {
-      uint32_t latency_diff = in_latency - out_latency;
-      output_processor->add_latency(latency_diff);
```
kinetiknz, line 147: This leaves `add_latency` unused, can we remove it now?

kinetiknz (approved):

src/cubeb_resampler.cpp
padenot: yes

PR #621, opened 2020-10-28
Don't use exception in the new `cubeb_aaudio.cpp`, and don't build with exception or RTTI

padenot:
We want to crash here.


padenot:
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4634 looks good, but not widely supported enough, so I did this manually.

kinetiknz (approved):

PR #622, opened 2020-10-30
Bring the AAudio backend to parity with the OpenSL backend

padenot:
This is lots of little commits, implementing missing features that we have in the other Android backend, and other fixes.

With this patch set, the AAudio backend is at parity in terms of round-trip latency and resilience (if not better, depending on the device) compared to the OpenSL backend, on the same device. In a second phase, we'll be able to take advantage of AAudio-specific features or other recent Android audio things (such as unprocessed input streams, or using the hardware AEC).

In particular, it's interesting to note that we can trigger the `FAST_MIXER` fast paths in native code **without** having to do JNI madness or `dlopen`ing `libmedia.so`, but this requires opening a dummy stream.

The latency figures are fairly accurate, and I believe it's possible to make them very accurate.

This is green when pushing to try on the Firefox CI but I'll spend some more time with it before making it the default, after testing on mode devices.

cc @nyorain, if you want to have a look.

### DIFF
--- a/src/cubeb_aaudio.cpp
+++ b/src/cubeb_aaudio.cpp
@@ -679,11 +679,11 @@ realize_stream(AAudioStreamBuilder * sb, const cubeb_stream_params * params,
   switch (params->format) {
     case CUBEB_SAMPLE_S16NE:
       fmt = AAUDIO_FORMAT_PCM_I16;
-      *frame_size = 2 * params->channels;
+      *frame_size = sizeof(int16_t) * params->channels;
       break;
     case CUBEB_SAMPLE_FLOAT32NE:
       fmt = AAUDIO_FORMAT_PCM_FLOAT;
-      *frame_size = 4 * params->channels;
+      *frame_size = sizeof(float) * params->channels;
       break;
     default:
       return CUBEB_ERROR_INVALID_FORMAT;
--- a/src/cubeb_aaudio.cpp
+++ b/src/cubeb_aaudio.cpp
@@ -92,14 +92,14 @@ using unique_lock = std::unique_lock<std::mutex>;
 using lock_guard = std::lock_guard<std::mutex>;
 
 enum class stream_state {
-  init = 0,
-  stopped,
-  stopping,
-  started,
-  starting,
-  draining,
-  error,
-  shutdown,
+  INIT = 0,
+  STOPPED,
+  STOPPING,
+  STARTED,
+  STARTING,
+  DRAINING,
+  ERROR,
+  SHUTDOWN,
 };
 
 struct cubeb_stream {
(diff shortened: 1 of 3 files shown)


nyorain:
LGTM, nice to see you found a way to implement it without jni!

src/cubeb_aaudio.cpp
```diff
@@ -1263,6 +1264,32 @@ aaudio_stream_get_position(cubeb_stream * stm, uint64_t * position)
...
+
+  res = WRAP(AAudioStream_getTimestamp)(stm->ostream, CLOCK_MONOTONIC, &pos, &ns);
+  if (res != AAUDIO_OK) {
+    LOG("aaudio_stream_get_latency, AAudioStream_getTimestamp: %s, returning memoized value", WRAP(AAudio_convertResultToText)(res));
+    // Expected when the stream is paused.
+    *latency = stm->latest_output_latency;
+    return CUBEB_OK;
+  }
+
+  int64_t read = WRAP(AAudioStream_getFramesRead)(stm->ostream);
```
kinetiknz, line 1269: Doesn't look like this is documented to return a result `< 0`, worth asserting just in case?

src/cubeb_aaudio.cpp
```diff
@@ -1290,6 +1291,36 @@ aaudio_stream_get_latency(cubeb_stream * stm, uint32_t * latency)
...
+
+  res = WRAP(AAudioStream_getTimestamp)(stm->istream, CLOCK_MONOTONIC, &pos, &ns);
+  if (res != AAUDIO_OK) {
+    // Expected when the stream is paused.
+    LOG("aaudio_stream_get_input_latency, AAudioStream_getTimestamp: %s, returning memoized value", WRAP(AAudio_convertResultToText)(res));
+    *latency = stm->latest_input_latency;
+    return CUBEB_OK;
+  }
+
+  int64_t written = WRAP(AAudioStream_getFramesWritten)(stm->istream);
```
kinetiknz, line 1297: Same question about assert as for `aaudio_stream_get_latency`.

kinetiknz (approved):
LGTM, thanks!

src/cubeb_aaudio.cpp
kinetiknz: I note this pattern (signed result that would usually only be positive, and doesn't signal errors) appears in other parts of the AAudio API, so perhaps this doesn't matter.

padenot: Probably stems from https://google.github.io/styleguide/cppguide.html#Integer_Types. In any case I traced the calls down into the android code, it goes down here: https://android.googlesource.com/platform/frameworks/av/+/refs/heads/master/media/libaaudio/src/client/AudioStreamInternalPlay.cpp#244, and it's always positive.

PR #623, opened 2020-10-30
Add a `clang-format` file and format `src/cubeb_aaudio.cpp`

padenot:
I kind of tried something based on what I think the style is, but it's probably not finished and @kinetiknz will know best.

This is based on #622 but doesn't contain functional changes, https://github.com/kinetiknz/cubeb/commit/77c4812f89117c01944490ebd3aece9137075da1 has all the formatting changes.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,13 @@
+IndentWidth: 2
+UseTab: Never
+ReflowComments: true
+PointerAlignment: Middle
+AlignAfterOpenBracket: Align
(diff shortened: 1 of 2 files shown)


kinetiknz (approved):
Thanks! Looks like a great starting point, we can always adjust over time as necessary.

If we're moving in the direction of automated formatting, I wonder if it'd be worth trying to align with Gecko's style over time.

PR #634, opened 2021-01-19
oss: Fix CPU spinning 100%

khng300:
As for output-only stream we do not wait on the buffer at all, the CPU usage would spin to 100%.

I would like this PR to be backported to Firefox 85 as well.

### DIFF
--- a/src/cubeb_oss.c
+++ b/src/cubeb_oss.c
@@ -781,6 +781,28 @@ oss_put_play_frames(cubeb_stream * s, unsigned int nframes)
   return 0;
 }
 
+static int
+oss_wait_playfd_for_space(cubeb_stream * s)
+{
+  /* For non-duplex stream we have to wait until we have space in the
+   * buffer */
+  int rate = s->play.info.sample_rate;
+  struct pollfd pfd;
+
+  pfd.events = POLLOUT|POLLHUP;
+  pfd.revents = 0;
+  pfd.fd = s->play.fd;
+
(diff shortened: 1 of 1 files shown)


khng300:
@jbeich Please test this patch on firefox 85, thanks.

jbeich:
Seems to work fine inside 11.4 i386 and 12.1 amd64 jails.

padenot:
> I would like this PR to be backported to Firefox 85 as well.

I've pinged the release managers about this. We're _very_ late in the release process for 85 (5 days until uplift: https://whattrainisitnow.com/), and it's not my call, but if we can, I'll prepare the backport.

padenot (approved):
In any case, I'm merging this now, as it looks fine to my non-knowledgeable eyes, and Jan says it works.

khng300:
Thanks for the help on that request.

Ka Ho

padenot:
They say it's too late, so you'll have to carry a patch for this one. I'm merging it to `mozilla-central` as we speak however: https://bugzilla.mozilla.org/show_bug.cgi?id=1687487.

khng300:
> They say it's too late, so you'll have to carry a patch for this one. I'm merging it to `mozilla-central` as we speak however: https://bugzilla.mozilla.org/show_bug.cgi?id=1687487.

Thanks for the help. I am okay with carrying downstream patches on the ports side.

padenot:
Final update on this: I just queued this for landing to `mozilla-central`.

Issue #638, opened 2021-01-21
Feature proposal: Prefer ALSA backend over others when multiple are enabled.


defcronyke:
I use alsa-jack-sink on Arch, and because the Arch maintainer enables both ALSA and JACK backends, this causes Firefox sound to not work properly on my system.

The reason is because cubeb decides to use the native JACK backend instead of the ALSA backend in this case, since not only is JACK compiled in, but it's also running on my system. This is a perfectly fine default behaviour for any users with a regular JACK setup on their system, however, I really need to be able to tell Firefox to prefer ALSA in this case, so it uses the alsa-jack-sink client and properly follows my ALSA output chain, rather than whatever behaviour of the native JACK client which I prefer to not be using.

To make this work, I implemented an environment variable `CUBEB_PREFER_ALSA`, which if you set it, cubeb will prefer to use the ALSA backend instead of anything else. The way I implemented it, the environment variable is only available if the ALSA backend is compiled into cubeb, so it shouldn't be able to do anything wild.

I would love it if you would consider this feature, since having this would allow me and any other alsa-jack-sink users to be able to use Firefox as its currently being packaged in the Arch Linux PKGBUILD script, and have proper audio output which is respecting our .asoundrc ALSA output chain settings. Anyone using alsa-jack-sink who is also having both the ALSA and the JACK backends compiled in would probably appreciate this.

It can be used as follows:
```shell
CUBEB_PREFER_ALSA=y
```

I'll send the PR now. Thanks for considering.

defcronyke:
You can ignore the first of those two commits mentioned above, something was weird with it so I deleted it.

defcronyke:
Nevermind. For some reason I had tried the media.cubeb.backend = "alsa" setting, but I must have done something else wrong while trying it and it lead me to believe something like the PR would be an appropriate workaround to that config setting not working properly. On second try, the setting is working perfectly, and this feature isn't needed.

PR #639, opened 2021-01-21
Issue #638 - Prefer ALSA backend

defcronyke:
Implement #638

Feature proposal: Prefer ALSA backend over others
when multiple are enabled.

I implemented an environment variable CUBEB_PREFER_ALSA.
It can be used as follows:

  CUBEB_PREFER_ALSA=y

### DIFF
--- a/src/cubeb.c
+++ b/src/cubeb.c
@@ -14,6 +14,14 @@
 
 #define NELEMS(x) ((int) (sizeof(x) / sizeof(x[0])))
 
+#if defined(USE_ALSA)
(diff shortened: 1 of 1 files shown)


defcronyke:
Nevermind. For some reason I had tried the media.cubeb.backend = "alsa" setting, but I must have done something else wrong while trying it and it lead me to believe something like the PR would be an appropriate workaround to that config setting not working properly. On second try, the setting is working perfectly, and this feature isn't needed.

PR #641, opened 2021-02-17
Remove cubeb_stream_reset_default_device API.

kinetiknz:
`cubeb_stream_reset_default_device` was a Windows-only hack added to work around a sandboxing restriction in Gecko content processes.  This is no longer an issue, so we can remove the API.

This is for https://bugzilla.mozilla.org/show_bug.cgi?id=1689517


ChunMinChang (approved):
LGTM!

PR #643, opened 2021-03-25
WASAPI: Use persistent volume and mute settings by default

ggrtk:
Currently, volume and mute settings for the WASAPI backend are not persistent across application restarts unless `CUBEB_STREAM_PREF_PERSIST` is set. However, having to manually set this flag is counterintuitive because the default IAudioClient behavior is that these settings are persistent (https://docs.microsoft.com/en-us/windows/win32/coreaudio/audclnt-streamflags-xxx-constants#remarks).

This commit makes the WASAPI backend use persistent settings by default unless specified by the `CUBEB_STREAM_PREF_NOPERSIST` flag, which replaces `CUBEB_STREAM_PREF_PERSIST`.

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -238,11 +238,10 @@ typedef enum {
                                        bluetooth devices. */
   CUBEB_STREAM_PREF_RAW = 0x08, /**< Windows only. Bypass all signal processing
                                      except for always on APO, driver and hardware. */
-  CUBEB_STREAM_PREF_PERSIST = 0x10, /**< Request that the volume and mute settings
-                                         should persist across restarts of the stream
-                                         and/or application. May not be honored for
-                                         all backends and platforms. */
-
+  CUBEB_STREAM_PREF_NOPERSIST = 0x10, /**< Request that volume and mute settings should
+                                           not persist across restarts of the stream
+                                           and/or application. Only affects the WASAPI
+                                           backend. */
   CUBEB_STREAM_PREF_JACK_NO_AUTO_CONNECT = 0x20  /**< Don't automatically try to connect
                                                       ports.  Only affects the jack
                                                       backend. */
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -2072,8 +2072,7 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
 
   DWORD flags = 0;
 
-  bool is_persist = stream_params->prefs & CUBEB_STREAM_PREF_PERSIST;
-  if (!is_persist) {
+  if (stream_params->prefs & CUBEB_STREAM_PREF_NOPERSIST) {
     flags |= AUDCLNT_STREAMFLAGS_NOPERSIST;
   }
 


padenot:
I think this makes sense. Cubeb is originally from Firefox, and web pages are unrelated to each other so it's non-persist by default (thinking of web pages each as different apps).

But cubeb is now being used by other people and projects more and more, and having a default that makes sense for the general case, and follows the platform default probably is the best.

@kinetiknz what do you think ? We'd then just add this `CUBEB_STREAM_PREF_NOPERSIST` flag to Firefox and all other projets would have better defautls.

kinetiknz:
I'm personally not in favour of changing the default stream configuration for cubeb's API and all backends.  cubeb's API is intended to offer a portable baseline across various OS/API backends and cubeb_streams were always intended to be ephemeral, I'm not sure we could consistently offer persistent stream behaviour for every backend (and/or it could be a considerable amount of work to support internally for backends without OS-level support).  I'd prefer to leave it as a backend-specific pref that each backend is free to reject during cubeb_stream_init if unsupported by that backend.

I'm happy to keep CUBEB_STREAM_PREF_PERSIST for those backends that can support it.  If the existing (expected) stream behaviour and prefs are unclear, we could address that by improving the documentation.

ggrtk:
This change still keeps it as a backend-specific preference that can be ignored. The WASAPI backend would be the only one affected, since none of the other backends currently honor the `CUBEB_STREAM_PREF_PERSIST` flag. I don't know if volume persistence settings can be exposed on the other audio backends as I'm not familiar with them. For Windows applications though, non-persistent volume settings is the exception rather than the norm, and this is reflected in other libraries like SDL2 and OpenAL Soft using persistent volume by default for their WASAPI backends.

ggrtk:
> We'd then just add this `CUBEB_STREAM_PREF_NOPERSIST` flag to Firefox and all other projets would have better defautls.

@padenot Does Firefox actually require non-persistent volume settings on Windows? Firefox 87.0 shows itself as a single application in the Windows volume mixer regardless of how many tabs are playing media, so I'm not sure you would even have to force non-persistent volume, unless I'm overlooking something here. IAudioClient3 also doesn't support non-persistent volumes, if the code path for it is ever reenabled in the future.

padenot:
There are lots of different volumes here.

Then there is a server-side source input gain, is a gain applied for each WASAPI stream of an application, pre-mixer. This is the regular channel fader on a mixing desk, and is what is used when folks change the `volume` attribute on a `<video>`. This is conceptually equivalent to multiplying each audio frame by a gain before returning, in the audio callback. This gain stage we don't want persisted in Firefox (because tabs are unrelated), but we're using the `IAudioStreamVolume` interface, that has nothing to do with volume persistence anyways, so we're good.  This is what `cubeb_stream_set_volume` uses.

Then there is a single volume for Firefox, because it's using `IAudioSessionControl::SetGroupingParam` and friends to group all different audio stream (from, say, different tabs) to a single group. Firefox is a program comprised of lots of processes for security reasons, so it is necessary to tell Windows those audio streams are in fact from the same application. This is an input gain for a group of streams, conceptually, and *it is what folks want persisted*, I assume. Playing two videos and moving this slider affects the volume of two videos. It's like a subgroup volume fader on a mixing desk. This happens outside of cubeb. Firefox never sets this volume however, we just use this facility to show a single volume fader for Firefox regardless of the process architecture. This is persisted across Firefox's restart, because there is no `...NO_PERSIST` flag on `IAudioSessionControl`.

Firefox _could_ change this itself, if it was using the `ISimpleAudioVolume` interface, but it doesn't do so, because we don't currently have a use-case to implement it.

And then there is the global system volume, which is the master fader on a desk, and this we don't touch either.

In the current state of things a program that doesn't do any of this, uses cubeb and wants their volume persisted has two options:
- Pass `CUBEB_STREAM_PREF_PERSIST` on Windows. This is a good option if the program only use a single process or a single audio output stream. It can be argued that this should be the platform's default. It can also be argued that this should be opt-in, if as Matthew states, cubeb is supposed to do the same on all platforms.
- Use the API to group their stream into a group, in which case the `CUBEB_STREAM_PREF_PERSIST` is no-op: the volume of each stream doesn't exist, the group volume is the volume that matters. This is external to cubeb, is what Firefox does and is appropriate for multi-process programs or programs that play multiple audio output stream.

With this patch applied, only the first point is changed (the simple and common case), the volume will be persisted by default. Also, as noted we'll eventually use `IAudioClient3` so it will be the only possible behaviour (cubeb won't be able to choose), but it's in fact already what we do.

I don't actually understand why we use `NO_PERSIST` in cubeb, since `IAudioSessionControl::SetGroupingParam` predates the landing of the WASAPI backend, and we've never touched `ISimpleAudioVolume`. It's present from the initial revision on the file (https://bugzilla.mozilla.org/show_bug.cgi?id=866675), and I think it must have been a mistake on my part at the time, to put it in. @kinetiknz, do you remember maybe ? You reviewed this 8 years ago, but I find no recorded chatter on this, from you or me.

I also remember that there was a bug about having lots of different volume sliders (one per tab playing audio?) at some stage, but I don't remember the details. This might be why `NO_PERSIST` was useful?

I think that removing it or doing anything with it will be no-op for Firefox, but it might be more ergonomic for other programs.

kinetiknz:
Apologies for the delay, and thanks for the extensive write-up Paul!  Given what you've described, I think the simplest approach is to drop `NOPERSIST` entirely.  I've pushed a change to do that in b608f59, which should address @ggrtk's issue.

PR #644, opened 2021-04-27
oss: Use the minimum space of both direction in full-duplex path

khng300:
There is a bug in the OSS cubeb code which use the previous number of
frames as input for the next data-exchange, when full-duplex is
activated. This causes noticable jitter, because the wrong number of
audio frames is exchanged.

The solution is to check both input and output audio DSP buffers and
select the minimum number of audio frames which can be transferred when
full duplex is activated.

### DIFF
--- a/src/cubeb_oss.c
+++ b/src/cubeb_oss.c
@@ -784,16 +784,32 @@ oss_put_play_frames(cubeb_stream * s, unsigned int nframes)
 static int
 oss_wait_playfd_for_space(cubeb_stream * s)
 {
-  /* For non-duplex stream we have to wait until we have space in the
-   * buffer */
-  int rate = s->play.info.sample_rate;
   struct pollfd pfd;
 
-  pfd.events = POLLOUT|POLLHUP;
+  pfd.events = POLLOUT | POLLHUP;
   pfd.revents = 0;
   pfd.fd = s->play.fd;
 
-  if (poll(&pfd, 1, s->nfr * 1000 + rate - 1 / rate) == -1) {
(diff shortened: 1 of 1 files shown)


khng300:
cc @hselasky

hselasky:
Looks good to me. Tested for a while.

src/cubeb_oss.c
```diff
@@ -808,25 +824,19 @@ static int
...
   cubeb_state state = CUBEB_STATE_STOPPED;
-  int trig = 0;
-  int drain = 0;
+  int trig = 0, drain = 0;
   const bool play_on = s->play.fd != -1, record_on = s->record.fd != -1;
-  long nfr = s->bufframes;
+  long nfr = 0;
 
   if (record_on) {
-    if (ioctl(s->record.fd, SNDCTL_DSP_SETTRIGGER, &trig)) {
```
kinetiknz, line 817: The [OSS v4.x docs](http://manuals.opensound.com/developer/SNDCTL_DSP_SETTRIGGER.html) state "_This ioctl call needs to be called twice. Otherwise the call will not have any effect. The first call is used to clear the direction bits. The second call returns them back to enabled which starts the device operation immediately._".  Can you explain why you're removing the first call here?

kinetiknz (approved):
Thanks, looks good to me.  I have one query regarding the SNDCTL_DSP_SETTRIGGER change.

src/cubeb_oss.c
khng300: That sounds unintended. I will add it back.

PR #645, opened 2021-05-05
memory_order_relaxed has changed in c++20

malytomas:
https://en.cppreference.com/w/cpp/atomic/memory_order

`std::memory_order::memory_order_relaxed` has been renamed to `std::memory_order::relaxed` in c++20.
However, `std::memory_order_relaxed` has been added for backwards compatibility and is usable with both versions of the standard.


kinetiknz:
LGTM, thank you!

PR #646, opened 2021-05-24
wasapi: Adjust output stream volume internally rather than using IAudioStreamVolume.

kinetiknz:
Attempt to address [BMO 1624067](https://bugzilla.mozilla.org/show_bug.cgi?id=1624067) by handling volume adjustment internally rather than using IAudioStreamVolume.

If this solves BMO 1624067, I'll remove the CUBEB_WASAPI_USE_IAUDIOSTREAMVOLUME path entirely.


padenot (approved):
Thanks! Hopefully that will do something.

PR #648, opened 2021-06-16
don't run resampler unless we have enough frames (fixes #647)

### LINKED ISSUE #647: cubeb crashes when resampling with RDP audio forwarding
Client: Windows 10 Home (version 2004)
Server: Windows Server 2019 Datacenter (version 1809)

If I create a cubeb input-only stream with a different sample rate than the host (I picked 16khz on a 44.1khz host), when forwarding my microphone over RDP, cubeb will crash after about a minute.

This does not happen with the macOS RDP client. This has been reproduced by two people on different Windows client and server versions.

Minimal reproduction case attached (compiled with mingw-w64, you may need to tweak it a bit if you build with msvc). 

[cubeb_test.c.zip](https://github.com/mozilla/cubeb/files/6662508/cubeb_test.c.zip)

If I run this cubeb_test, after around a minute of streaming, I get 0 frames passed to data_cb, then cubeb crashes with a divide by zero exception. I've traced that to `resampled_frame_count` being zero here: https://github.com/mozilla/cubeb/blob/master/src/cubeb_resampler.cpp#L235

Looks like the issue happens when we get exactly one frame of input from the OS:

```
cubeb_resampler_fill() -> 140
cubeb_resampler_fill(input_frames_count=441, output_frames_needed=0)
output_for_input(441) -> 161
data cb: 161
cubeb_resampler_fill() -> 444
cubeb_resampler_fill(input_frames_count=582, output_frames_needed=0)
output_for_input(582) -> 211
data cb: 211
cubeb_resampler_fill() -> 582
cubeb_resampler_fill(input_frames_count=1, output_frames_needed=0)
output_for_input(1) -> 0
data cb: 0
```

Seems like if the OS doesn't send us enough samples at once to satisfy the resampler, `samples / ratio` will be less than `1.0` and `floor()` will return `0`, causing the divide by zero.

For 44.1khz / 16khz, the ratio is `2.75` input samples to each resampled sample, so getting fewer than 3 samples means a crash. More extreme ratios (which could maybe trigger with a less broken audio device than RDP) would be 48khz / 8khz at 6:1, 96khz / 8khz at 12:1, or 192khz / 8khz at 24:1.

lunixbochs:
if input_frames_count is less than the minimum number of frames required to resample, resampled_frames_count will be zero, and the function will crash with a divide by zero.

For 44.1khz / 16khz, the ratio is 2.75 input samples to each resampled sample, so getting fewer than 3 samples means a crash.

See also https://github.com/mozilla/cubeb/issues/647

### DIFF
--- a/src/cubeb_resampler.cpp
+++ b/src/cubeb_resampler.cpp
@@ -222,6 +222,12 @@ cubeb_resampler_speex<T, InputProcessor, OutputProcessor>
   * callback. */
   input_processor->input(input_buffer, *input_frames_count);
 
+  /* resampled_frame_count == 0 happens if the resampler
+   * doesn't have enough input frames buffered to produce 1 resampled frame. */
+  if (resampled_frame_count == 0) {
+      return *input_frames_count;
+  }
+
   size_t frames_resampled = 0;
   resampled_input = input_processor->output(resampled_frame_count, &frames_resampled);
   *input_frames_count = frames_resampled;


kinetiknz:
Looks good, thanks!  Apologies for the delay merging this.

PR #649, opened 2021-06-26
sndio: default max_channels to 8 for OUTPUT and 2 for INPUT

landryb:
cf https://bugzilla.mozilla.org/show_bug.cgi?id=1633769


padenot:
Thanks!

PR #650, opened 2021-07-08
Fix audio input when the audio device is a citrix device

padenot:
When connecting to a remote machine via Citrix, the devices on the remote machine exposes a _very_ small buffer size. This is too low for (e.g.) doing a WebRTC call with a robust audio input stream in Firefox. It's in fact not that much of a problem to increase the buffer size between `audiodg.exe` and cubeb, and makes the input a lot better. The buffer size itself doesn't affect latency that much, it's apparently something like a ring buffer that a bit bigger, so more robust on various under-runs, but latency is still kept quite low. I've tested this on the Citrix setup (where it improves the quality greatly), and on a couple local setups (where it seem to add a few milliseconds of latency maybe, but I'm not sure).

Also, `cubeb_wasapi.cpp` is currently grabbing the input data from the output event. This has been working well for years, but doesn't work too well on Citrix that has minuscule buffer sizes: if the output event is each 10ms, and the input event is around 133 frames at 48k (typical values for the Citrix setup I've tested this on), we end up missing a lot of input because it doesn't appear to be buffered for long enough. I think that as a followup up, to further increase robustness against load, for when the output's callback load is very high, we might want to spin a thread just for input. I have half the patch written for this.

Additionally, this adds logging to quickly check if a setup is under-running, or if the input if drifting vs. the output, and fixes some weirdness in `wasapi_get_min_latency` that would return the number to use with `IAudioClient3`, but this path has been disabled for now.

This also increases quality of the input with my Sony WH1000-XM3 (locally), it was suffering with more or less the same symptomes: input buffer size too small, device period too small.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -1746,11 +1746,12 @@ wasapi_get_min_latency(cubeb * ctx, cubeb_stream_params params, uint32_t * laten
      synchronizing the stream and the engine.
      http://msdn.microsoft.com/en-us/library/windows/desktop/dd370871%28v=vs.85%29.aspx */
 
-  #ifdef _WIN32_WINNT_WIN10
-    *latency_frames = hns_to_frames(params.rate, minimum_period);
-  #else
+   // #ifdef _WIN32_WINNT_WIN10
+  #if 0
+     *latency_frames = hns_to_frames(params.rate, minimum_period);
+   #else
     *latency_frames = hns_to_frames(params.rate, default_period);
-  #endif
+   #endif
 
   LOG("Minimum latency in frames: %u", *latency_frames);
 
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -2151,22 +2151,17 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
   if (rv == CUBEB_OK) {
     const char* HANDSFREE_TAG = "BTHHFENUM";
     size_t len = sizeof(HANDSFREE_TAG);
-    if (direction == eCapture &&
-        strlen(device_info.group_id) >= len &&
-        strncmp(device_info.group_id, HANDSFREE_TAG, len) == 0) {
-      // Rather high-latency to prevent constant under-runs in this particular
-      // case of an input device using bluetooth handsfree.
+    if (direction == eCapture) {
       uint32_t default_period_frames = hns_to_frames(device_info.default_rate, default_period);
-      latency_frames = default_period_frames * 4;
-      stm->input_bluetooth_handsfree = true;
-      LOG("Input is a bluetooth device in handsfree, latency increased to %u frames from a default of %u", latency_frames, default_period_frames);
-    } else {
-      uint32_t minimum_period_frames = hns_to_frames(device_info.default_rate, minimum_period);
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -1351,10 +1360,18 @@ wasapi_stream_render_loop(LPVOID stream)
               (!has_input(stm) && has_output(stm)));
       is_playing = stm->refill_callback(stm);
       break;
-    case WAIT_OBJECT_0 + 3: /* input available */
-      if (has_input(stm) && has_output(stm)) { continue; }
-      is_playing = stm->refill_callback(stm);
+    case WAIT_OBJECT_0 + 3: { /* input available */
+      HRESULT rv = get_input_buffer(stm);
```
kinetiknz, line 1364: I think we need a `linear_input_buffer->clear()` in `close_wasapi_stream()` to avoid a potential glitch on device change for duplex streams where we've called `get_input_buffer()` but the "output available" event hasn't triggered and cleared the buffer during regular processing.

src/cubeb_wasapi.cpp
```diff
@@ -353,6 +353,9 @@ struct cubeb_stream {
   /* This needs an active audio input stream to be known, and is updated in the
    * first audio input callback. */
   std::atomic<int64_t> input_latency_hns { LATENCY_NOT_AVAILABLE_YET };
+
+  size_t total_input_frames = 0;
```
kinetiknz, line 357: A comment noting these two are only used for logging would be good.  I don't think we need to worry about clearing them between device changes.

src/cubeb_wasapi.cpp
```diff
@@ -2138,22 +2156,17 @@ int setup_wasapi_stream_one_side(cubeb_stream * stm,
...
-      latency_frames = std::max(latency_frames, minimum_period_frames);
-      stm->input_bluetooth_handsfree = false;
-      LOG("Input is a not bluetooth handsfree, latency %s to %u frames (minimum %u)", latency_frames < minimum_period_frames ? "increased" : "set", latency_frames, minimum_period_frames);
+      if (strlen(device_info.group_id) >= len &&
+          strncmp(device_info.group_id, HANDSFREE_TAG, len) == 0) {
+        stm->input_bluetooth_handsfree = true;
+      } else {
+        stm->input_bluetooth_handsfree = false;
+      }
+      latency_frames = default_period_frames * 8;
```
kinetiknz, line 2167: I know the previous `* 4` didn't have a comment either, but a note about how this value was derived (empirically?) would be appreciated.

src/cubeb_wasapi.cpp
```diff
@@ -1351,10 +1360,18 @@ wasapi_stream_render_loop(LPVOID stream)
...
-      if (has_input(stm) && has_output(stm)) { continue; }
-      is_playing = stm->refill_callback(stm);
+    case WAIT_OBJECT_0 + 3: { /* input available */
+      HRESULT rv = get_input_buffer(stm);
+      if (FAILED(rv)) {
+        return rv;
+      }
+
+      if (!has_output(stm)) {
+        is_playing = stm->refill_callback(stm);
```
kinetiknz, line 1370: `refill_callback_input()` still calls `get_input_buffer()` internally; I think that needs the same change you made to `refill_callback_duplex()`.

kinetiknz (approved):
LGTM, thanks.

PR #651, opened 2021-07-24
Add a multiple input devices test

ChunMinChang:
Add a test to check if we can open multiple input devices at the same time. This is a simple check when I debug an issue for [BMO 1238038](https://bugzilla.mozilla.org/show_bug.cgi?id=1238038). Everything works well on the cubeb end but I think it's no harm to add a test here. The duplex stream is to simulate the [duplex stream Firefox uses in GraphDriver](https://searchfox.org/mozilla-central/rev/290f33a3245f61f34ec5f80f548e6584823ba9ad/dom/media/GraphDriver.cpp#663). The input-only stream is used to simulate the second microphone we will open.


kinetiknz (approved):
LGTM, thanks!

PR #653, opened 2021-07-27
Increase delay to really test the pulse backends

padenot:
Turns out it was testing the preroll, and with it gone from the
pulse-rust backend in https://github.com/mozilla/cubeb-pulse-rs/pull/69,
this fails on (at least) docker-on-AWS hosts on Mozilla's infra.

Also, lots of tests fail for me locally on the latest Ubuntu LTS, but I'm not really sure why, it seems to deadlock on the main loop lock when doing lots of start/stop operations in a short period of time.

5000ms seem to work on the test boxes though, they are really slow to start (1000ms doesn't work consistently for example).

### DIFF
--- a/test/test_sanity.cpp
+++ b/test/test_sanity.cpp
@@ -612,7 +612,7 @@ TEST(cubeb, drain)
   r = cubeb_stream_start(stream);
   ASSERT_EQ(r, CUBEB_OK);
 
-  delay(500);
+  delay(5000);
 
   do_drain = 1;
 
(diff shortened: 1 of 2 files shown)


kinetiknz:
Thanks, LGTM.

I'm curious if the test hangs still happen if you temporarily revert the preroll buffering change - maybe that exposed a latent timing issue in the code?

padenot:
> I'm curious if the test hangs still happen if you temporarily revert the preroll buffering change - maybe that exposed a latent timing issue in the code?

What I think is happening is that there's a deadlock somewhere on the mainloop lock that is triggered by our tests being sometimes extremely aggressive with starting/stopping the stream. With the preroll thing, this called the callback and completed the test sometimes during the start call, so this wasn't happening. I bug reports upstream with the same setup and back traces I think. I'll keep an eye open in any case -- I spent a lot of time trying to understand this.

PR #654, opened 2021-07-27
Align code styles with `.clang-format` settings

ChunMinChang:
Format all the code under `include` and `src` except those files under
`src/speex` with style setting in `.clang-format` file by the following
script:

```sh

FILE_LIST="$(find "include" "src" -not -path "src/speex/*" | grep -E ".*(\.cpp|\.c|\.h|\.hpp|\.hh)$")"

echo "Files found to format:\n---\n$FILE_LIST\n---"

clang-format --verbose -i $FILE_LIST
```

### DIFF
--- a/include/cubeb/cubeb.h
+++ b/include/cubeb/cubeb.h
@@ -7,9 +7,9 @@
 #if !defined(CUBEB_c2f983e9_c96f_e71c_72c3_bbf62992a382)
 #define CUBEB_c2f983e9_c96f_e71c_72c3_bbf62992a382
 
+#include "cubeb_export.h"
 #include <stdint.h>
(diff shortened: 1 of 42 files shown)


kinetiknz:
Thanks for doing this!

ChunMinChang:
It would be great to add a script in this repo to format the code automatically. We can run it before merging the PRs so we don't need to worry about the styles.

kinetiknz:
> It would be great to add a script in this repo to format the code automatically. We can run it before merging the PRs so we don't need to worry about the styles.

Agreed.  I've made a start in https://github.com/mozilla/cubeb/pull/655, we can always improve it later.

PR #655, opened 2021-07-27
Add clang-format-check target to CMake and run on CI.

kinetiknz:
Adds a new `check_format` job using a new `clang-format-check`  (non-default) CMake target.  This job only runs on Linux, but it checks all files.


ChunMinChang (approved):
Fantastic, thanks!

PR #656, opened 2021-07-31
winmm: fix broken audio when sample count is overflowed

roytam1:
Microsoft wave audio docs say "samples are the preferred time format in which
to represent the current position", but relying on this causes problems on
Windows XP, the only OS cubeb_winmm is used on.

While the wdmaud.sys driver internally tracks a 64-bit position and ensures no
backward movement, the WinMM API limits the position returned from
waveOutGetPosition() to a 32-bit DWORD (this applies equally to XP x64). The
higher 32 bits are chopped off, and to an API consumer the position can appear
to move backward.

In theory, even a 32-bit TIME_SAMPLES position should provide plenty of
playback time for typical use cases before this pseudo wrap-around, e.g:
    (2^32 - 1)/48000 = ~24:51:18 for 48.0 kHz stereo;
    (2^32 - 1)/44100 = ~27:03:12 for 44.1 kHz stereo.
In reality, wdmaud.sys doesn't provide a TIME_SAMPLES position at all, only a
32-bit TIME_BYTES position, from which wdmaud.drv derives TIME_SAMPLES:
    SamplePos = (BytePos * 8) / BitsPerFrame,
    where BitsPerFrame = Channels * BitsPerSample,
Per dom\media\AudioSampleFormat.h, desktop builds always use 32-bit FLOAT32
samples, so the maximum for TIME_SAMPLES should be:
    (2^29 - 1)/48000 = ~03:06:25;
    (2^29 - 1)/44100 = ~03:22:54.
This might still be OK for typical browser usage, but there's also a bug in the
formula above: BytePos * 8 (BytePos << 3) is done on a 32-bit BytePos, without
first casting it to 64 bits, so the highest 3 bits, if set, would get shifted
out, and the maximum possible TIME_SAMPLES drops unacceptably low:
    (2^26 - 1)/48000 = ~00:23:18;
    (2^26 - 1)/44100 = ~00:25:22.

To work around these limitations, we just get the position in TIME_BYTES,
recover the 64-bit value, and do our own conversion to samples.

Original patch by mixit @ msfn.org, and permission granted for submitting PR: https://msfn.org/board/topic/182647-my-browser-builds-part-3/?do=findComment&comment=1202896

### DIFF
--- a/src/cubeb_winmm.c
+++ b/src/cubeb_winmm.c
@@ -121,6 +121,10 @@ struct cubeb_stream {
   CRITICAL_SECTION lock;
   uint64_t written;
   float soft_volume;
+  /* For position wrap-around handling: */
+  size_t frame_size;
+  DWORD prev_pos_lo_dword;
+  DWORD pos_hi_dword;
 };
 
 static size_t
@@ -557,6 +561,10 @@ winmm_stream_init(cubeb * context, cubeb_stream ** stream,
     winmm_refill_stream(stm);
   }
 
+  stm->frame_size = bytes_per_frame(stm->params);
+  stm->prev_pos_lo_dword = 0;
+  stm->pos_hi_dword = 0;
+
   *stream = stm;
 
   return CUBEB_OK;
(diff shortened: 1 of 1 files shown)


kinetiknz:
Many thanks to you and mixit for thoroughly investigating this and reporting it with a fix.  It's a shame this bug has persisted so long - I suspect this is partially due to most of the internal testing being done on Windows 7, which may not have this bug in wdmaud.sys.

PR #657, opened 2021-09-19
build(cmake): use system GTest when available

Tachi107:
This speeds up compilation times and makes downstream packaging easier.

It could be opt-in, with an option like `USE_SYSTEM_GTEST`.

It seems that you internally use GoogleTest 1.6, but everything seems to also work with version 1.10 (Debian)


kinetiknz:
Thanks for the PRs, and sorry for the delay reviewing them.

> This speeds up compilation times and makes downstream packaging easier.
> 
> It could be opt-in, with an option like `USE_SYSTEM_GTEST`.

I'll merge this as-is for now.  We can always add `USE_SYSTEM_GTEST` later if it seems necessary.

> It seems that you internally use GoogleTest 1.6, but everything seems to also work with version 1.10 (Debian)

We're not using 1.6 for any specific reason.  It'd be good to update the vendored version at some stage.

PR #659, opened 2021-09-22
build(cmake): use system speex when available

### LINKED ISSUE #658: Use system speex when available

Thanks

Tachi107:
Since you use unpatched speex using the already installed libraries doesn't cause issues (all tests pass).

I converted the speex `OBJECT` library into a regular `STATIC` lib; it is easier to reason about it and doesn't require to add definitions like `OUTSIDE_SPEEX` in all targets that use the library, and also simplifies things when using the library found by pkg-config (you don't need to have special if statements everywhere, you can just alias the imported target to `speex` and everything just works).

Lastly, this makes downstream life's significantly easier.

Fixes #658

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,22 +66,34 @@ else()
   string(REPLACE "/EHsc" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS}) # Disable Exceptions
 endif()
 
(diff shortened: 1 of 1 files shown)


Tachi107:
This seems to not work when libspeex is not installed on the system. Converting to draft for now

kinetiknz:
Superseded by PR #662.

PR #660, opened 2021-09-24
build(cmake): add version and soversion to the built library

Tachi107:
This makes the shared library usable when installed system-wide.

`cubeb_VERSION` and `cubeb_VERSION_MAJOR` are set when calling `project(cubeb VERSION x.y.z)`


kinetiknz:
Thanks!

PR #662, opened 2021-10-03
build(cmake): add option to disable lazy load of libs and modernize the CMake file

### LINKED ISSUE #658: Use system speex when available

Tachi107:
Add `LAZY_LOAD_LIBS` option to disable lazy loading of shared libraries

Add `USE_SANITIZERS` to disable use of sanitizers when not needed

Add SONAME to the shared library

Use system speex when `LAZY_LOAD_LIBS` is off

Use pkg-config to look for libs when `LAZY_LOAD_LIBS` is off (requires CMake 3.7)

Use `doxygen_add_docs` instead of custom target for API docs generation (requires CMake 3.9)

Use system GTest when available

Use `gtest_discover_tests` instead of `add_tests` for better CTest integration (requires CMake 3.10)

Move embedded speex source to a `subprojects`, to better show that it is a third party library and to avoid conflicting headers in `src/cubeb-speex-resampler.h` when using system headers

Closes #657, fixes #658, closes #659, closes #660


kinetiknz:
This looks good, thank you.  I merged #657 and #660 already, so this will need to be rebased please.

Just confirming that you consider #659 obsoleted by this PR, too?

Tachi107:
> Just confirming that you consider #659 obsoleted by this PR, too?

Yep, I'll rebase in a few hours. Thanks for merging the other PRs :D

Tachi107:
I'm splitting the changes in different PRs, closing this

PR #665, opened 2021-11-01
jack: possibility to disable lazy load of lib

Tachi107:
From #662

### DIFF
--- a/src/cubeb_alsa.c
+++ b/src/cubeb_alsa.c
@@ -660,11 +660,19 @@ init_local_config_with_workaround(char const * pcm_name)
 
   lconf = NULL;
 
-  if (*WRAP(snd_config) == NULL) {
+  snd_config_t * gconf;
+
+#ifndef DISABLE_LIBASOUND_DLOPEN
+  gconf = *WRAP(snd_config);
+#else
+  gconf = WRAP(snd_config);
+#endif
+
+  if (gconf == NULL) {
     return NULL;
   }
 
-  r = WRAP(snd_config_copy)(&lconf, *WRAP(snd_config));
+  r = WRAP(snd_config_copy)(&lconf, gconf);
   if (r < 0) {
     return NULL;
   }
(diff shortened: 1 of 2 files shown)


Tachi107:
CI is failing because apt can't find some packages, maybe it is necessary to run `apt update` before `apt install`

kinetiknz:
> CI is failing because apt can't find some packages, maybe it is necessary to run `apt update` before `apt install`

Thanks, that should be fixed now.

src/cubeb_alsa.c
```diff
@@ -668,6 +669,16 @@ init_local_config_with_workaround(char const * pcm_name)
...
+#else
+  if (WRAP(snd_config) == NULL) {
+    return NULL;
+  }
+
+  r = WRAP(snd_config_copy)(&lconf, WRAP(snd_config));
+  if (r < 0) {
+    return NULL;
+  }
+#endif
```
kinetiknz, line 681: Rather than duplicate this small piece of code, perhaps we can do this:
```
snd_config_t * gconf;

#ifndef DISABLE_LIBASOUND_DLOPEN
gconf = *WRAP(snd_config);
#else
gconf = WRAP(snd_config);
#endif

if (g_config == NULL) {
...
```
```

kinetiknz (approved):
One minor nitpick, otherwise looks good.  Thanks!

src/cubeb_alsa.c
Tachi107: Done :)

I don't really know if it is necessary though, I did it because I the compiler warned me about `*WRAP(snd_config)`

kinetiknz: Thanks!

It is necessary, because the wrapped type is declared as `snd_config_t ** cubeb_snd_config` where the original is `snd_config_t * snd_config`.  For the function pointers, this doesn't matter because `foo(...)` and `(*foo)(...)` are equivalent.  But I should've suggested fixing the wrapper instead - `#define WRAP(x) (*cubeb_##x)` would avoid the extra deref for `snd_config` and make this all a bit tidier.

PR #667, opened 2021-11-01
build: add LAZY_LOAD_LIBS option and always link with threads

Tachi107:
Users can now disable the lazy loading of libraries, and use classic linkage instead

From #662

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # TODO
 # - backend selection via command line, rather than simply detecting headers.
 
-cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.7 FATAL_ERROR)
 project(cubeb
   VERSION 0.0.0)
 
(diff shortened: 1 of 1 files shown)


kinetiknz (approved):
Thanks for doing this, looks good!

It'd be nice to avoid duplicating the target definitions for these backends, if possible.

Tachi107:
Ok, I'll clean this up later today

Tachi107:
Not that much cleaner, but at least `target_sources` and some target defines are now not duplicated. 

I've also changed the way `-pthread` is handled; it was previously added only if some backends were active, even if `cubeb_log` requires it with some compilers, as it uses `std::thread`. I wanted to submit the change as another PR (from commit https://github.com/Tachi107/cubeb/commit/b724fc1170c4210fb86c725a0e9ec4717f3dc6eb), but squeezing it here allowed me to clean this up a bit more.

kinetiknz (approved):
Thanks again!

PR #669, opened 2021-11-06
build: add USE_SANITIZERS option

Tachi107:
This is useful to users who can't recursively clone the repo or that simply want to build the project from the downloadable tarball (like Linux distributions, for example).

From #662


kinetiknz:
Thanks!

PR #670, opened 2021-11-06
build: use system speex when possible

### LINKED ISSUE #658: Use system speex when available

Tachi107:
Following #658 (comment), the speex library is now handled like a normal dependency: cubeb will link against the system version if available, and fall back to the bundled one if not.

I've also added a `BUNDLE_SPEEX` option, so that you can force the use of the bundled library if needed (e.g. creating a standalone libcubeb on a system where libspeex is available).

I also had to move the bundled library to a separate folder. As `src` is always added as an include path, the headers in `src/speex` would conflict with system headers. And it also clears the relationship between cubeb and speex. I choose the "subprojects" name to follow the Meson convention, since CMake does not have one. A bit OT, but if you're curious you can see their rationale [here](https://mesonbuild.com/Subprojects.html#why-must-all-subprojects-be-inside-a-single-directory)

Lastly, I added `cubeb_log.cpp` to the list of sources of `test_resampler`, as I was getting linking errors when building with `BUILD_SHARED_LIBS=true`

Fixes #658


Tachi107:
Strange, I'm not getting that EXPORT issue on my machine... Working on it.

Tachi107:
Fixed :D

kinetiknz:
Thanks, this is much cleaner.  Appreciate all the build system cleanups!

PR #672, opened 2021-11-09
build: minor CMake cleanups

Tachi107:
These are minor cleanups that do not affect the behaviour of the script.

In particular, I removed most `DESTINATION` arguments from `install()`, since they are automatically deduced based on the type of file to be installed.

In #662 I also added `doxygen_add_docs` and `gtest_discover_tests`, but they require CMake version 3.9 and 3.10 respectively, and I think that the advantages they provide are not worth dropping support for older versions


kinetiknz:
Thanks!

PR #673, opened 2021-11-09
cubeb_oss: changes to buffering logic

khng300:
The change is to reduce the number of dropped frames when playing
YouTube videos.  The most significant part is to set trigger threshold
to be low enough, such that as long as the audio hardware's DMA engine
moves data the callback is triggered without waiting longer.  Besides
that, unify oss_wait_*fd_for_space() for both directions into a single
function.  Buffer size is also not required to match the other
direction.

Sponsored by:	The FreeBSD Foundation


kinetiknz:
Thanks!

PR #675, opened 2021-11-10
OSS: Fix warnings in the initialization path

khng300:
There are some warnings that I did not catch (for no reason, or some reasons I did not recall).


kinetiknz (approved):
Thanks!

PR #678, opened 2021-11-27
cubeb_oss: Fix for buffer overrun.

khng300:
This patch is proposed by @hselasky to fix an issue that causes corruption when upstream buffer size is larger than our specified buffer size.

The corresponding ticket is https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=260061.


kinetiknz (approved):
LGTM, thanks.

PR #680, opened 2021-12-06
cubeb_oss: Fix recording buffer size computation.

khng300:
This is a patch from @hselasky . To calculate maximum buffer size, use bi.fragstotal instead of bi.fragments since bi.fragments is usually zero for recording direction.


kinetiknz (approved):
Thanks!

PR #681, opened 2021-12-07
Fix build on MinGW

padenot:
Original downstream patch by Tom Ritter <tom@mozilla.com>.

See his investigation here: https://bugzilla.mozilla.org/show_bug.cgi?id=1740606#c0.


kinetiknz (approved):
Thanks!

PR #682, opened 2022-01-05
wasapi: Speed up enumerate_devices by caching default device IDs for each flow/role instead of requerying for every wasapi_create_device.

kinetiknz:
Locally (on a i9-9900KS with ~60 input/~60 output devices), this reduces a single enumerate_devices call from ~1100ms to ~250ms.


padenot (approved):
Nice, thanks!

PR #684, opened 2022-01-10
build: use libbsd when using OSS and glibc

Tachi107:
Some systems offer OSS v4 but do not have BSD-only functions like `strlcpy`, used in cubeb_oss.c. An example is Debian GNU/kFreeBSD, that uses a FreeBSD kernel with GNU userspace utilities, like glibc.

With this patch cubeb will be able to build on these systems thanks to the help of libbsd, that privides most libc functions that you would expect on BSDs.

~Please note that I did not test this patch yet, but I'll do it as soon as possible.~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -284,9 +284,22 @@ if(HAVE_SYS_SOUNDCARD_H)
   try_compile(USE_OSS "${PROJECT_BINARY_DIR}/compile_tests"
     ${PROJECT_SOURCE_DIR}/cmake/compile_tests/oss_is_v4.c)
   if(USE_OSS)
-    target_sources(cubeb PRIVATE
-      src/cubeb_oss.c)
-    target_compile_definitions(cubeb PRIVATE USE_OSS)
+    # strlcpy is not available on BSD systems that use glibc,
+    # like Debian kfreebsd, so try using libbsd if available
+    include(CheckSymbolExists)
+    check_symbol_exists(strlcpy string.h HAVE_STRLCPY)
+    if(NOT HAVE_STRLCPY)
+      pkg_check_modules(libbsd-overlay IMPORTED_TARGET libbsd-overlay)
+      if(libbsd-overlay_FOUND)
+        target_link_libraries(cubeb PRIVATE PkgConfig::libbsd-overlay)
+        set(HAVE_STRLCPY true)
+      endif()
+    endif()
+    if (HAVE_STRLCPY)
+      target_sources(cubeb PRIVATE
+        src/cubeb_oss.c)
+      target_compile_definitions(cubeb PRIVATE USE_OSS)
+    endif()
   endif()
 endif()
 


CMakeLists.txt
```diff
@@ -284,9 +284,22 @@ if(HAVE_SYS_SOUNDCARD_H)
...
+    include(CheckSymbolExists)
+    check_symbol_exists(strlcpy string.h HAVE_STRLCPY)
+    if(NOT HAVE_STRLCPY)
+      pkg_check_modules(libbsd-overlay IMPORTED_TARGET libbsd-overlay)
+      if(libbsd-overlay_FOUND)
+        target_link_libraries(cubeb PRIVATE PkgConfig::libbsd-overlay)
+        set(HAVE_STRLCPY true)
+      endif()
+    endif()
+    if (HAVE_STRLCPY)
```
jbeich, line 298: I don't like silently disabling cubeb_oss.c. Instead put `message(FATAL_ERROR "strlcpy not found, required by OSS backend")`, so users can install `libbsd` or debug why `strlcpy` failed to detect.

CMakeLists.txt
```diff
@@ -284,9 +284,22 @@ if(HAVE_SYS_SOUNDCARD_H)
   try_compile(USE_OSS "${PROJECT_BINARY_DIR}/compile_tests"
     ${PROJECT_SOURCE_DIR}/cmake/compile_tests/oss_is_v4.c)
   if(USE_OSS)
-    target_sources(cubeb PRIVATE
-      src/cubeb_oss.c)
-    target_compile_definitions(cubeb PRIVATE USE_OSS)
+    # strlcpy is not available on BSD systems that use glibc,
```
jbeich, line 287: Better reword to remove "BSD systems that use" comment:
- NetBSD and OpenBSD don't use OSS (libossaudio is a compat shim over SunAudio)
- Only Debian/kFreeBSD uses GNU libc (discontinued Gentoo/FreeBSD used BSD libc)
- Linux can use OSSv4 from opensound.com (not as good as FreeBSD)
- `strlcpy` exists in musl (Alpine) and bionic (Android)
- macOS and Android can be considered BSD due to libc but have non-BSD kernels

CMakeLists.txt
jbeich: Nevermind, if an OS update on FreeBSD breaks `check_symbol_exists` it'd probably be noticed very quickly. ;)

Tachi107:
I can confirm that building cubeb on Debian GNU/kFreeBSD with this patch applied fixes the build failures

kinetiknz:
Holding off on merging this until @jbeich's comments are addressed.  Note I saw them come through via email, but they aren't showing up on the PR for me.

Tachi107:
I don't see them either :/

kinetiknz:
@jbeich ping, would you mind commenting on this please?

jbeich:
Works fine on FreeBSD. Like before cubeb_oss.c is enabled by default but only built when relevant features are auto-detected.

I've removed my review to avoid overfocusing on a code comment when the code itself was OK.

Tachi107:
@jbeich it's always nice to hear criticism, so go ahead :)
I'm not an expert in BSDs, so my comment could be wrong in some ways.

PR #685, opened 2022-01-17
cubeb_oss: Force at maximum 32ms service intervals.

khng300:
Also when the latency is specified to be very big.
This might be a bug in some upper layer, although I see
other cubeb audio drivers doing the same.

This "fixes" audio sync with youtube videos.

### DIFF
--- a/src/cubeb_oss.c
+++ b/src/cubeb_oss.c
@@ -97,6 +97,7 @@ struct oss_stream {
   int fd;
   void * buf;
   unsigned int bufframes;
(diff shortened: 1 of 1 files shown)


hselasky:
This patch has been used successfully with Firefox in FreeBSD for some time:

https://www.freshports.org/www/firefox/

hselasky:
@kinetiknz : Ping.

kinetiknz (approved):
Thanks!

PR #686, opened 2022-01-19
cubeb_oss: Force at maximum 32ms service intervals.

hselasky:
Also when the latency is specified to be very big.
This might be a bug in some upper layer, although I see
other cubeb audio drivers doing the same.

This "fixes" audio sync with youtube videos.

### DIFF
--- a/src/cubeb_oss.c
+++ b/src/cubeb_oss.c
@@ -97,6 +97,7 @@ struct oss_stream {
   int fd;
   void * buf;
   unsigned int bufframes;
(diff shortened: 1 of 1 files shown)


hselasky:
@kinetiknz @khng300 FYI

This patch has been used with the FreeBSD Firefox port for a while without any issues.

https://www.freshports.org/www/firefox/

--HPS

hselasky:
Oops - double patch submit.

PR #687, opened 2022-02-24
 Assert device collection callbacks were unregistered before cubeb_destroy is called.

kinetiknz:
This primarily affects wasapi, but I've added asserts to the audiounit and pulse backends for the new test to verify.  I'll update the Rust versions later.

Also includes 3e3c9e1783af11656ce82ee058cffe7d3ff1d292, small fix for failing test_loopback on Windows.


test/test_duplex.cpp
```diff
@@ -173,13 +167,49 @@ TEST(cubeb, duplex_collection_change)
...
+
+TEST(cubeb, duplex_collection_change_no_unregister)
+{
+  cubeb * ctx;
+  int r;
+
+  r = common_init(&ctx, "Cubeb duplex example with collection change");
+  ASSERT_EQ(r, CUBEB_OK) << "Error initializing cubeb library";
+  std::unique_ptr<cubeb, decltype(&cubeb_destroy)> cleanup_cubeb_at_exit(
+      ctx, test_cubeb_destroy_failure);
```
ChunMinChang, line 208: nit: I would prefer inlining `test_cubeb_destroy_failure` in this test

ChunMinChang (approved):
LGTM.

Should we rename this PR to "Assert if device collection callbacks are unregistered before cubeb_destroy is called" instead? 

Is this related to https://github.com/mozilla/audioipc/pull/147?

test/test_duplex.cpp
kinetiknz: Done.

kinetiknz:
> Is this another issue unrelated to the topic of this PR?

It's a small ride-along fix for a test failure I noticed on Windows.

> Should we rename this PR to "Assert if device collection callbacks are unregistered before cubeb_destroy is called" instead?

I've reworded the commit message/PR title.

> Is this related to [mozilla/audioipc#147](https://github.com/mozilla/audioipc/pull/147)?

Yeah, these asserts would've caught a bug fixed in that PR.

PR #691, opened 2022-03-05
pulse: Lazy prototype loopback implementation

nikp123:
**DO NOT MERGE**

# Context
I've just recently found about this project and I really like the idea of it and thought I'd use it as my main audio library if it manages to do everything I'd want it to be able to do. Unfortunately, loopback mode seems to be WASAPI only despite it being technically possible to pull off (without even user intervention) on both PulseAudio and PipeWire, however these modes are not commonly documented so you may take them with a grain of salt.

# What this pull request is?
This is a really hacky implementation of said loopback mode on PulseAudio (ie. monitor mode). Consider it a RFC before I either polish it or you merge it. It is menat to demonstrate possible functionality before anything else.

# How does the magic work? ie. how the code works
Basically it's a bit hacky, but the trick is to create a pulseaudio mainloop (ie create a sink) so that we are able to fetch the default sink's name and turn into into a monitor source by appending ```.monitor``` to it's name. Any ideas on simplyfing this code would be appreciated.

And of course your opinions/thoughts on this. I, for one, would like to see this functionality enabled in the library (if possible). Thanks.

### DIFF
--- a/src/cubeb_pulse.c
+++ b/src/cubeb_pulse.c
@@ -40,6 +40,12 @@
   X(pa_cvolume_set)                                                            \
   X(pa_cvolume_set_balance)                                                    \
   X(pa_frame_size)                                                             \
+  X(pa_mainloop_free)                                                          \
+  X(pa_mainloop_get_api)                                                       \
+  X(pa_mainloop_iterate)                                                       \
+  X(pa_mainloop_new)                                                           \
+  X(pa_mainloop_run)                                                           \
+  X(pa_mainloop_quit)                                                          \
   X(pa_operation_get_state)                                                    \
   X(pa_operation_unref)                                                        \
   X(pa_proplist_gets)                                                          \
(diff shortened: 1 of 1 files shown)


padenot:
Yes, this would work well, although it's possible to reuse lots of code we have. You can for example enumerate all devices using either the internal or external API, and pick a monitor device as your input device. The "loopback" feature is WASAPI-only historically because it's somewhat possible to achieve this externally (monitor device are real devices, unlike the WASAPI case where you pass an output device as input device with the flag), but I reckon an alignment would be nice. It's not possible to generally implement this on macOS (outside of Blackhole/Dante Virtual Soundcard/etc.).

The current and up to date PulseAudio backend for cubeb is here: https://github.com/mozilla/cubeb-pulse-rs (it has more than a few bug fixes) and is written in Rust. It requires a rust toolchain for compilation, but it's really straightforward to set it up.

nikp123:
@padenot i'll try to get the rust implementation going if that's what's currently used, but yeah i share the same feelings about the current code, this is just hacked together to show how it could be done.

padenot:
Dropping `cubeb-pulse-rs` into `cubeb/src/` and configuring the cmake project with `BUILD_RUST_LIBS` is all you need to do.

nikp123:
Closing this as the priority is on the rust one anyway.

PR #693, opened 2022-03-22
Add device selection to cubeb-test

ChunMinChang:
It's a lot easier to develop/debug device-related features/issues if we
can select input or/and output device for the audio stream in cubeb-test


kinetiknz:
LGTM, thanks!

PR #695, opened 2022-03-23
Avoid unnecessary API calls when initializing a stream

ChunMinChang:
This PR avoids calling API when we don't really need it, and avoid overwriting `input_bluetooth_handsfree` during setting output side.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -2238,6 +2238,7 @@ setup_wasapi_stream_one_side(cubeb_stream * stm,
                              stm->device_enumerator.get(), device.get(),
                              &default_devices) == CUBEB_OK) {
       // This multiplicator has been found empirically.
+      XASSERT(device_info.latency_hi > 0);
       uint32_t latency_frames = device_info.latency_hi * 8;
       LOG("Input: latency increased to %u frames from a default of %u",
           latency_frames, device_info.latency_hi);
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -2225,15 +2225,6 @@ setup_wasapi_stream_one_side(cubeb_stream * stm,
     flags |= AUDCLNT_STREAMFLAGS_EVENTCALLBACK;
   }
 
-  // Sanity check the latency, it may be that the device doesn't support it.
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -2225,42 +2225,47 @@ setup_wasapi_stream_one_side(cubeb_stream * stm,
...
+
+    wasapi_default_devices default_devices(stm->device_enumerator.get());
+    cubeb_device_info device_info;
+    if (wasapi_create_device(stm->context, device_info,
+                             stm->device_enumerator.get(), device.get(),
+                             &default_devices) == CUBEB_OK) {
+      // Sanity check the latency, it may be that the device doesn't support it.
+      REFERENCE_TIME minimum_period;
+      REFERENCE_TIME default_period;
+      hr = audio_client->GetDevicePeriod(&default_period, &minimum_period);
```
kinetiknz, line 2243: `wasapi_create_device` already calls `GetDevicePeriod`, so we can just check if `device_info.latency_hi` is non-zero here and use that for the `default_period` value.

kinetiknz (approved):
Nice clean up, thank you.

src/cubeb_wasapi.cpp
ChunMinChang: Nice caught! I've updated the patch

src/cubeb_wasapi.cpp
```diff
@@ -2225,42 +2225,36 @@ setup_wasapi_stream_one_side(cubeb_stream * stm,
...
+  if (direction == eCapture) {
+    stm->input_bluetooth_handsfree = false;
+
+    wasapi_default_devices default_devices(stm->device_enumerator.get());
+    cubeb_device_info device_info;
+    if (wasapi_create_device(stm->context, device_info,
+                             stm->device_enumerator.get(), device.get(),
+                             &default_devices) == CUBEB_OK) {
+      // This multiplicator has been found empirically.
+      uint32_t latency_frames = device_info.latency_hi * 8;
```
kinetiknz, line 2241: nit: Perhaps add an assert for `device_info.latency_hi > 0`.  I don't think `GetDevicePeriod` will fail for an active device, but I'm not 100% sure.

PR #696, opened 2022-03-23
Don't reset device if CUBEB_STREAM_PREF_DISABLE_DEVICE_SWITCHING is set

ChunMinChang:
We should not always reset the device when the current device is unplugged. The device reset or not should follow the `CUBEB_STREAM_PREF_DISABLE_DEVICE_SWITCHING` pref.


kinetiknz (approved):
LGTM, thanks

Issue #694, opened 2022-03-22
Disable automatic device switching if user doesn't follow the system default device


ChunMinChang:
To unblock [BMO 1735201](https://bugzilla.mozilla.org/show_bug.cgi?id=1735201) and [BMO 1238038](https://bugzilla.mozilla.org/show_bug.cgi?id=1238038), we should disable automatic device switching by default unless users ask to follow the system default device. That should be the default behavior in Firefox.

ChunMinChang:
To clarify a bit more, we should only switch devices automatically when user asks to use the system default device. Using the system default means that the users pass a `NULL` `cubeb_devid` in `cubeb_stream_init`. 
https://github.com/mozilla/cubeb/blob/5a2a20c6055d26e0a30fae8f1cf6f6cb975c5c97/include/cubeb/cubeb.h#L524
https://github.com/mozilla/cubeb/blob/5a2a20c6055d26e0a30fae8f1cf6f6cb975c5c97/include/cubeb/cubeb.h#L531

If users pass a non-`NULL` `cubeb_devid`, then it's no the default device, even if the device happens to be the current system default device. User must pass `NULL` explicitly to ask following the system default.

In brief, switching device or not can be summarized into the following situations:

For _input_ stream:
| device | unplug input device | unplug output device | default input change | default output change |
| ------- | -------------------- | -------------------- | --------------------- | ---------------------- |
| default | Yes | N/A | Yes | N/A |
| non-default | No | N/A | No | N / A |

For _output_ stream:
| device | unplug input device | unplug output device | default input change | default output change |
| ------- | -------------------- | -------------------- | --------------------- | ---------------------- |
| default | N/A | Yes | N/A | Yes |
| non-default | N/A | No |  N/A | No |

For _duplex_ stream:
| device | unplug input device | unplug output device | default input change | default output change |
| ------- | -------------------- | -------------------- | --------------------- | ---------------------- |
| default input, default output |  Yes | Yes | Yes | Yes |
| default input, non-default output |  Yes | No | Yes | No |
| non-default input, default output | No | Yes | No | Yes |
| non-default input, non-default output | No | No | No | No |

ChunMinChang:
I think it's a good time to revisit #167 again. I believe the WebRTC spec allow users to select the output device now: https://webrtc.github.io/samples/src/content/devices/input-output/, and Chrome has implemented that but we haven't. We should check the WebRTC spec and think how we should handle the device-switching policy in cubeb.

[...]

PR #697, opened 2022-03-23
Switch device only when the users don't specifiy a particular device

ChunMinChang:
We should not switch device behind the scenes if the user asks to use a particular device. If the user-selected device is unavailable, we should fire an error callback to notify the user so they can take the actions by their needs.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -2704,12 +2704,15 @@ wasapi_stream_init(cubeb * context, cubeb_stream ** stream,
     return rv;
   }
 
-  if (!((input_stream_params ? (input_stream_params->prefs &
-                                CUBEB_STREAM_PREF_DISABLE_DEVICE_SWITCHING)
-                             : 0) ||
-        (output_stream_params ? (output_stream_params->prefs &
-                                 CUBEB_STREAM_PREF_DISABLE_DEVICE_SWITCHING)
-                              : 0))) {
+  // Follow the system default devices when specifying default devices
+  // explicitly and CUBEB_STREAM_PREF_DISABLE_DEVICE_SWITCHING is not set.
+  if ((!input_device && input_stream_params &&
+       !(input_stream_params->prefs &
+         CUBEB_STREAM_PREF_DISABLE_DEVICE_SWITCHING)) ||
+      (!output_device && output_stream_params &&
+       !(output_stream_params->prefs &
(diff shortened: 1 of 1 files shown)


ChunMinChang:
This is based on #696. I'll move it from draft to official PR once #696 is merged.

ChunMinChang:
This implements the #694 for Windows part

src/cubeb_wasapi.cpp
```diff
@@ -2398,6 +2398,9 @@ setup_wasapi_stream(cubeb_stream * stm)
   XASSERT((!stm->output_client || !stm->input_client) &&
           "WASAPI stream already setup, close it first.");
 
+  std::unique_ptr<const wchar_t[]> picked_output_device_id(
```
kinetiknz, line 2412: nit: `selected_output_device_id` might be a slightly better name.

src/cubeb_wasapi.cpp
```diff
@@ -2110,6 +2112,10 @@ setup_wasapi_stream_one_side(cubeb_stream * stm,
   }
 
   stm->stream_reset_lock.assert_current_thread_owns();
+  // If user doesn't specify a particular device, we can choose another one when
+  // the given devid is unavailable.
+  bool fallbackable =
```
kinetiknz, line 2122: nit: Maybe call this `allow_fallback` .

src/cubeb_wasapi.cpp
```diff
@@ -2704,12 +2704,15 @@ wasapi_stream_init(cubeb * context, cubeb_stream ** stream,
     return rv;
   }
 
-  if (!((input_stream_params ? (input_stream_params->prefs &
-                                CUBEB_STREAM_PREF_DISABLE_DEVICE_SWITCHING)
-                             : 0) ||
-        (output_stream_params ? (output_stream_params->prefs &
-                                 CUBEB_STREAM_PREF_DISABLE_DEVICE_SWITCHING)
-                              : 0))) {
+  // Follow the system default devices when specifying default devices
```
kinetiknz, line 2712: "when not specifying devices explicitly"?

kinetiknz (approved):
LGTM, thanks!

src/cubeb_wasapi.cpp
ChunMinChang: That's a better description. I've updated it! Thanks!

PR #702, opened 2022-04-03
wasapi: Reduce device enumeration cost by excluding unnecessary devices

kinetiknz:
- bb2721e996e1ead9c3f7c6da100b54d50422e617: Excludes non-user-visible devices (DEVICE_STATE_NOTPRESENT) from the device list
- 5a5d6fc2026b276b1d6b8922a5019df50c10131f: Query only active devices when matching BT output device to BT input device

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -240,8 +240,9 @@ wasapi_create_device(cubeb * ctx, cubeb_device_info & ret,
 void
 wasapi_destroy_device(cubeb_device_info * device_info);
 static int
-wasapi_enumerate_devices(cubeb * context, cubeb_device_type type,
-                         cubeb_device_collection * out);
+wasapi_enumerate_devices_internal(cubeb * context, cubeb_device_type type,
+                                  cubeb_device_collection * out,
+                                  DWORD state_mask);
 static int
 wasapi_device_collection_destroy(cubeb * ctx,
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -2394,10 +2395,10 @@ wasapi_find_bt_handsfree_output_device(cubeb_stream * stm)
     return nullptr;
   }
 
-  int rv = wasapi_enumerate_devices(
+  int rv = wasapi_enumerate_devices_internal(
       stm->context,
       (cubeb_device_type)(CUBEB_DEVICE_TYPE_INPUT | CUBEB_DEVICE_TYPE_OUTPUT),
-      &collection);
+      &collection, DEVICE_STATE_ACTIVE);
```
padenot, line 2401: I assume you've tested that `DEVICE_STATE_ACTIVE` includes, says, the handsfree output "device" of a bluetooth headset that's currently in A2DP mode?

padenot (approved):

padenot:
Nice optimization, thanks!

src/cubeb_wasapi.cpp
kinetiknz: Yep, in a normal configuration.  Windows allows the user to disconnect or disable each of the headset and headphone devices separately, in which case they'd now be filtered out (which is desired, since we wouldn't be able to open them).

PR #704, opened 2022-04-13
wasapi: Fail stream setup instead of asserting if device latency unavailable.

kinetiknz:
Follow up to https://github.com/mozilla/cubeb/pull/695#discussion_r832770698.  Before that PR, we would fail setup when GetDevicePeriod failed.  This restores that behaviour by verifying `latency_hi` is valid.

This fixes [BMO 1764394](https://bugzilla.mozilla.org/show_bug.cgi?id=1764394).

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -2291,8 +2291,12 @@ setup_wasapi_stream_one_side(cubeb_stream * stm,
     if (wasapi_create_device(stm->context, device_info,
                              stm->device_enumerator.get(), device.get(),
                              &default_devices) == CUBEB_OK) {
+      if (device_info.latency_hi == 0) {
+        LOG("Input: could not query latency_hi to guess safe latency");
+        wasapi_destroy_device(&device_info);
+        return CUBEB_ERROR;
+      }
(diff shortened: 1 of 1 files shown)


src/cubeb_wasapi.cpp
```diff
@@ -2291,8 +2291,12 @@ setup_wasapi_stream_one_side(cubeb_stream * stm,
     if (wasapi_create_device(stm->context, device_info,
                              stm->device_enumerator.get(), device.get(),
                              &default_devices) == CUBEB_OK) {
+      if (device_info.latency_hi == 0) {
```
padenot, line 2294: It's better to let this be 0. Windows will then pick something default: see the `hnsPeriodicity` parameter description at https://docs.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-initialize.

padenot (changes requested):

src/cubeb_wasapi.cpp
padenot: I misunderstood, I thought this was a device that was returning 0 but was otherwise working here. Maybe that's possible as well?

PR #710, opened 2022-06-22
fix synchronization mistakes in ring buffer

ad8e:
The index loads are over-optimistically relaxed, so I changed them to the right ones.

### DIFF
--- a/src/cubeb_ringbuffer.h
+++ b/src/cubeb_ringbuffer.h
@@ -110,8 +110,8 @@ public:
     assert_correct_thread(producer_id);
 #endif
 
-    int rd_idx = read_index_.load(std::memory_order_relaxed);
     int wr_idx = write_index_.load(std::memory_order_relaxed);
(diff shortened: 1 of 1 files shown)


src/cubeb_ringbuffer.h
```diff
@@ -110,8 +110,8 @@ template <typename T> class ring_buffer_base {
     assert_correct_thread(producer_id);
 #endif
 
-    int rd_idx = read_index_.load(std::memory_order_relaxed);
     int wr_idx = write_index_.load(std::memory_order_relaxed);
+    int rd_idx = read_index_.load(std::memory_order_acquire);
```
ad8e, line 114: This needs a LoadStore synchronization, hence acquire semantics.

kinetiknz:
Thanks for the fixes!  Fortunately this code has only been used in the logging code, which is run-time disabled by default.

PR #713, opened 2022-07-07
Update build instructions

japm48:
Changes in the instructions:
 - Use cmake version 3.14 as minimum (required as per `CMakeLists.txt`).
 - Build inside a `./build/` directory instead of a parent directory, as it is usual in cmake projects (and it is already in `.gitignore`).
 - Use `cmake -B`.

Note: we could also change `cd build && ctest` to `ctest --test-dir ./build` but that requires cmake v3.20+.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,13 +1,12 @@
 # Build instructions for libcubeb
 
-You must have CMake v3.1 or later installed.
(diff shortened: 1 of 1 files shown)


kinetiknz:
Thanks!

PR #717, opened 2022-07-21
 Add logging statement when input audio dropping happens 

padenot:
I need this to debug something in the field, now that we're close to having a way to easily getting logs.


padenot:
This is trivial, merging now.

PR #722, opened 2022-09-15
cmake: fix install rules

abouvier:
* don't install cmake test binaries
* fix variable names in Doxyfile.in
* install doc files


kinetiknz:
Thanks!

PR #728, opened 2022-10-18
Properly compare against current log level

F0bes:
In the case that the current log level is 0 (Disabled), we don't want to see if that is **less than or equal to** the logs log level, that will always evaluate to true.

Upstreamed from 
https://github.com/PCSX2/pcsx2/pull/7233/commits/a60fd7e9361a6e72eeec9e19ba0a6c0886cbac31


kinetiknz:
Thanks! Looks like this got mixed up in 6b474ee4045561f65eb5a720205307db809812b4.  Merged with syntax fix in 27d2a102b0b75d9e49d43bc1ea516233fb87d778.

PR #730, opened 2022-11-29
Null check g_cubeb_log_callback before using

padenot:
When logging asynchrously, there's ample time for the pointer to be nulled.

### DIFF
--- a/src/cubeb.c
+++ b/src/cubeb.c
@@ -351,6 +351,8 @@ cubeb_destroy(cubeb * context)
   }
 
   context->ops->destroy(context);
+
+  cubeb_set_log_callback(CUBEB_LOG_DISABLED, NULL);
 }
 
 int
--- a/src/cubeb_log.cpp
+++ b/src/cubeb_log.cpp
@@ -65,47 +65,67 @@ public:
   void push(char const str[CUBEB_LOG_MESSAGE_MAX_SIZE])
   {
     cubeb_log_message msg(str);
-    msg_queue.enqueue(msg);
+    msg_queue->enqueue(msg);
   }
   void run()
   {
-    std::thread([this]() {
+    assert(logging_thread.get_id() == std::thread::id());
+    assert(msg_queue);
+    logging_thread = std::thread([this]() {
       CUBEB_REGISTER_THREAD("cubeb_log");
-      while (true) {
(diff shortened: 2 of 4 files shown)


kinetiknz (approved):
I'm not 100% sure about this.  If `g_cubeb_log_callback` can be nulled while async logging is active, then there's still a small window for a race between the null check and use here.  To address that, we'd either need to ensure async logging is never active when this is changed, or use a "logging disabled" no-op callback that is atomically swapped in instead of nulling the callback.

padenot:
> To address that, we'd either need to ensure async logging is never active when this is changed,

I went this route, and did other things as well. It's a lot bigger, a tad more complex, but I added tests (that pass locally when stress tested (a lot more iterations on a bunch of runs) on aarch64/M1 max and x86_64/Linux).

src/cubeb.c
```diff
@@ -350,6 +350,8 @@ cubeb_destroy(cubeb * context)
     return;
   }
 
+  cubeb_set_log_callback(CUBEB_LOG_DISABLED, NULL);
```
kinetiknz, line 353: Is it possible to clear the log callback after `destroy(context)`?

src/cubeb_log.cpp
```diff
@@ -65,47 +65,65 @@ class cubeb_async_logger {
...
+  {
+    msg_queue.reset(
+        new lock_free_queue<cubeb_log_message>(CUBEB_LOG_MESSAGE_QUEUE_DEPTH));
+    shutdown_thread = false;
+    run();
+  }
+  void stop()
+  {
+    shutdown_thread = true;
+    logging_thread.join();
```
kinetiknz, line 100: The test failures look like multiple calls to disable logging hitting the case where it's already been set to `std::thread()`.

kinetiknz (approved):
Thanks for doing this!

src/cubeb.c
padenot: Yes, and it's in fact better if the backend logs in `destroy(context);`.

padenot:
Matthew, I ended up slightly modifying the test to remove `delay(...)` calls and use atomics to ensure that callbacks were called, because otherwise the test was super unreliable on the CI here. I think it's better regardless.

PR #732, opened 2022-12-08
Allow the asynchronous logging system to safely delete its storage

padenot:
Turns out the previous code wasn't safe.

Here I'm using `compare_exchange` to grab the queue exclusively (swapping with a `nullptr`) when pushing a message, in a way that it can't be deallocated.

When deallocating, the thread that wants to do so first waits for the consumer thread to stop, by joining, and then synchronizes with the real-time audio thread by grabbing the queue itself (same way, by setting to `nullptr`). At this point, the real-time audio thread starts dropping the messages on the floor (or has already observered that logging has been disabled, so doesn't log).

We can then proceed with the deallocation.

I've added a stress-test that has found a bunch of issues while running for longs periods of time, and now runs without issues under ASAN and TSAN, on x86_64.

### DIFF
--- a/src/cubeb_log.cpp
+++ b/src/cubeb_log.cpp
@@ -25,7 +25,12 @@ const size_t CUBEB_LOG_MESSAGE_MAX_SIZE = 256;
  * messages. */
 const size_t CUBEB_LOG_MESSAGE_QUEUE_DEPTH = 40;
 /** Number of milliseconds to wait before dequeuing log messages. */
-#define CUBEB_LOG_BATCH_PRINT_INTERVAL_MS 10
+const size_t CUBEB_LOG_BATCH_PRINT_INTERVAL_MS = 10;
+
+void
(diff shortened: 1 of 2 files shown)


src/cubeb_log.cpp
```diff
@@ -86,44 +99,62 @@ class cubeb_async_logger {
...
     }
   }
-
-private:
-  cubeb_async_logger() {}
   /** This is quite a big data structure, but is only instantiated if the
-   * asynchronous logger is used.*/
-  std::unique_ptr<lock_free_queue<cubeb_log_message>> msg_queue;
+   * asynchronous logger is used. The two pointers point to the same object, but
+   * the first one can be temporarily null when a message is beeing enqueued. */
```
kinetiknz, line 153: nit: typo in `beeing`

kinetiknz (approved):

PR #735, opened 2022-12-13
Implement client-side clock interpolation for the AAudio backend, and accurately compute audio latencies

padenot:
I'm going to add commits for the input-only and duplex case later, since this also provides the latencies we need there, but it's going to be more or less the same.

@kinetiknz, with this, I believe we can attempt to reenable AAudio in Firefox.

I've attempted to fix the latency reporting in the OpenSL backend, without much success, but maybe it's doable as well. AAudio's API is a lot clearer.

The latency computation part is essentially what is done [in Oboe](https://github.com/google/oboe/blob/03db05df380486643bcf02eab9b3f1350b8069ad/src/aaudio/AudioStreamAAudio.cpp#L682), written by the Android audio people, so it should be correct.

A/V sync and clock resolution look perfect on my device, with all audio devices I checked.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -400,6 +400,7 @@ if(BUILD_TESTS)
 
   cubeb_add_test(duplex)
   cubeb_add_test(logging)
+  cubeb_add_test(triple_buffer)
 
   if (USE_WASAPI)
     cubeb_add_test(overload_callback)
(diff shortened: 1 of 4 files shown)


kinetiknz (approved):
LGTM, thanks!

PR #738, opened 2023-01-05
Prefer the AAudio backend over the OpenSL ES backend

padenot:
The last issues have been fixed, and it's now better (and also uses an API that isn't deprecated).

This falls back to OpenSL ES on older devices.


ChunMinChang (approved):
Let's try this!

PR #739, opened 2023-01-08
Support jack under NetBSD

ryoon:
Jack works under NetBSD.
Enable strdup(3) for NetBSD like FreeBSD to fix build.


padenot:
Thanks for the patch, this should arrive shortly in Firefox if that's what you care about for this.

PR #740, opened 2023-01-20
Re-enable and polish IAudioClient3 to achieve lower latencies

Filoppi:
I made this PR for Dolphin ([link](https://github.com/dolphin-emu/dolphin/pull/11466)) and it's currently in testing, though I thought to propise to Mozilla as well, to potentially get some feedback.

**This PR re-activates ```IAudioClient3``` on WASAPI**. This allows PCs with Windows 10+ to achieve lower latencies:
[Microsoft doc](https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/low-latency-audio)
theoretically the **latency should be pretty similar to the WASAPI exclusive** mode (depending on the device driver), but without its limitations.
If not supported, it will fall back to the classic/old ```IAudioClient``` audio mode.

Most of the code was already there, though **Mozilla disabled it because it gave problems with input streams** ([bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=1590652)). For that reason, we are only using it with output streams, and additionally, I have polished the code a lot and added handling of a few edge/error cases that were not acknowledged before.
**Chromium has safely used ```IAudioClient3``` for years**, so I have analyzed its implementation, and multiple other open source applications, to find out what might be **improved here**.

One major problem was that the latency set by the user was in samples (48Khz on Dolphin), not in ms, but cubeb failed to convert it for device actual output sample rate, so the latency could have been set to extremely low or high values, which could have made it starve for samples.
~~Also, cubeb failed to check if the mixer format it generated was fully supported by ```IAudioClient3```, and thus sometimes it wouldn't initialize correctly. I'm not 100% sure about this, but I've been testing this with @eliasreid, who can confirm as his audio device only supports 24bit formats, and cubeb forced 16bit on it and Dolphin produced crackled sounds (it didn't seem to be a latency issue). See ```CHECK_MIXER_FORMAT_SUPPORT``` for more.~~ (This was only partially true)

The code is kind of done, but there are quite a few things that **could benefit from testing**.
I've added some defines at the top of the files, which are all meant for testing, and I think the current configuration is the best, but if something is wrong with your output (or ```IAudioClient3``` fails to activate).

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -33,6 +33,31 @@
 #include "cubeb_tracing.h"
 #include "cubeb_utils.h"
 
+// Some people have reported glitches with IAudioClient3 capture streams:
+// http://blog.nirbheek.in/2018/03/low-latency-audio-on-windows-with.html
+// https://bugzilla.mozilla.org/show_bug.cgi?id=1590902
+#define ALLOW_AUDIO_CLIENT_3_FOR_INPUT 0
+// IAudioClient3::GetSharedModeEnginePeriod() seem to return min latencies
+// bigger than IAudioClient::GetDevicePeriod(), which is confusing (10ms vs
+// 3ms), though the default latency is usually the same and we should use the
+// IAudioClient3 function anyway, as it's more correct
+#define USE_AUDIO_CLIENT_3_MIN_PERIOD 1
+// If this is true, we allow IAudioClient3 the creation of sessions with a
+// latency above the default one (usually 10ms).
+// Whether we should default this to true or false depend on many things:
+// -Does creating a shared IAudioClient3 session (not locked to a format)
+//  actually forces all the IAudioClient(1) sessions to have the same latency?
+//  I could find no proof of that.
+// -Does creating a shared IAudioClient3 session with a latency >= the default
+//  one actually improve the latency (as in how late the audio is) at all?
+// -Maybe we could expose this as cubeb stream pref
+//  (e.g. take priority over other apps)?
+#define ALLOW_AUDIO_CLIENT_3_LATENCY_OVER_DEFAULT 1
+// If this is true and the user specified a target latency >= the IAudioClient3
+// max one, then we reject it and fall back to IAudioClient(1). There wouldn't
+// be much point in having a low latency if that's not what the user wants.
+#define REJECT_AUDIO_CLIENT_3_LATENCY_OVER_MAX 0
+
 // Windows 10 exposes the IAudioClient3 interface to create low-latency streams.
 // Copy the interface definition from audioclient.h here to make the code
 // simpler and so that we can still access IAudioClient3 via COM if cubeb was
@@ -1867,6 +1892,44 @@ wasapi_get_min_latency(cubeb * ctx, cubeb_stream_params params,
     return CUBEB_ERROR;
   }
 
+#if USE_AUDIO_CLIENT_3_MIN_PERIOD
+  // This is unreliable as we can't know the actual mixer format cubeb will
+  // ask for later on (nor we can branch on ALLOW_AUDIO_CLIENT_3_FOR_INPUT),
+  // and the min latency can change based on that.
+  com_ptr<IAudioClient3> client3;
+  hr = device->Activate(__uuidof(IAudioClient3), CLSCTX_INPROC_SERVER, NULL,
+                        client3.receive_vpp());
+  if (SUCCEEDED(hr)) {
+    WAVEFORMATEX * mix_format = nullptr;
+    hr = client3->GetMixFormat(&mix_format);
+
+    if (SUCCEEDED(hr)) {
+      uint32_t default_period = 0, fundamental_period = 0, min_period = 0,
+               max_period = 0;
+      hr = client3->GetSharedModeEnginePeriod(mix_format, &default_period,
+                                              &fundamental_period, &min_period,
+                                              &max_period);
+
+      auto sample_rate = mix_format->nSamplesPerSec;
+      CoTaskMemFree(mix_format);
+      if (SUCCEEDED(hr)) {
+        // Print values in the same format as IAudioDevice::GetDevicePeriod()
+        REFERENCE_TIME min_period_rt(frames_to_hns(sample_rate, min_period));
+        REFERENCE_TIME default_period_rt(
+            frames_to_hns(sample_rate, default_period));
+        LOG("default device period: %I64d, minimum device period: %I64d",
+            default_period_rt, min_period_rt);
+
+        *latency_frames = min_period;
+
+        LOG("Minimum latency in frames: %u", *latency_frames);
+
+        return CUBEB_OK;
+      }
+    }
+  }
+#endif
+
   com_ptr<IAudioClient> client;
   hr = device->Activate(__uuidof(IAudioClient), CLSCTX_INPROC_SERVER, NULL,
                         client.receive_vpp());
@@ -1886,18 +1949,8 @@ wasapi_get_min_latency(cubeb * ctx, cubeb_stream_params params,
   LOG("default device period: %I64d, minimum device period: %I64d",
       default_period, minimum_period);
 
-  /* If we're on Windows 10, we can use IAudioClient3 to get minimal latency.
-     Otherwise, according to the docs, the best latency we can achieve is by
-     synchronizing the stream and the engine.
-     http://msdn.microsoft.com/en-us/library/windows/desktop/dd370871%28v=vs.85%29.aspx
-   */
-
-  // #ifdef _WIN32_WINNT_WIN10
-#if 0
-     *latency_frames = hns_to_frames(params.rate, minimum_period);
-#else
+  // The minimum_period is only relevant in exclusive streams.
   *latency_frames = hns_to_frames(params.rate, default_period);
-#endif
 
   LOG("Minimum latency in frames: %u", *latency_frames);
 
@@ -1987,7 +2040,10 @@ handle_channel_layout(cubeb_stream * stm, EDataFlow direction,
   if (hr == S_FALSE) {
     /* Channel layout not supported, but WASAPI gives us a suggestion. Use it,
        and handle the eventual upmix/downmix ourselves. Ignore the subformat of
-       the suggestion, since it seems to always be IEEE_FLOAT. */
+       the suggestion, since it seems to always be IEEE_FLOAT.
+       This fallback doesn't update the bit depth, so if a device
+       only supported bit depths cubeb doesn't support, so IAudioClient3
+       streams might fail */
     LOG("Using WASAPI suggested format: channels: %d", closest->nChannels);
     XASSERT(closest->wFormatTag == WAVE_FORMAT_EXTENSIBLE);
     WAVEFORMATEXTENSIBLE * closest_pcm =
@@ -2031,12 +2087,12 @@ initialize_iaudioclient2(com_ptr<IAudioClient> & audio_client)
   return CUBEB_OK;
 }
 
-#if 0
 bool
 initialize_iaudioclient3(com_ptr<IAudioClient> & audio_client,
                          cubeb_stream * stm,
                          const com_heap_ptr<WAVEFORMATEX> & mix_format,
-                         DWORD flags, EDataFlow direction)
+                         DWORD flags, EDataFlow direction,
+                         REFERENCE_TIME latency_hns)
 {
   com_ptr<IAudioClient3> audio_client3;
   audio_client->QueryInterface<IAudioClient3>(audio_client3.receive());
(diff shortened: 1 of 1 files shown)


Filoppi:
I'm in the process of polishing the code and I've already removed many of the test settings as they revealed useless.
This does fix some bugs though, as just re-enabling `IAudioClient3` without the additional changes can output cracking sounds on some devices.
My next step kind of depends on you if you are interested. Specifically, I was wondering what's your stance on the user/client preferred latency. Is it "mandatory" to respect it, or it is just taken as "target" latency. IAudioClient(1) doesn't always seem to care about that value, so I suppose we should also kind of ignore it with IAudioClient3? Because it's unlikely that the latency asked by the user will be one of the ones accepted by the IAudioClient3 audio engine.

kinetiknz:
Thanks for working on this! We'd like to ship IAC3 support in Firefox at some point, but haven't had a chance to revisit the issues discovered during the previous rollout attempt.  I'll take a look at the changes so far over the next day or so and get back to you with any feedback.

padenot:
> My next step kind of depends on you if you are interested. Specifically, I was wondering what's your stance on the user/client preferred latency. Is it "mandatory" to respect it, or it is just taken as "target" latency. IAudioClient(1) doesn't always seem to care about that value, so I suppose we should also kind of ignore it with IAudioClient3? Because it's unlikely that the latency asked by the user will be one of the ones accepted by the IAudioClient3 audio engine.

It's more or less always been a "desirable target", not all backends can ensure this parameter is respected, so I don't think there's any problem here.

Filoppi:
@kinetiknz @padenot I have polished the code and removed all the test branches that turned out unneccessary from testing.
There are still some defines on the top of the code.
The one that is of most interest to me is ```ALLOW_AUDIO_CLIENT_3_LATENCY_OVER_DEFAULT ```, because a comment in cubeb stated that creating a IAudioSession3 shared stream (not locked to a format) would also lock all IAudioSession(1) streams to its same latency, but I haven't been able to verify that information, nor find it anywhere else. I don't think it is true, so I think we could safely keep the code as if ```ALLOW_AUDIO_CLIENT_3_LATENCY_OVER_DEFAULT``` was 1.

These are some good references for ```IAudioClient3``` code:
https://chromium.googlesource.com/chromium/src/media/+/master/audio/win/core_audio_util_win.cc
https://chromium.googlesource.com/external/webrtc/+/HEAD/modules/audio_device/win/core_audio_utility_win.cc
https://chromium.googlesource.com/external/webrtc/+/HEAD/modules/audio_device/win/core_audio_base_win.cc
https://github.com/sidewayss/cubeb/blob/master/src/cubeb_wasapi.cpp
https://github.com/microsoft/Windows-universal-samples/blob/main/Samples/WindowsAudioSession/cpp/WASAPIRenderer.cpp
https://github.com/microsoft/Windows-classic-samples/blob/main/Samples/Win7Samples/multimedia/audio/CaptureSharedEventDriven/WASAPICapture.cpp
https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/audio-processing-object-architecture
https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/audio-signal-processing-modes

Filoppi:
@kinetiknz don't really want to bother, though it would be cool to know what's your plan with this. If the chances of it getting merged any time soon are small, then I'll consider making a custom branch for Dolphin (as there isn't one now).
Thanks!

padenot:
I want to re-enable this, but haven't had time to sit down and look at your code (thanks for working on this, btw!). I've requested the review for myself, so I'll be reminded to look at it, and I'll find some time to do so, hopefully soon.

Filoppi:
> I want to re-enable this, but haven't had time to sit down and look at your code (thanks for working on this, btw!). I've requested the review for myself, so I'll be reminded to look at it, and I'll find some time to do so, hopefully soon.

I'm sorry. I'm going to ask for an ETA again. If it's still unknown, I'll try to branch this on the dolphin repo.

padenot:
Sorry for the trouble. We can probably land something as long as we can keep the current behaviour, so that we can both do QA and experiment in our own timelines.

Filoppi:
I did not test IAudioClient3 on input, but it's not enabled by default `ALLOW_AUDIO_CLIENT_3_FOR_INPUT`.
You can either test it or leave it off for now. It should be coded correctly, but to avoid delays in merging this I left it off.
Ultimately it's not as important as output.

sidewayss:
fyi - For many users, like me, the entire reason for low latency is for input, so that you can play along with the output by plugging in your electric guitar or keyboard or whatever.  With the currently latency that is simply implausible.

Filoppi:
> We can merge support for output only, and continue iterating on the input side if need be, we've taken too much time to merge this.

All good? Do I need to do anything for this to proceed?

padenot:
> All good? Do I need to do anything for this to proceed?

All good, in the sense that it's green both here, and on the Firefox infra, but that's not real machine with their diverse hardware, drivers and other moving bits. I'm having a final look at the code now. I'm thinking maybe we want to make it a bit more configurable than the current code, I'll take care of that by adding patches on top of your branch. Using flags maybe, Windows only.

All that to say that I'm looking into it, I don't think there's much to do but wait, sorry it takes so long.

src/cubeb_wasapi.cpp
```diff
@@ -1867,6 +1892,44 @@ wasapi_get_min_latency(cubeb * ctx, cubeb_stream_params params,
...
+      CoTaskMemFree(mix_format);
+      if (SUCCEEDED(hr)) {
+        // Print values in the same format as IAudioDevice::GetDevicePeriod()
+        REFERENCE_TIME min_period_rt(frames_to_hns(sample_rate, min_period));
+        REFERENCE_TIME default_period_rt(
+            frames_to_hns(sample_rate, default_period));
+        LOG("default device period: %I64d, minimum device period: %I64d",
+            default_period_rt, min_period_rt);
+
+        *latency_frames = min_period;
```
padenot, line 1923: Looks like `min_period` is in `hns` and `latency_frames` wants to be in frames. We need to check whether we can make this roundtrip properly if we need to have an exact value to trigger a fast path, or if this generally works.

src/cubeb_wasapi.cpp
```diff
@@ -1867,6 +1892,44 @@ wasapi_get_min_latency(cubeb * ctx, cubeb_stream_params params,
     return CUBEB_ERROR;
   }
 
+#if USE_AUDIO_CLIENT_3_MIN_PERIOD
+  // This is unreliable as we can't know the actual mixer format cubeb will
+  // ask for later on (nor we can branch on ALLOW_AUDIO_CLIENT_3_FOR_INPUT),
+  // and the min latency can change based on that.
```
padenot, line 1898: We can probably use a sentinel value for the latency here, if this is a problem.

src/cubeb_wasapi.cpp
```diff
@@ -2081,28 +2135,59 @@ initialize_iaudioclient3(com_ptr<IAudioClient> & audio_client,
...
     LOG("Requested latency %i was adjusted to %i", old_requested_latency,
         requested_latency);
   }
 
-  hr = audio_client3->InitializeSharedAudioStream(flags, requested_latency,
+  DWORD new_flags = flags;
+  // Always add these flags to IAudioClient3, they might help
+  // if the stream doesn't have the same format as the audio engine.
+  new_flags |= AUDCLNT_STREAMFLAGS_AUTOCONVERTPCM;
+  new_flags |= AUDCLNT_STREAMFLAGS_SRC_DEFAULT_QUALITY;
```
padenot, line 2179: `cubeb` handles SRC (using our own resampler with predictable performance and latency, from `speex_dsp`), so this is likely unnecessary. Similarly, `cubeb` has its own channel matrixer, that works in a way that all application code using `cubeb` will work in the same way regardless of the underlying backend.

OTOH, this probably doesn't hurt.

src/cubeb_wasapi.cpp
```diff
@@ -2081,28 +2135,59 @@ initialize_iaudioclient3(com_ptr<IAudioClient> & audio_client,
...
+
+  hr = audio_client3->InitializeSharedAudioStream(new_flags, requested_latency,
                                                   mix_format.get(), NULL);
+  // This error should be returned first even if
+  // the period was locked (AUDCLNT_E_ENGINE_PERIODICITY_LOCKED)
+  if (hr == AUDCLNT_E_INVALID_STREAM_FLAG) {
+    LOG("Got AUDCLNT_E_INVALID_STREAM_FLAG, removing some flags");
```
padenot, line 2186: Have we hit this in practice? I'm unsure what's going on. Naively, I thought that adding the flags in `new_flags` would make the stream **more** likely to be initialized, since converters are used.

padenot: I'll test this with weird setups on a real machine, I think. Is there a well known app that locks the stream params I can use for testing?

src/cubeb_wasapi.cpp
```diff
@@ -2114,22 +2199,37 @@ initialize_iaudioclient3(com_ptr<IAudioClient> & audio_client,
...
 
-  if (current_period >= default_period) {
-    LOG("Current shared mode engine period %i too high, not using IAudioClient",
-        current_period);
+#if REJECT_AUDIO_CLIENT_3_LATENCY_OVER_MAX
+  // Reject IAudioClient3 if we can't respect the user target latency.
+  // We don't need to check against default_latency anymore,
+  // as the current_period is already the best one we could get.
+  if (old_requested_latency > current_period) {
+    LOG("Requested latency %i greater than currently locked shared mode "
```
padenot, line 2227: I don't know about your use case, but this is fine for us I think.

src/cubeb_wasapi.cpp
```diff
@@ -33,6 +33,31 @@
 #include "cubeb_tracing.h"
 #include "cubeb_utils.h"
 
+// Some people have reported glitches with IAudioClient3 capture streams:
+// http://blog.nirbheek.in/2018/03/low-latency-audio-on-windows-with.html
+// https://bugzilla.mozilla.org/show_bug.cgi?id=1590902
+#define ALLOW_AUDIO_CLIENT_3_FOR_INPUT 0
```
padenot, line 39: I might switch those to a number of flags. This way, we can wire them up in application code, and users that are willing to experiment or help us debug can do so without getting test builds and such.

For us it would in in `about:config`, but I suppose any command line flag or settings panel would work.

padenot (commented):
Generally looks good, I'm going to test run it with a collection of weird devices I have accumulated over the years (but there's a few bank holidays coming up here in France, so likely next week). I'm also wondering if Windows 11 or more recent versions of 10 behave properly with input and output (that would be ideal!).

Thanks a lot in any case for the patch, and sorry for the time it takes...

src/cubeb_wasapi.cpp
Filoppi: I put these there for you to make tests and changes.
I am happy for you make any changes you want to the CL, as long as of course its still sourced from this one :).

src/cubeb_wasapi.cpp
Filoppi: `min_period` is in frames: https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient3-getsharedmodeengineperiod

src/cubeb_wasapi.cpp
Filoppi: it's likely to not hurt, I can't know how it works internally, but I'm sure it doesn't do anything if input/output have the same format already.

src/cubeb_wasapi.cpp
Filoppi: Yes, now it's been a while, but I'm pretty sure I got this at some point.

src/cubeb_wasapi.cpp
Filoppi: I think Dolphin could use this as on weak CPUs it can struggle to feed data to the audio stream, and thus having a bigger buffer is preferred and avoids starvation.
That said, I'm ok if this is removed if you prefer so.

Filoppi:
@padenot is there anything specific you'd still like me to address?

Filoppi:
@padenot sorry to bother again, any updates on this?

padenot:
Sorry for the delay. We're almost done with fixing the issues found after merging https://github.com/mozilla/cubeb/pull/542 (on WASAPI, there's still a crash that eludes us), and we're busy fixing two issues on this library: macOS low-latency audio glitches when running duplex streams on specific configurations, and quite a number of crashes on Android. Crashes and existing quality issues take precedence over optimizing existing well functioning code.

That said, I've been playing with this a bit more, and it looks like this is behaving better on recent Windows, so it's likely that when we've fixed those two issues on macOS and Android, we'll get this merged. Having so many users with so diverse hardware running Firefox makes even rare bugs important, and I think it's better for everybody to have stable code on the `master` branch, even if this means some downstream users having to carry patches, or have performance that could be slightly better.

Has this code been running somewhere in production?

Filoppi:
@padenot Nice. No, this code has only been tested by me and a couple of testers online.

padenot:
Yeah, that's the real issue with all this. I appeared to work fine on the machine we have, so we shipped it, and the latencies were indeed lower, but it was a train wreck in the field, and we had to disable it in emergency.

It was a long time ago though.

Filoppi:
@padenot 

Yeah. I understand, but it was like 4-5 years ago now? There's been years of development by MS, and they should have fixed any bugs with it.
Also I think my implementation tackles some issues with the original one and is a bit more safe.

padenot:
Yeah, I've been chatting with other folks that are doing real-time audio on Windows and deploy to lots of random systems, and it seems to be that it's in better shape, so I'm hopeful.

src/cubeb_wasapi.cpp
stenzek: Running this change locally, it seems like the period returned by `GetSharedModeEnginePeriod()` is relative to the mixing sample rate, not the rate passed in `params`. Therefore, if the device is set to a higher sample rate, the latency in frames will incorrectly? be higher, despite it presumably covering the same amount of time due to resampling. Not sure if this was the inaccuracy you mentioned in the comment above.

I ended up changing this to `*latency_frames = hns_to_frames(params.rate, min_period_rt);`, to match the non-IAC3 path below.

Filoppi: I'm willing to go back and test this if @padenot would be happy to consider merging it after :)

mirh:
Bonjour? It has been another 3 years now.

PR #745, opened 2023-03-20
opensl: Attempt to avoid a race with active callbacks during shutdown.

kinetiknz:
Based on examining crash dumps and investigating OpenSL workarounds in liboboe, there is a potential race between active buffer callbacks and stream shutdown, resulting in a buffer callback attempting to use a new buffer after stream destroy has freed the backing allocation.

Ideally there would be a concrete event to watch for to ensure the last callbacks have completed before destroying the stream, but I'm not aware of one and the existing workaround in liboboe relies on an arbitrary ("long enough") sleep between stopping and destroying streams 😞.


padenot (approved):
😢

PR #750, opened 2023-04-24
Change the WinMM backend so that it doesn't call the user callback during initialization, and other fixes

padenot:


ChunMinChang:
> Lots of test pass, but I'm stuck on:
> 
> > D:\a\cubeb\cubeb\src\cubeb_wasapi.cpp:2381:Unable to initialize audio client for capture: 80070005

I don't understand why the cubeb_wasapi failed when its code remains the same.

padenot:
> I don't understand why the cubeb_wasapi failed while its code remains the same.

This PR enables tests on Windows. Until now, we were only building. Windows tests for cubeb are run on treeherder on the mozilla infra, but I'm trying to enable them here. I'm trying to run the tests twice: once with the default backend (this means WASAPI, because we're only testing on recent Windows), and once by forcing `winmm`, to get some coverage.

The problem is that the VMs on GitHub action don't have audio devices, I'm trying to address this.

padenot:
> which seems to be a permission problem, only on the capture side. Probably some configuration we need.

I've been able to repro locally on Windows 11 by toggling this "Microphone access" setting:

Screenshot 2023-04-25 164652

Turns out it's possible to toggle this using the registry, this is what the second step in https://github.com/mozilla/cubeb/pull/750/commits/f8fd4bf52316b81c965a4654225d7de6b77f476a is doing.

padenot:
@kinetiknz, @ChunMinChang, it took some digging and a few tries, but this is now green -- tests are now run using the `winmm` and `wasapi` backends on github actions. The `winmm` backend's behaviour is now aligned a bit more with the other backends.

When running on a backend that doesn't implement audio input (it's really just the `winmm` backend when used as fallback at this point), tests that need to do anything related to audio input are skipped, but all others still run and now succeed on Windows.

If we change the runner type, we'll have to do an adjustment: https://github.com/actions/runner-images/issues/2528#issuecomment-1050295010. If we do it now (on the current version, 2019), the tests fail, because the action errors out, saying the audio service is already running.

Nothing else has been changed (I hope!).

kinetiknz (approved):
That's awesome, thanks!

src/cubeb_winmm.c
```diff
@@ -486,7 +507,11 @@ winmm_stream_init(cubeb * context, cubeb_stream ** stream,
 
   stm->params = *output_stream_params;
 
-  stm->data_callback = data_callback;
+  // Data callback is set to the user-provided data callback after
+  // the initialization and potentiall preroll callback calls are done, because
```
kinetiknz, line 511: nit: Typo in `potentiall`

PR #752, opened 2023-07-25
When using AAudio, reinitialize the stream if a the device in use is disconnected

padenot:
and other cleanups (clang tidy, format).

The important parts are cb626a421b3ab55114914c25e9a80d69fe1a96ad and cb626a421b3ab55114914c25e9a80d69fe1a96ad, the rest is just cleanups and also the ability to log without changing a `#define`. Hopefully the messages are clear enough to understand what's going on, despite the long commits.

This fixes [BMO#1841876](https://bugzilla.mozilla.org/show_bug.cgi?id=1841876).

Some bits of the doc that are useful:

https://developer.android.com/ndk/reference/group/audio#aaudiostream_errorcallback (what can one do or shouldn't do within the error callback)
https://developer.android.com/ndk/reference/group/audio#aaudiostreambuilder_seterrorcallback (explaining what we're doing here -- largely similar to what we do on other backends).
https://developer.android.com/ndk/guides/audio/aaudio/aaudio#disconnected-streams

This also changes the build to use C++17, so it match what Gecko uses. It also update googletest, so it works compiles in C++17.


src/cubeb_aaudio.cpp
```diff
@@ -837,11 +853,60 @@ aaudio_input_data_cb(AAudioStream * astream, void * user_data,
...
+    stream_state state = stm->state.load();
+    bool was_playing = state == stream_state::STARTED ||
+                       state == stream_state::STARTING ||
+                       state == stream_state::DRAINING;
+    int err = aaudio_stream_stop_locked(stm, lock);
+    // error ignored.
+    aaudio_stream_destroy_locked(stm, lock);
+    err = aaudio_stream_init_impl(stm, lock);
+
+    stm->in_use.store(true);
```
padenot, line 874: I think I need to remove this.

kinetiknz: Yeah, looks like `stm->in_use` is only modified in `aaudio_stream_init`/`aaudio_stream_destroy` so should always be `true` in this situation.

kinetiknz (approved):
LGTM, thanks!

PR #753, opened 2023-07-26
wasapi: Avoid reconfiguring inactive streams

kinetiknz:
This is intended to fix [BMO 1838241](https://bugzilla.mozilla.org/show_bug.cgi?id=1838241).  Since https://github.com/mozilla/cubeb/commit/febf49d0898d1c1691235639b0796b91df522394 was merged, the render thread exists for the lifetime of the stream (rather than just the duration between a stream start and stop) - this results in the render thread handling reconfigure events for inactive (stopped) streams, which is not necessary as stream start will handle reconfiguration if the audio device was lost while the stream was stopped.

Moving the acquisition of the stream_reset_lock in 4c6c317a37d1ef6034b3ca117bffa0906650eb1b should be sufficient to avoid the race/crash seen in BMO 1838241. d1e46c727e5284c71880657f493c7580d9cd86ea adds an `active` flag to the stream and uses that to ignore reconfigure events on the render thread for inactive streams, allowing stream start to handle reconfiguration when necessary.

### DIFF
--- a/src/cubeb_wasapi.cpp
+++ b/src/cubeb_wasapi.cpp
@@ -452,6 +452,9 @@ struct cubeb_stream {
    * called and the render loop thread has exited, destroy this stream object.
    */
   LONG ref_count = 0;
+
+  /* True if the stream is active, false if inactive. */
+  bool active = false;
 };
 
 class monitor_device_notifications {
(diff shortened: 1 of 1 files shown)


padenot:
Thanks!

PR #754, opened 2023-07-31
Don't attempt to use an unavailable cubeb_stream in aaudio_stream_init.

padenot:


padenot:
@kinetiknz merging this trivial fix directly.

PR #759, opened 2023-09-22
Add internal utility to log audio streams to disk from real-time audio callbacks

padenot:
I'll expose this to rust next. Example usage in `test_audio_dump.cpp`.

This is going to be quite useful to diagnose exactly what comes out of the system level callbacks, and compare to what is resampled and delivered at the application level, etc.

@Pehrsons

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,6 +78,7 @@ endif()
 
 add_library(cubeb
   src/cubeb.c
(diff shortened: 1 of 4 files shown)


padenot:
This seems to be working well from my testing.

kinetiknz (approved):
LGTM, thanks!

minor nit: externally has a typo (externaly) in a couple of comments

PR #766, opened 2023-11-09
Assorted Android fixes

padenot:
This series does the following:
- fixes a bug on OpenSL (classic confusion audio sample vs. audio frame)
- fixes a crash about logging in OpenSL and AAudio backend (this isn't very good but works well for what we need)
- fixes an issue in which cubeb errored out when attempting to read from an input device that was disconnected: this happened when opening a Bluetooth input that was changing protocol or something like that. This is largely the same logic as other backends: device disconnected -> reinit
- make it so `latency_frames` is the buffer size that will be passed to the callback, allowing aligning buffer size in the entire processing chain, yielding improved performance
- increase the buffer size to be 3 times the burst size (triple buffering), this seems to behave appropriately, but it's likely that I'll do further testing / tuning. This improves robustness significantly.

### DIFF
--- a/src/cubeb_aaudio.cpp
+++ b/src/cubeb_aaudio.cpp
@@ -620,7 +620,9 @@ aaudio_get_latency(cubeb_stream * stm, aaudio_direction_t direction,
   auto result = WRAP(AAudioStream_getTimestamp)(stream, CLOCK_MONOTONIC,
                                                 &hw_frame_index, &hw_tstamp);
   if (result != AAUDIO_OK) {
-    LOG("AAudioStream_getTimestamp failure.");
+    LOG("AAudioStream_getTimestamp failure for %s: %s",
+        is_output ? "output" : "input",
+        WRAP(AAudio_convertResultToText)(result));
     return 0;
   }
 
@@ -689,9 +691,6 @@ aaudio_duplex_data_cb(AAudioStream * astream, void * user_data,
   stream_state state = atomic_load(&stm->state);
   int istate = WRAP(AAudioStream_getState)(stm->istream);
   int ostate = WRAP(AAudioStream_getState)(stm->ostream);
-  ALOGV("aaudio duplex data cb on stream %p: state %ld (in: %d, out: %d), "
-        "num_frames: %ld",
-        (void *)stm, state, istate, ostate, num_frames);
 
   // all other states may happen since the callback might be called
   // from within requestStart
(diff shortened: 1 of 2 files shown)


src/cubeb_aaudio.cpp
```diff
@@ -712,7 +715,13 @@ aaudio_duplex_data_cb(AAudioStream * astream, void * user_data,
   long in_num_frames =
       WRAP(AAudioStream_read)(stm->istream, stm->in_buf.data(), num_frames, 0);
   if (in_num_frames < 0) { // error
-    stm->state.store(stream_state::ERROR);
+    if (in_num_frames == AAUDIO_STREAM_STATE_DISCONNECTED) {
+      LOG("AAudioStream_read: %s (reinitializing)",
+          WRAP(AAudio_convertResultToText)(in_num_frames));
+      reinitialize_stream(stm);
```
kinetiknz, line 722: It looks like we try to handle this in `aaudio_error_cb` also, does that error callback not fire in this case?  Seems fine to trigger reinit from here, just curious about the behaviour.

src/cubeb_aaudio.cpp
```diff
@@ -1120,12 +1120,28 @@ aaudio_stream_init_impl(cubeb_stream * stm, lock_guard<mutex> & lock)
     WRAP(AAudioStreamBuilder_setUsage)(sb, output_preset);
     WRAP(AAudioStreamBuilder_setDirection)(sb, AAUDIO_DIRECTION_OUTPUT);
     WRAP(AAudioStreamBuilder_setDataCallback)(sb, out_data_callback, stm);
+    assert(stm->latency_frames < std::numeric_limits<int>::max());
+    LOG("Frames per callback set to %d for output", stm->latency_frames);
+    WRAP(AAudioStreamBuilder_setFramesPerDataCallback)(sb, static_cast<int>(stm->latency_frames));
```
kinetiknz: nit: int32_t rather than int to match the signature of setFramesPerDataCallback, but it's not a big deal given int is 32 bits on sane platforms.

kinetiknz (approved):
Awesome, thanks!

src/cubeb_aaudio.cpp
padenot: This was curious to me as well, but our colleague Dan can repro 100% of the time by joining a call with his bluetooth earbuds **connected prior to the call** (so I think it's caused by a bluetooth protocol change), in Firefox, on a Pixel 6 with Android 14. I cannot reproduce this behaviour on my Pixel 4 with Android 13 or other devices I've tried.

When reading Android's sources, this is what I see:

https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/libaaudio/src/client/AudioStreamInternalCapture.cpp;l=246;drc=ac0eae95447f4ac9200cf2c25f938e877d1b4a78;bpv=1;bpt=1 is the location where this callback can be called, in the callback loop. We see this returns early if no callback is set (which is our case on the input side, since we call `AAudioStream_read(...)` from the output callback). I don't see this called at another location in this file.

However, what I see is this: https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/libaaudio/src/legacy/AudioStreamLegacy.cpp;l=223;drc=ac0eae95447f4ac9200cf2c25f938e877d1b4a78;bpv=1;bpt=1, in the `/legacy/` folder with the `Legacy` suffix on the file name, and this calls the error callback when disconnected, so I think this matches what we're seeing: it's called on older android version, not called on more recent android version.

PR #772, opened 2023-12-22
audiounit: Resample to match output device hardware sample rate.

Steveice10:
In some cases, weird sample rates can cause AudioUnit output units to misbehave. Instead, use resampling for non-native rates to make sure this does not happen.

Please let me know if there's any extra cases I need to cover in the code for this, I've tested with basic stream playback via a callback at 32768 Hz and this fixes the issue I was having.


src/cubeb_audiounit.cpp
```diff
@@ -2709,11 +2710,18 @@ audiounit_setup_stream(cubeb_stream * stm)
     input_unconverted_params.rate = stm->input_hw_rate;
   }
 
+  cubeb_stream_params output_unconverted_params;
+  if (has_output(stm)) {
+    output_unconverted_params = stm->output_stream_params;
+    /* Use the rate of the output device. */
+    output_unconverted_params.rate = stm->output_hw_rate;
```
padenot, line 2717: We could imagine this being a pass-through for common rates: 44100, 48000, some of their multiple and divisors, that are common (8k, 16k, 32k, 22.05k, 88.2k, 96k, etc.). It this something you can put in?

If we do this, the OS can sometimes reconfigure itself to not resample, or to resample more efficiently.

Looks good otherwise.

Steveice10: If I'm understanding correctly, we would just skip using the hardware rate here for those common rates?

padenot: Yes, we'd continue using the OS's resampler for those common rates. It should be equivalent, except that our resampler induces a certain latency (like most resamplers), which we can try to avoid or minimize by using OS facility.

Steveice10: Okay, logic revised.

padenot (approved):
Thanks a lot for the patch and addressing my comments, I'm merging this now.

PR #775, opened 2023-12-29
Checking buffer, buffer_in, buffer_out pointers for non NULL values

Monsterovich:
Fixes https://github.com/mozilla/cubeb/issues/774


padenot (approved):
Thanks for the patch, this should reach Firefox shortly.

PR #777, opened 2024-02-01
Update GitHub Actions macOS builders

aeiouaeiouaeiouaeiouaeiouaeiou:
- According to the [roadmap](https://github.com/actions/runner-images/issues/9255) all pre macos-13 VMs will be removed by Q4 2024.
- Enable Dependabot to track changes to the CI configuration.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ jobs:
       BUILD_TYPE: ${{ matrix.type }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, windows-2019, macos-11]
+        os: [ubuntu-20.04, windows-2019, macos-13, macos-14]
         type: [Release, Debug]
 
     steps:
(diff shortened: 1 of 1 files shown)


aeiouaeiouaeiouaeiouaeiouaeiou:
Interesting why the test step freezes on both builders.

Pehrsons:
@aeiouaeiouaeiouaeiouaeiouaeiou  I found eventually this is because of microphone permissions on the runners. A screenshot will reveal it. See https://github.com/mozilla/cubeb-coreaudio-rs/pull/211 for the fix in the cubeb-coreaudio-rs backend, and https://github.com/actions/runner-images/issues/9330 for a filed upstream issue.

aeiouaeiouaeiouaeiouaeiouaeiou:
@Pehrsons installing the BlackHole driver seems to have helped here as well.

Pehrsons:
Ah what was the issue without BlackHole? The runners seem to have the NullAudio virtual devices (from an Apple code sample, AFAICT) out of the box.

PR #778, opened 2024-02-07
384kHz ought to be enough for anybody

padenot:
Bug 1877319 is caused by someone using a sound card running at 384kHz. It's the max I've seen in products online.


Pehrsons (approved):
LGTM with the comment in cubeb.h updated also.

PR #783, opened 2024-04-15
alsa: make dummy stream_set_name to suppress warnings in terminal

chelovechishko:
Just suppress warnings in terminal every time you start (stop) audio playing, when you start browser from console.


padenot:
This is best fixed in Firefox.

padenot:
Wrote a patch for this here: https://bugzilla.mozilla.org/show_bug.cgi?id=1891709, thanks for reporting, I reckon it can be annoying.

chelovechishko:
Lol.

PR #784, opened 2024-04-29
Invalidate timing info buffers when destorying AAudio stream.

jhlin:
`aaudio_stream_get_position()` returns incorrect result because `aaudio_stream_init()` recycled destroyed stream where the `timing_info` buffers contain stale data.

### DIFF
--- a/src/cubeb_aaudio.cpp
+++ b/src/cubeb_aaudio.cpp
@@ -1049,6 +1049,8 @@ aaudio_stream_destroy_locked(cubeb_stream * stm, lock_guard<mutex> & lock)
     stm->istream = nullptr;
   }
 
+  stm->timing_info.invalidate();
+
   if (stm->resampler) {
     cubeb_resampler_destroy(stm->resampler);
     stm->resampler = nullptr;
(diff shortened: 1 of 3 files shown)


jhlin:
This is to fix Fenix bug [1872354](https://bugzilla.mozilla.org/show_bug.cgi?id=1872354)

src/cubeb_aaudio.cpp
```diff
@@ -1014,6 +1014,8 @@ aaudio_stream_destroy_locked(cubeb_stream * stm, lock_guard<mutex> & lock)
          stm->state == stream_state::ERROR ||
          stm->state == stream_state::SHUTDOWN);
 
+  stm->timing_info.invalidate();
```
kinetiknz, line 1017: I think `invalidate` should happen after after `ostream` and `istream` are stopped, otherwise there's the potential for an audio callback to call `compute_and_report_latency_metrics` between `invalidate` and the rest of the stream destruction.

kinetiknz (changes requested):
Thanks for fixing this! One comment, but overall this looks good to me.

src/cubeb_aaudio.cpp
jhlin: Good catch! I've amended the commit and moved `invalidate()` to after streams stopped. Thanks for the review.

kinetiknz (approved):
Thanks!

PR #786, opened 2024-05-22
Clamp to positive value when computing latency for an output stream

padenot:
Diagnosed by jolin@mozilla.com, the same code is also found in Chromium.

@jhlin r? (I'm not sure how to flag you as reviewer on GitHub).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,7 +81,7 @@ jobs:
 
     - name: Configure CMake
       shell: bash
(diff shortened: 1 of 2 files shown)


jhlin:
LGTM. r+.

jhlin:
@kinetiknz I uploaded the patch to the try server and got [build breaks for 64-bit Android targets](https://treeherder.mozilla.org/jobs?repo=try&revision=d7bf446a9c8d5eb71c0c88a18bf658d0ea9de6d7) while all GitHub checks passed. Are `build-android` 32-bit targets only?

padenot:
Fixed in more recent patches, if you can take another look @jhlin ? Commit [2d8568a](https://github.com/mozilla/cubeb/pull/786/commits/2d8568aedcdf77bb750dab2a618301aad73ff1ae) made it reproduce the failure you saw on Treeherder, and then the two subsequent commits fixed it.

PR #789, opened 2024-06-21
AAudio: don't count the time of stopped stream and replace stop request with pause. 

jhlin:
If a stream has previously been stopped and then started, the elapsed time in between should not be included when calculating the position.

Also, `AAudioStream_requestStop()` doesn't behave as expected.

### DIFF
--- a/src/cubeb_aaudio.cpp
+++ b/src/cubeb_aaudio.cpp
@@ -23,6 +23,7 @@
 #include <memory>
 #include <mutex>
 #include <thread>
+#include <variant>
 #include <vector>
 
 using namespace std;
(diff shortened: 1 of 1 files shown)


src/cubeb_aaudio.cpp
```diff
@@ -135,6 +136,56 @@ struct AAudioTimingInfo {
...
+  void clear()
+  {
+    std::lock_guard lock(mutex);
+    if (time.index() == 1 || std::get<0>(time) > 0) {
+      time.emplace<0>(0);
+    }
+  }
+
+private:
+  std::mutex mutex;
```
padenot, line 178: Unfortunately, we can't use locks here because we're on a real-time thread. Thankfully, we only have a few arithmetic operations to do here so we can use `std::atomic`.

padenot (changes requested):

src/cubeb_aaudio.cpp
jhlin: Thanks a lot for the review! I've updated the logic to use `std::atomic` rather than `std::mutex`. Please let me know if any issues.

jhlin:
In the latest revision, the synchronization logic was removed—access to states/values only in user-called functions.

jhlin:
Not sure why [the test failed on Windows](https://github.com/mozilla/cubeb/actions/runs/9686094771/job/26727735282?pr=789). The patches only affect Android behavior.

jhlin:
@Pehrsons @padenot suggested I seek your help for review. Could you please help? Please let me know if you have any questions. Thanks!

Pehrsons:
The windows failure is probably an intermittent. It succeeded after a re-run.

Just looking fundamentally at this I am on board with fixing the interpolation (we also have other issues in the area, see https://bugzilla.mozilla.org/show_bug.cgi?id=1902263).
But why do we want to pause rather than stop? Is the concern that `AAudioStream_requestStop` flushes, so throws away some data that we lose track of?

I see an issue:
If the client calls `AAudioStream_requestPause` via `cubeb_stream_stop`, then after some time and some seeks (i.e. the next audio frame the client will write is not immediately after the last frame it wrote) have passed, it calls `AAudioStream_requestStart` via `cubeb_stream_start`, wouldn't the `AAudioStream` initially play old audio? That seems wrong.

That the flushed data doesn't reach the user's ears doesn't seem like a problem. It's short enough so they're unlikely to notice. For the sake of `cubeb_stream_get_position` though, we should probably treat the flushed audio as played.

Let me know if I understood this right!

jhlin:
Thanks a lot for the review!

> The windows failure is probably an intermittent. It succeeded after a re-run.
> 
> Just looking fundamentally at this I am on board with fixing the interpolation (we also have other issues in the area, see https://bugzilla.mozilla.org/show_bug.cgi?id=1902263). But why do we want to pause rather than stop? Is the concern that `AAudioStream_requestStop` flushes, so throws away some data that we lose track of?

Exactly, the already buffered data in the previous data callback will be flushed by `AAudioStream_requestStop()` but not `AAudioStream_requestPause()`.

> 
> I see an issue: If the client calls `AAudioStream_requestPause` via `cubeb_stream_stop`, then after some time and some seeks (i.e. the next audio frame the client will write is not immediately after the last frame it wrote) have passed, it calls `AAudioStream_requestStart` via `cubeb_stream_start`, wouldn't the `AAudioStream` initially play old audio? That seems wrong.

Luckily, seeking destroys then initializes audio stream, so old data isn't played.

> 
> That the flushed data doesn't reach the user's ears doesn't seem like a problem. It's short enough so they're unlikely to notice. For the sake of `cubeb_stream_get_position` though, we should probably treat the flushed audio as played.

TBH, I never noticed skipped audio. I only found that some video frames were dropped when resuming a paused video because the audio clock advanced further than expected. This change didn't eliminate the issue completely in my tests, but it seems mitigated.

Besides, this behavior seems inconsistent with the other backends.

> 
> Let me know if I understood this right!

Pehrsons:
> Luckily, seeking destroys then initializes audio stream, so old data isn't played.

Ok, so that's true for Gecko but might not be for other clients.

> TBH, I never noticed skipped audio. I only found that some video frames were dropped when resuming a paused video because the audio clock advanced further than expected. This change didn't eliminate the issue completely in my tests, but it seems mitigated.

Makes sense that some dropped video frames is more noticable.

> Besides, this behavior seems inconsistent with the other backends.

Out of curiosity I went through the common ones as I'm not familiar with all:
- CoreAudio [is unclear](https://developer.apple.com/documentation/audiotoolbox/1440513-audiooutputunitstop?language=objc).
- Pulse [corks, which doesn't flush](https://freedesktop.org/software/pulseaudio/doxygen/stream_8h.html#a14e698233ac2d246646651955ab0ec7b), unless the data cb already told it to drain, then it [waits for the drain](https://searchfox.org/mozilla-central/rev/6b17ef1f53b0bc68a0d746599c01ad8723c80880/third_party/rust/cubeb-pulse/src/backend/stream.rs#652-657).
- Wasapi [doesn't seem to flush](https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-stop).

Paul confirms that cubeb_stream_stop is more for pausing also, let me take a deeper look in a bit.

src/cubeb_aaudio.cpp
```diff
@@ -432,12 +490,12 @@ update_state(cubeb_stream * stm)
       }
       break;
     case stream_state::STOPPING:
-      assert(!istate || istate == AAUDIO_STREAM_STATE_STOPPING ||
-             istate == AAUDIO_STREAM_STATE_STOPPED);
-      assert(!ostate || ostate == AAUDIO_STREAM_STATE_STOPPING ||
-             ostate == AAUDIO_STREAM_STATE_STOPPED);
-      if ((!istate || istate == AAUDIO_STREAM_STATE_STOPPED) &&
-          (!ostate || ostate == AAUDIO_STREAM_STATE_STOPPED)) {
```
Pehrsons, line 440: It seems to me stopping and stopped should be considered invalid or handled.

src/cubeb_aaudio.cpp
```diff
@@ -135,6 +136,66 @@ struct AAudioTimingInfo {
...
+
+private:
+  // Union value to keep track of time elapsed in stop state.
+  // <0>: timestamp when entering stop state, or 0 (zero) when not in use.
+  // <1>: elapsed time between stop and start states.
+  // Valid states:
+  // index == 0 && get<0> == 0: stream has no stop record
+  // index == 0 && get<0> > 0: stream stopped
+  // index == 1: stream stopped and then started
+  std::variant<uint64_t, uint64_t> time;
```
Pehrsons, line 194: If these `uint64_t`s could be distinct types so you can do something like `std::get<StopTimestamp>(time)`, I think this would be easier to read.

src/cubeb_aaudio.cpp
```diff
@@ -135,6 +136,66 @@ struct AAudioTimingInfo {
   uint32_t input_latency;
 };
 
+class position_interpolator {
```
Pehrsons, line 139: Please give a high-level overview of what this class' intended behavior is. Expected inputs and outputs and that sort of thing.

src/cubeb_aaudio.cpp
```diff
@@ -135,6 +136,66 @@ struct AAudioTimingInfo {
...
+    }
+    time.emplace<0>(timestamp - prev);
+  }
+
+  void start(uint64_t timestamp)
+  {
+    assert(time.index() == 0 && (std::get<0>(time) > 0 /* stopped */ ||
+                                 callback_timestamp == 0 /* inited */));
+    if (std::get<0>(time) > 0) {
+      // Has been stopped: calculate elipsed time.
```
Pehrsons, line 158: s/elipsed/elapsed/
(and below)

Pehrsons (commented):
Some early feedback. I'll have to look closer at the interpolator later.

Pehrsons (approved):
The logic seems to check out to me, thanks!
I left some ideas for making it easier to read. Do with them as you wish.
I also left a comment about AAUDIO_STREAM_STATE_{STOPPING|STOPPED}. Please at a minimum address this one before landing.

src/cubeb_aaudio.cpp
jhlin: Sorry but I don't follow. In the patch, only `AAUDIO_STREAM_STATE_{PAUSING|PAUSED}` are allowed, and `AAUDIO_STREAM_STATE_{STOPPING|STOPPED}` are considered invalid. Am I missing something?

src/cubeb_aaudio.cpp
jhlin: Good idea. I will do it in the next revision. Thanks!

src/cubeb_aaudio.cpp
jhlin: > Please give a high-level overview of what this class' intended behavior is. Expected inputs and outputs and that sort of thing.

I will add comments for the class in the next revision.

> It seems to me we are really tracking the time spent in the starting/started states with a stale callback timestamp.

Agree. Calling `AAudioStream_requestStop()` does generate a new callback invocation immediately after starting to get more data, but since it flushes previously buffered data, the position always jumps ahead, and video frames get dropped. I hope calling `AAudioStream_requestPause()` can make it less serious.

src/cubeb_aaudio.cpp
Pehrsons: Re invalid I was thinking higher up under `handle invalid stream states`. Maybe we cannot do that because we still call `AAudioStream_requestStop` in some other cases?

jhlin: Thanks for the explanation. Yes, since [`AAudioStream_requestStop()` is called in case `stream_state::DRAINING`](https://github.com/mozilla/cubeb/blob/master/src/cubeb_aaudio.cpp#L406), I don't think `AAUDIO_STREAM_STATE_{STOPPING|STOPPED}` need to be listed in the `handle invalid stream states` conditions.

PR #791, opened 2024-07-04
Reset stream params on init

Pehrsons:
Without a reset, a duplex stream being reused as an output stream will open an input stream and provide an input buffer, which is not expected.


kinetiknz (approved):

kinetiknz:
Thanks!

PR #797, opened 2024-08-30
wasapi: Fix default device handling.

mutexlox-signal:
Prior to https://github.com/mozilla/cubeb/pull/682, a device could be default for all of
MULTIMEDIA, NOTIFICATION, and VOICE.

After that PR, the first two were mutually exclusive with the third, I
believe unintentionally.

Restore this behavior so that, e.g., on devices with only one output device,
querying for the "voice" device returns that device.


mutexlox-signal:
@padenot could you please take a look? :)

mutexlox-signal:
Thank you!

PR #798, opened 2024-09-13
windows: Set debug options if appropriate.

mutexlox-signal:
If building with debug, also build standard libraries with debug. This ensures that any libraries linking against cubeb won't have mismatched symbol definitions.


mutexlox-signal:
I think this needs to be slightly smarter to specify dynamic vs static, will try to address that...

padenot:
Look promising, can you fix the build of tests?

kinetiknz (approved):
Thanks for fixing this!

PR #799, opened 2024-09-18
Add submodules for tier-1 rust backends.

mutexlox-signal:
Per https://github.com/mozilla/cubeb/wiki/Backend-Support the rust backends for pulse and coreaudio are tier-1 and the older C++ backends are tier-4. So, add these backends to cubeb by default.

I can also set BUILD_RUST_LIBS to ON by default in this PR, but I'm not sure if there's some reason to avoid doing that.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,11 @@
 [submodule "cmake/sanitizers-cmake"]
 	path = cmake/sanitizers-cmake
 	url = https://github.com/arsenm/sanitizers-cmake
+[submodule "src/cubeb-coreaudio-rs"]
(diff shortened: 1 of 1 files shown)


mutexlox-signal:
cc @padenot  for a review :)

padenot:
302 @kinetiknz, who last looked at maybe merging those (iirc?).

mutexlox-signal:
Hi @kinetiknz , any chance of taking a look here? :)

kinetiknz:
I'll eventually merge the repositories, since it's a pain to make ABI-breaking changes (among other things) with the current structure.  This is a good first step though, so let's merge it.  Thanks!

PR #802, opened 2024-10-11
Add a log callback that takes a formatted message.

mutexlox-signal:
In some contexts (e.g. cubeb-rs), it is not possible to pass a method that formats a string (for instance, because `c_variadic` is not stable in rust, and because it only supports compile-time constant format strings).

So, add a callback method that takes a `const char*` that has already been printf-style formatted.


mutexlox-signal:
cc @padenot  :)

mutexlox-signal:
gentle ping on this :)

padenot:
I don't remember exactly how we set things up, aren't the macros exposed on the rust side as well? There's a bit of custom code for rust.

mutexlox-signal:
This is for capturing logs and saving them to output, rather than adding to logs. I do not see anything in cubeb-rs right now that turns on logs (https://github.com/search?q=repo%3Amozilla%2Fcubeb-rs%20cubeb_set_log_callback&type=code or https://docs.rs/cubeb/latest/cubeb/), but if I'm missing something pointers would be appreciated!

mutexlox-signal:
The problem I'm trying to solve is that https://docs.rs/cubeb/latest/cubeb/ffi/fn.cubeb_set_log_callback.html is not really usable for the reasons mentioned in the original comment: the lack of c_variadic, the requirement for compile-time constant format strings, plus also the fact that rust does not use C format string syntax.

ChunMinChang:
If you'd like to have a log with variadic parameters, something like this should work: https://github.com/mozilla/cubeb-coreaudio-rs/commit/15246000d7ca150b3c271a82922b22fa293d8200

mutexlox-signal:
I don't believe that that does what I want, as my ultimate goal is to be able to write the logs from cubeb to the same file as those for the rest of my library, using rust's logging macros, rather than just to write them to stdout (or even any specific fd).  If I just wrote the logs to an fd, they would be out-of-order with the messages that go through rust's logging macros.

If it were possible to use something like https://github.com/mozilla/cubeb-coreaudio-rs/commit/15246000d7ca150b3c271a82922b22fa293d8200 to generate a `char*` that I could then pass to rust's logging macros, that would work, but I do not believe that's possible.

mutexlox-signal:
though, I suppose i could write an `extern C` function in rust that takes a char* and call that from the callback... hmm, let me try that.

mutexlox-signal:
Thanks for the pointer, I was able to make this work in https://github.com/mozilla/cubeb-rs/pull/98 without any changes to cubeb.

PR #807, opened 2024-12-28
cmake: fix build with ninja generator

abouvier:
Fix this error when using `Ninja` generator and `Rust` backend `cmake -G Ninja -D BUILD_RUST_LIBS=ON ...`
```
ninja: error: '/build/cubeb/src/cubeb-pulse-rs/target/release/libcubeb_pulse.a', needed by 'libcubeb.so.0.0.0', missing and no known rule to make it
```
Because of [this](https://cmake.org/cmake/help/latest/command/add_dependencies.html) and [that](https://cmake.org/cmake/help/latest/policy/CMP0058.html).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -345,6 +345,7 @@ if(USE_PULSE AND USE_PULSE_RUST)
     DOWNLOAD_COMMAND ""
     CONFIGURE_COMMAND ""
     BUILD_COMMAND cargo build --features=gecko-in-tree COMMAND cargo build --release --features=gecko-in-tree
+    BUILD_BYPRODUCTS "${PROJECT_SOURCE_DIR}/src/cubeb-pulse-rs/target/$<IF:$<CONFIG:Debug>,debug,release>/libcubeb_pulse.a"
     BUILD_ALWAYS ON
     BINARY_DIR "${PROJECT_SOURCE_DIR}/src/cubeb-pulse-rs"
     INSTALL_COMMAND ""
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -345,6 +345,7 @@ if(USE_PULSE AND USE_PULSE_RUST)
     DOWNLOAD_COMMAND ""
     CONFIGURE_COMMAND ""
     BUILD_COMMAND cargo build --features=gecko-in-tree COMMAND cargo build --release --features=gecko-in-tree
+    BUILD_BYPRODUCTS "${PROJECT_SOURCE_DIR}/src/cubeb-pulse-rs/target/$<IF:$<CONFIG:Debug>,debug,release>/libcubeb_pulse.a"
```
mutexlox-signal, line 348: drive-by: will this work if libcubeb-pulse is cross-compiled for a different architecture?

abouvier: > drive-by: will this work if libcubeb-pulse is cross-compiled for a different architecture?

Why not? It's the same file that is referenced by the `target_link_libraries` call below.

mutexlox-signal: Ah right, thanks for pointing that out. I think that's broken, too (cross-compiled rust would live in target/arch-triple/{debug,release}, if i recall correctly), so this won't make it broken in any _new_ ways.

mutexlox-signal:
cc @kinetiknz

JoshuaVandaele (approved):
Tested on Linux, indeed fixed using the Ninja generator.

PR #811, opened 2025-01-23
Fix input_needed_for_output to not return one less than needed, causing draining down the line

padenot:
The calculation was incorrect. This meant that in some circumstances,
e.g. playing a 44.1Hz stream on a 96kHz interface, with a block of 441
frames on Windows, rounding was eventually returning 440 instead of 441
in the callback, leading to the stream entering the draining phase.


padenot:
This fixes [BMO#1940319](https://bugzilla.mozilla.org/show_bug.cgi?id=1940319).

test/test_resampler.cpp
```diff
@@ -1160,5 +1165,61 @@ TEST(cubeb, individual_methods)
...
+          block_size = target_rate / 100;
+        }
+        cubeb_stream_params out_params = {};
+        out_params.channels = 2;
+        out_params.rate = target_rate;
+        out_params.format = CUBEB_SAMPLE_FLOAT32NE;
+
+        cubeb_resampler * resampler = cubeb_resampler_create(
+            nullptr, nullptr, &out_params, source_rate, data_cb, nullptr,
+            CUBEB_RESAMPLER_QUALITY_DEFAULT, CUBEB_RESAMPLER_RECLOCK_NONE);
```
karlt, line 1206: CUBEB_RESAMPLER_QUALITY_VOIP should make the test run a bit faster.
I don't expect the resampler latency to affect any of the input output block size arithmetic after the prefill at intialization.

karlt (approved):

karlt (approved):

src/cubeb_resampler_internal.h
```diff
@@ -288,15 +288,18 @@ template <typename T> class cubeb_resampler_speex_one_way : public processor {
     assert(output_frame_count >= 0); // Check overflow
     int32_t unresampled_frames_left =
         samples_to_frames(resampling_in_buffer.length());
-    int32_t resampled_frames_left =
-        samples_to_frames(resampling_out_buffer.length());
```
karlt, line 292: Please explain in the commit message that `resampling_out_buffer.length()` is removed from the equation because resampled frames are not carried over from one `output()` operation to the next.

src/cubeb_resampler_internal.h
```diff
@@ -288,15 +288,18 @@ template <typename T> class cubeb_resampler_speex_one_way : public processor {
     assert(output_frame_count >= 0); // Check overflow
```
karlt, line 288: >   * exactly `output_frame_count` resampled frames. This can return a number
>   * slightly bigger than what is strictly necessary, but it guaranteed that the
>   * number of output frames will be exactly equal. */

The description is not really describing what the function is doing.

Unless this function knows the internal state of the speex resampler, it cannot know how to produce exactly `output_frame_count` resampled frames from the speex resampler.
- Consider resampling at a ratio of 2 input frames to 3 output frames.  Sometimes an input of 1 frame will produce 1 frame of output, but sometimes 2 frames of output.

The `cubeb_resampler_speex_one_way::output()` should return exactly `output_frame_count` frames, but that is because `output_frame_count` is passed to `output()`.

The function can calculate a number of input frames that is known sufficient to produce (at least) `output_frame_count` resampled frames.

src/cubeb_resampler_internal.h
```diff
@@ -288,15 +288,18 @@ template <typename T> class cubeb_resampler_speex_one_way : public processor {
...
-    int32_t resampled_frames_left =
-        samples_to_frames(resampling_out_buffer.length());
-    float input_frames_needed =
-        (output_frame_count - unresampled_frames_left) * resampling_ratio -
-        resampled_frames_left;
-    if (input_frames_needed < 0) {
+    float input_frames_needed_frac =
+        static_cast<float>(output_frame_count) * resampling_ratio;
+    if (input_frames_needed_frac < 0) {
       return 0;
```
karlt, line 294: `resampling_ratio` is [always non-negative](https://searchfox.org/mozilla-central/rev/a965e3c683ecc035dee1de72bd33a8d91b1203ed/media/libcubeb/src/cubeb_resampler_internal.h#174-175,177), as is `length()`, so float arithmetic will always produce a non-negative `input_frames_needed_frac`, and so this path is never taken.

I'll uncomfortable with using `float` arithmetic because I don't know that we can prove that `n / m * (i * m) >= n * i` over the ranges of rates that we care about.  Replacing `resampling_ratio` with a fraction, with integer arithmetic, would provide assurance of this.  We'd then want some checks on the ranges for the integer arithmetic.

src/cubeb_resampler_internal.h
```diff
@@ -288,15 +288,18 @@ template <typename T> class cubeb_resampler_speex_one_way : public processor {
...
     }
-    return (uint32_t)ceilf(input_frames_needed);
+    auto input_frame_needed =
+        static_cast<uint32_t>(ceilf(input_frames_needed_frac));
+    if (static_cast<float>(unresampled_frames_left) > resampling_ratio &&
+        resampling_ratio > 1) {
+      input_frame_needed -= unresampled_frames_left;
```
karlt, line 300: I can't think why `unresampled_frames_left` should be ignored when less than `resampling_ratio` nor when `resampling_ratio <= 1`.

Can `unresampled_frames_left` always be subtracted from `input_frame_needed`?

The main defect that I see in the old code is that `unresampled_frames_left` was subtracted from `output_frame_count` rather than from an input frame count.
Did you observe a problem with that defect resolved?
If so, then what were the specific parameters?

karlt (changes requested):
I see now that github interprets review of a commit message as review of the entire pull request.

Thank you for testing thoroughly.
I expect that this function should be able to be simple without special cases. The special cases make me suspect that something else may not be quite right.

src/cubeb_resampler_internal.h
karlt: This was an attempt with always subtracting `unresampled_frames_left`, which fails with 16000 -> 44100 wanting 96 output samples.  I'm still looking into this:
https://pernos.co/debug/KvmgNpBzyzyI7mFKNzXHhw/index.html

padenot: My reasoning here is that I wanted to limit the inflation of the input buffer request. In previous iterations of this fix, I had something that passed all test cases. However, looking closely at the large number of debugging printouts I had added, the resampling input buffer was becoming bigger and bigger, because I wasn't taking it into account.

I then decided to still subtract from the input frames needed, but only if we're more or less sure that we're not going back to the situation in which we don't feed the resampler enough frames. This depends on the resampling ratio.

karlt: When `resample_ratio < 1`, the number of frames carried in `resampling_in_buffer` is never included in the calculation and so could accumulate, I assume due to the rounding up.  At least it is hard to prove from speex resampler logic that they would not.

The `static_cast<float>(unresampled_frames_left) > resampling_ratio` test means that additional frames are provided sometimes, but sometimes not, and it's hard to reason about the cadence providing additional frames at the right time.

src/cubeb_resampler_internal.h
```diff
@@ -288,15 +288,18 @@ template <typename T> class cubeb_resampler_speex_one_way : public processor {
...
-        resampled_frames_left;
-    if (input_frames_needed < 0) {
+    float input_frames_needed_frac =
+        static_cast<float>(output_frame_count) * resampling_ratio;
+    if (input_frames_needed_frac < 0) {
       return 0;
     }
-    return (uint32_t)ceilf(input_frames_needed);
+    auto input_frame_needed =
+        static_cast<uint32_t>(ceilf(input_frames_needed_frac));
```
karlt, line 297: I would be inclined to try:
```c
    // speex_resample()` can be irregular in its consumption of input samples.
    // Provide one more frame than the number that would be required with
    // regular consumption, to make the speex resampler behave more regularly,
    // and so predictably.
    auto input_frame_needed =
        1 + static_cast<uint32_t>(ceilf(input_frames_needed_frac));
```
I haven't proven that's sufficient, but I'm hoping that means that `std::min(unresampled_frames_left, input_frame_needed)` can always be subtracted and so the number of frames passed to the resampler does not depend on the number in `resampling_in_buffer`.

That might also happen to provide some safety from `float` arithmetic, but that's not the reason for the +1.

karlt (commented):
I spent several hours trying to prove that a certain calculation would produce enough input for an output, but I have not yet come up with something concrete.

The resampler advance can be understood from the [code](https://searchfox.org/mozilla-central/rev/a965e3c683ecc035dee1de72bd33a8d91b1203ed/media/libspeex_resampler/src/resample.c#591-596) in `resampler_basic_zero()` (or any of the other `resampler_basic_func`s).

The resampler can be considered to count at `num_rate * den_rate` quantums per
second, where `den_rate = out_rate / GCF(in_rate, out_rate)` and `num_rate =
in_rate / GCF(in_rate, out_rate)`.

`samp_frac_num` counts in these quantums.
Each `last_sample` unit represents `den_rate` quantums.
Adding the two together gives a position.

Each resampler iteration advances by `num_rate` quantums (`1/den_rate` seconds), and produces one output frame.

Initially the first input and output frames correspond to 0 quantums.

Each `speex_resample()` call leaves the resampler at a position exactly aligned with an output frame, so `output_frame_count` output frames requires an advance of `output_frame_count * num_rate` quantums.

Input frames intervals are `den_rate` quantums, but the first input frame is aligned differently on different `speex_resample()` calls, and there is an additional complicating factor that the speex resampler will sometimes consume input frames that it doesn't actually use.

`last_sample` values corresponds to input frame positions, but `last_sample` can get ahead of the number of input frames consumed.  In the last iteration of the `resampler_basic_func`, `last_sample` can [advance beyond](https://searchfox.org/mozilla-central/rev/a965e3c683ecc035dee1de72bd33a8d91b1203ed/media/libspeex_resampler/src/resample.c#588,591,596) `*in_len`.  The input frames stepped over are not needed yet.  [Input frames are consumed](https://searchfox.org/mozilla-central/rev/a965e3c683ecc035dee1de72bd33a8d91b1203ed/media/libspeex_resampler/src/resample.c#907-908) up to `last_sample`, even the frames that weren't needed.

The consuming of some unused input frames aligns better with the expected ratio of input to output frames given the alignment at 0 quantums.  When `last_sample` advances beyond `*in_len`, those extra frames are pulled on the next `speex_resample()` call, pulling more frames than expected.

src/cubeb_resampler_internal.h
karlt: That still passes the test at least.

padenot: That works, and if we subtract the number of frames we haven't resampled yet (remaining from last time), there is no increase of the sizes. This is now tested.

padenot:
I'll probably add some test that checks that the output is sensible, iirc we have an utility for that somewhere (feed a sine wave at a certain freq, check that we get that at the output).

src/cubeb_resampler_internal.h
padenot: I've amended the comment.

src/cubeb_resampler_internal.h
```diff
@@ -280,23 +280,25 @@ template <typename T> class cubeb_resampler_speex_one_way : public processor {
...
     int32_t resampled_frames_left =
         samples_to_frames(resampling_out_buffer.length());
-    float input_frames_needed =
-        (output_frame_count - unresampled_frames_left) * resampling_ratio -
-        resampled_frames_left;
-    if (input_frames_needed < 0) {
-      return 0;
-    }
-    return (uint32_t)ceilf(input_frames_needed);
+    output_frame_count -= resampled_frames_left;
```
karlt, line 322: I don't think we want  we want `resampled_frames_left` in the equation because `resampling_out_buffer` does not carry over from one `output()` operation to the next.

Not sure, but `input_needed_for_output()` may not be used with the `output()` method that uses `resampling_out_buffer`.

src/cubeb_resampler_internal.h
```diff
@@ -280,23 +280,25 @@ template <typename T> class cubeb_resampler_speex_one_way : public processor {
   }
 
   /** Returns the number of frames to pass in the input of the resampler to have
-   * exactly `output_frame_count` resampled frames. This can return a number
-   * slightly bigger than what is strictly necessary, but it guaranteed that the
-   * number of output frames will be exactly equal. */
+   * to at least have `output_frame_count` resampled frames. */
```
karlt, line 314: "to" and "have" repeated. One set can be removed.

src/cubeb_resampler_internal.h
```diff
@@ -280,23 +280,25 @@ template <typename T> class cubeb_resampler_speex_one_way : public processor {
...
+    output_frame_count -= resampled_frames_left;
+    float input_frames_needed_frac =
+        static_cast<float>(output_frame_count) * resampling_ratio;
+    // speex_resample()` can be irregular in its consumption of input samples.
+    // Provide one more frame than the number that would be required with
+    // regular consumption, to make the speex resampler behave more regularly,
+    // and so predictably.
+    auto input_frame_needed =
+        1 + static_cast<uint32_t>(ceilf(input_frames_needed_frac));
+    input_frame_needed -= unresampled_frames_left;
```
karlt, line 331: We don't want `input_frame_needed` underflowing when `unresampled_frames_left` is large.
Subtracting `std::min(unresampled_frames_left, input_frame_needed)` would be one way to deal with that.

test/test_resampler.cpp
```diff
@@ -6,7 +6,12 @@
  */
 #ifndef NOMINMAX
 #define NOMINMAX
+#include "cubeb/cubeb.h"
+#include "cubeb_log.h"
+#include "cubeb_resampler.h"
```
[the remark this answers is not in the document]
karlt: Not resolved.

karlt (approved):
Thanks!

src/cubeb_resampler_internal.h
padenot: Removed, and the test still passes, thanks.

test/test_resampler.cpp
```diff
@@ -1160,5 +1167,305 @@ TEST(cubeb, individual_methods)
...
+          char buf[256];
+          sprintf(buf, "test-%dHz-to-%dhz-%d-block.wav", source_rate,
+                  target_rate, block_size);
+          cubeb_audio_dump_stream_init(session, &dump_stream, out_params, buf);
+          cubeb_audio_dump_start(session);
+        }
+
+        cubeb_resampler * resampler = cubeb_resampler_create(
+            nullptr, nullptr, &out_params, source_rate, data_cb, &state,
+            CUBEB_RESAMPLER_QUALITY_DEFAULT, CUBEB_RESAMPLER_RECLOCK_NONE);
```
karlt, line 1395: Did you intentionally change this back to `_DEFAULT` in "Test improvements: check continuity"?

karlt (changes requested):
The large tolerance implies that either something is not right in the continuity test or what it is testing.

Would you like to drop that commit from the PR and deal with that separately, to get the draining fix landed?
Or do you think the continuity test is important to check that the drain fix is not breaking continuity?

test/test_resampler.cpp
```diff
@@ -1160,5 +1169,322 @@ TEST(cubeb, individual_methods)
...
+  for (size_t i = 0; i < len; ++i) {
+    float c = std::cos(phase_incr * static_cast<float>(i));
+    float s = std::sin(phase_incr * static_cast<float>(i));
+    sum_cos += signal[i] * c;
+    sum_sin += signal[i] * s;
+  }
+
+  float amplitude = 2.0f * std::sqrt(sum_cos * sum_cos + sum_sin * sum_sin) /
+                    static_cast<float>(len);
+  float phi = std::atan2(sum_sin, sum_cos);
```
karlt, line 1225: `phi = std::atan2(sum_sin, sum_cos)` is for `std::cos(phase_incr * i - phi)`.
I suspect `phi = std::atan2(sum_cos, sum_sin)` would be appropriate for `std::sin(phase_incr * i + phi)`.  That improves the MSE in some cases so that it is less than 0.0001.

In other cases the MSE is still close to 0.32 because the fitted amplitude is small.  The signal is not close to zero, which implies that the fitting of the sine is failing.  I don't know why that is.  Perhaps it might plausibly be due to accumulation of rounding causing phase variation in the source signal, but i have not observed clear evidence for or against that.

karlt (changes requested):
I'll hold off on looking at the threading until the test seems to be passing.

test/test_resampler.cpp
```diff
@@ -1160,5 +1167,305 @@ TEST(cubeb, individual_methods)
   ASSERT_EQ(frames_needed2, 0u);
 }
 
+struct sine_wave_state {
+  float phase = 0.0f;
```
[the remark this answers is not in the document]
padenot: I've been listening, performing spectrum analysis and zooming in on the time domain on the wave form, and it looks ok for our 1khz sine.

test/test_resampler.cpp
```diff
@@ -1160,5 +1167,305 @@ TEST(cubeb, individual_methods)
...
+  float * out = static_cast<float *>(output_buffer);
+  float phase_increment = 2.0f * M_PI * state->frequency / state->sample_rate;
+
+  for (int i = 0; i < nframes; i++) {
+    float sample = sin(state->phase);
+    state->phase += phase_increment;
+    if (state->phase > 2.0f * M_PI) {
+      state->phase -= 2.0f * M_PI;
+    }
+    out[i] = sample * 0.8;
```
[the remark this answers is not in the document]
padenot: I was concerned about slight overshoots caused by the resampling and was using external tooling that ended up doing float32 -> integer conversion on the `.wav` files dumped and there was some clipping, introducing distortion when I was plotting the spectrum, etc., so decided to simply lower the amplitude.

test/test_resampler.cpp
```diff
@@ -1160,5 +1167,305 @@ TEST(cubeb, individual_methods)
...
+    sse += diff * diff;
+  }
+
+  return sse;
+}
+
+// Finds the offset of the start of an input_freq sine wave sampled at
+// target_rate in data. Remove the leading silence from data.
+size_t
+find_sine_start(std::vector<float> & data, float input_freq, float target_rate)
```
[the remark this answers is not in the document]
karlt: This side-effect of changing the size of `data` is not undone for the subsequent iterations.  Perhaps using `skipped` as an offset for the insert might be clearer/easier than resizing `data`.

karlt (changes requested):
This is reassuring about the behavior of cubeb.
Once the identified issues in the test are resolved, the maximum MSE is 0.001920.
With more aggression stripping leading zeros, and perhaps also initial samples containing oscillations from the slope discontinuity, the maximum error could be even less.

karlt:
> With more aggression stripping leading zeros, and perhaps also initial samples containing oscillations from the slope discontinuity, the maximum error might be even less.

Because the resampler is a symmetric FIR filter, skipping twice the number of zero samples before the sine starts would skip the errors due to the slope discontinuity and cover over inaccuracies in finding the first sine sample.  I guess there's a choice between testing for precise residuals in most of the resampler output or accepting less precise residuals but testing the early samples in the sine.

test/test_resampler.cpp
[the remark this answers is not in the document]
padenot: I have changed this to be a pure function because of what you said, it's much nicer indeed.

test/test_resampler.cpp
padenot: Yes, I was trying things out and forgot to revert.

test/test_resampler.cpp
padenot: Oh yes what a silly mistake... I did lots of tries on this yesterday, including generating a perfect high resolution sine at the target rate, and calculating their MSE and comparing the value from the resampled sine and the "perfect" sine, and it was super high still on the "perfect" signal, so started to suspect an error in my math but didn't find the mistake.

test/test_resampler.cpp
```diff
@@ -1160,5 +1169,322 @@ TEST(cubeb, individual_methods)
...
+                resampler, nullptr, nullptr, data.data(), effective_block_size);
+            ASSERT_EQ(got, effective_block_size);
+            cubeb_resampler_stats stats = cubeb_resampler_stats_get(resampler);
+
+            // This roughly finds the start of the sine wave and strips it from
+            // `data`. This isn't stricly necessary but helps having cleaner
+            // dumps for manual inspection in e.g. Audacity. In all our test
+            // cases the resampler latency happens to be smaller than a block.
+            if (skipped == 0) {
+              skipped = find_sine_start(data, input_freq, target_rate);
```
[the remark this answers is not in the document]
padenot: This has been reworked to be a pure function and it lessens the complexity.

test/test_resampler.cpp
```diff
@@ -1160,5 +1167,305 @@ TEST(cubeb, individual_methods)
...
+  for (size_t i = 0; i < len; ++i) {
+    float c = std::cos(phase_incr * static_cast<float>(i));
+    float s = std::sin(phase_incr * static_cast<float>(i));
+    sum_cos += signal[i] * c;
+    sum_sin += signal[i] * s;
+  }
+
+  float amplitude = 2.0f * std::sqrt(sum_cos * sum_cos + sum_sin * sum_sin) /
+                    static_cast<float>(len);
+  float phi = std::atan2(sum_sin, sum_cos);
```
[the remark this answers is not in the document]
karlt: I don't know how significant the error here is, but the phase should constant so it is not necessary to perform the fit over the entire resampled duration.
That means that the duration of the fit can be selected so as to improve the accuracy of this `2/n * I` approximation.  The residuals can still be calculated over most of the resampled output (excluding the leading frames affected by zeros).

We want an integer number of 1/1000 s periods.  If a single period were selected, then at 44.1kHz, it would be covered by 44.1 samples.  We want an exact number of samples, so multiply the period by 10.  That makes the `2/n * I` approximation as accurate as it can be because multiplying the period further just duplicates the same contributions to D_0^T D_0.  I doubt larger periods would increase the accuracy of `D_0^T x` either, if the resampler is behaving as expected, because `x` should be periodic in the same way.

This is equivalent to a couple of the terms of a DFT.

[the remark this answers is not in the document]
karlt: I think that formulation would then be exact because the off-diagnal `sin.cos` terms should cancel out to be exactly zeros, and the components of the diagonal terms would be evenly balanced either side of 0.5.

test/test_resampler.cpp
```diff
@@ -1256,57 +1361,138 @@ TEST(cubeb, resampler_typical_uses)
...
+        // Code to print JSON to plot externally
+        // printf("\t[%d,%d,%d,%lf,%lf,%lf],\n", source_rate, target_rate,
+        //                  effective_block_size, mse, amplitude, phase);
+        double resampling_ratio =
+            static_cast<double>(source_rate) / target_rate;
+        // Accept larger error for large resampling ratio, and large blocks.
+        // The reason why large blocks result in higher MSE isn't yet well
+        // understood.
+        if (resampling_ratio >= 24) {
+          ASSERT_LT(mse, 0.18);
```
karlt: This is still a large error, given the mean sample energy is 0.32.

test/test_resampler.cpp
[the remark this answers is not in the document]
karlt:
```suggestion
  float phase = 0.0f;
```

test/test_resampler.cpp
```diff
@@ -1256,57 +1361,138 @@ TEST(cubeb, resampler_typical_uses)
...
+        float mse = sse / resampled.size();
+
+        // Code to print JSON to plot externally
+        // printf("\t[%d,%d,%d,%lf,%lf,%lf],\n", source_rate, target_rate,
+        //                  effective_block_size, mse, amplitude, phase);
+        double resampling_ratio =
+            static_cast<double>(source_rate) / target_rate;
+        // Accept larger error for large resampling ratio, and large blocks.
+        // The reason why large blocks result in higher MSE isn't yet well
+        // understood.
```
karlt: It is due to a large number of increments in the input phase.  For block size 2048 and resampling ratio 24, the number of input samples is ~ 49e6.  Single precision floating point has precision of ~ 1 / 16e6.  The errors add up to large phase differences.

test/test_resampler.cpp
```diff
@@ -1160,5 +1169,351 @@ TEST(cubeb, individual_methods)
...
+  sine_wave_state(float freq, int rate) : frequency(freq), sample_rate(rate) {}
+};
+
+long
+data_cb(cubeb_stream * stream, void * user_ptr, void const * input_buffer,
+        void * output_buffer, long nframes)
+{
+  sine_wave_state * state = static_cast<sine_wave_state *>(user_ptr);
+  float * out = static_cast<float *>(output_buffer);
+  float phase_increment = 2.0f * M_PI * state->frequency / state->sample_rate;
```
karlt, line 1185:
```suggestion
  double phase_increment = 2.0f * M_PI * state->frequency / state->sample_rate;
```

test/test_resampler.cpp
```diff
@@ -1160,5 +1169,351 @@ TEST(cubeb, individual_methods)
...
+
+  for (int i = 0; i < nframes; i++) {
+    float sample = sin(state->phase);
+    state->phase += phase_increment;
+    if (state->phase > 2.0f * M_PI) {
+      state->phase -= 2.0f * M_PI;
+    }
```
karlt, line 1192:
```suggestion
    float sample = sin(phase_increment * state->count);
    state->step_count ++;
```

test/test_resampler.cpp
```diff
@@ -1160,5 +1169,351 @@ TEST(cubeb, individual_methods)
...
+// https://www.youtube.com/watch?v=afQszl_OwKo and videos of the same series.
+// In practice here we're sending a perfect 1khz sine wave into a good
+// resampler, and despite the resampling ratio being quite extreme sometimes,
+// we're expecting a very good fit.
+float
+fit_sine(const std::vector<float> & signal, float sample_rate, float frequency,
+         float & out_amplitude, float & out_phase)
+{
+  const size_t len = signal.size();
+  float phase_incr = 2.0 * M_PI * frequency / sample_rate;
```
karlt, line 1212: Rounding errors are multiplied by the sample count.
```suggestion
  double phase_incr = 2.0 * M_PI * frequency / sample_rate;
```

test/test_resampler.cpp
```diff
@@ -1160,5 +1169,351 @@ TEST(cubeb, individual_methods)
...
+                    static_cast<float>(len);
+  float phi = std::atan2(sum_cos, sum_sin);
+
+  out_amplitude = amplitude;
+  out_phase = phi;
+
+  // Compute sum of squared errors relative to the fitted sine wave
+  float sse = 0.0;
+  for (size_t i = 0; i < len; ++i) {
+    float fit = amplitude * std::sin(phase_incr * i + phi);
```
karlt, line 1233: Worth using 0.8 for the amplitude, so that a small resampled output does not produce small residuals.

test/test_resampler.cpp
```diff
@@ -1256,57 +1361,138 @@ TEST(cubeb, resampler_typical_uses)
...
-                                             data.data(), block_size);
-          ASSERT_EQ(got, block_size);
+                                             data.data(), effective_block_size);
+          ASSERT_EQ(got, effective_block_size);
           cubeb_resampler_stats stats = cubeb_resampler_stats_get(resampler);
+
+          // This roughly finds the start of the sine wave and strips it from
+          // `data`. This isn't stricly necessary but helps having cleaner
+          // dumps for manual inspection in e.g. Audacity. In all our test
+          // cases the resampler latency happens to be smaller than a block.
```
karlt: The result of `find_sine_start()` is always less than one block because it won't increase `stripped` beyond `data.size() - POINTS`.  That's good because it ensures that always some samples will not be stripped.

However, the resampler latency is about 768 for the block size 96, ratio 1/24 case.

Find the sine in `resampled` so that all the latency can be removed.  This is contributing about 0.002 to MSE for the case above.

karlt (changes requested):
At CUBEB_RESAMPLER_QUALITY_DEFAULT, these changes reduce maximum MSE to
MSE=0.000119, effective_block=2048, resampling_ratio=2.756250, input-block=5644.800000, skipped=32

I haven't looked into why the start of the sine is not found for _VOIP.

test/test_resampler.cpp
```diff
@@ -1160,5 +1169,351 @@ TEST(cubeb, individual_methods)
...
+  float amplitude = 2.0f * std::sqrt(sum_cos * sum_cos + sum_sin * sum_sin) /
+                    static_cast<float>(len);
+  float phi = std::atan2(sum_cos, sum_sin);
+
+  out_amplitude = amplitude;
+  out_phase = phi;
+
+  // Compute sum of squared errors relative to the fitted sine wave
+  float sse = 0.0;
+  for (size_t i = 0; i < len; ++i) {
```
karlt, line 1232:
```suggestion
  for (size_t i = 0; i < signal.size(); ++i) {
```

test/test_resampler.cpp
```diff
@@ -1160,5 +1169,351 @@ TEST(cubeb, individual_methods)
...
+// engineers" chapter 19.1, and explained at
+// https://www.youtube.com/watch?v=afQszl_OwKo and videos of the same series.
+// In practice here we're sending a perfect 1khz sine wave into a good
+// resampler, and despite the resampling ratio being quite extreme sometimes,
+// we're expecting a very good fit.
+float
+fit_sine(const std::vector<float> & signal, float sample_rate, float frequency,
+         float & out_amplitude, float & out_phase)
+{
+  const size_t len = signal.size();
```
karlt, line 1211:
```suggestion
  // The formulation below is exact for samples spanning an integer number of periods.
  // The 10 is so that frequency=1000 divides exactly for sample_rate=44100.
  const size_t fit_len = std::min(signal.size(), (size_t)sample_rate * 10 / (size_t)frequency);
```

karlt (commented):
This would reduce the remaining MSEs to < 1e-6.
If you don't want to do this, then please explain the approximation and when it would be exact.

CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ if(BUILD_TESTS)
       message(FATAL_ERROR "Could not find googletest: run\n\tgit submodule update --init --recursive\nin base git checkout")
     endif()
     add_definitions(-DGTEST_HAS_TR1_TUPLE=0 -DGTEST_HAS_RTTI=0)
+    add_definitions(-DGTEST_ENABLED=1)
```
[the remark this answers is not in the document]
karlt: My config is different because `find_package(GTest)` gives
```
Found GTest: /lib64/cmake/GTest/GTestConfig.cmake (found version "1.14.0")
```
GTEST_ENABLED should be defined whether the system or subdirectory GTest package is used.

test/test_resampler.cpp
padenot: Done

test/test_resampler.cpp
padenot: That formula didn't quite work, so I did something simpler that does.

test/test_resampler.cpp
```diff
@@ -1160,5 +1169,342 @@ TEST(cubeb, individual_methods)
...
+// resampler, and despite the resampling ratio being quite extreme sometimes,
+// we're expecting a very good fit.
+float
+fit_sine(const std::vector<float> & signal, float sample_rate, float frequency,
+         float & out_amplitude, float & out_phase)
+{
+  // The formulation below is exact for samples spanning an integer number of
+  // periods.
+  const size_t fit_len =
+      (signal.size() / (sample_rate / frequency)) * (sample_rate / frequency);
```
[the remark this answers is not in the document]
padenot: I have restored using `signal.size()` but have added a comment about low iteration count.

PR #822, opened 2025-07-31
Remove `cubeb_stream_get_current_device`

marcin-serwin:
While looking into `cubeb_stream_get_current_device` I've noticed some issues:
* It seems like the function is supposed to return strings but `src/cubeb_audiounit.cpp` returns non-native endian integers. This could crash `cubeb-rs` since [it asserts](https://github.com/mozilla/cubeb-rs/blob/f2ea00d57e71a4bb7c924f15b523838925bb34c4/cubeb-core/src/device.rs#L81) that the returned value is a valid UTF-8 string.
* Also in `src/cubeb_audiounit.cpp`, the function does not free resources in case of errors, e.g., if the call to `audiounit_get_default_device_name` fails then `*device` is not freed causing leak.
* The [`device_destroy`](https://github.com/mozilla/cubeb-pulse-rs/blob/e0ddf430162f9113a9a90dc52b349d273baa3ec5/src/backend/stream.rs#L850) function in `cubeb-pulse-rs` does not free the allocated strings, it only frees the `cubeb_device` so it always causes memory leaks.

Instead of fixing this issues one by one I figured that it may be beneficial to simplify the API usage and avoid the leaks by not allocating memory in the first place.

`audiounit_stream_get_current_device` is removed, it's not implemented in the Rust version so I assume it's not super necessary. Strings and the `cubeb_device` are returned directly instead of being allocated on the heap. This causes the destroying function to become unnecessary and it is removed in  the last commit.

This is an API and ABI break but code changes for clients should be minor, comparable to the ones present in `test/test_devices.cpp`.

I'm a bit unsure about the removal of the destroying function. It would be a noop in the current backends, however, in the future it may be actually necessary to allocate something for a backend. What do you think?


padenot:
(https://searchfox.org/mozilla-central/source/dom/media/GraphDriver.cpp#1143-1198)

PR #827, opened 2025-09-15
Only start log thread if using CUBEB_LOG_VERBOSE.

mutexlox-signal:
test/test_logging.cpp should already verify the case of turning logging between verbose and off in the `logging` test and `logging_stress` flips between disabled and normal.


padenot (approved):
r+ as discussed.

PR #841, opened 2026-04-20
Assorted resampler fixes

padenot:
This stems from BMO#2029067 but includes nice things, notably the allocation bits, that could be significant (depending on the input/output rate ratio). Also includes simplifications.

### DIFF
--- a/src/cubeb_resampler.cpp
+++ b/src/cubeb_resampler.cpp
@@ -131,16 +131,21 @@ template <typename T, typename InputProcessor, typename OutputProcessor>
 cubeb_resampler_speex<T, InputProcessor, OutputProcessor>::
     cubeb_resampler_speex(InputProcessor * input_processor,
                           OutputProcessor * output_processor, cubeb_stream * s,
-                          cubeb_data_callback cb, void * ptr)
+                          cubeb_data_callback cb, void * ptr,
+                          cubeb_resampler_direction direction)
     : input_processor(input_processor), output_processor(output_processor),
       stream(s), data_callback(cb), user_ptr(ptr)
 {
-  if (input_processor && output_processor) {
+  switch (direction) {
+  case cubeb_resampler_direction::DUPLEX:
     fill_internal = &cubeb_resampler_speex::fill_internal_duplex;
-  } else if (input_processor) {
+    break;
+  case cubeb_resampler_direction::INPUT:
     fill_internal = &cubeb_resampler_speex::fill_internal_input;
-  } else if (output_processor) {
+    break;
+  case cubeb_resampler_direction::OUTPUT:
     fill_internal = &cubeb_resampler_speex::fill_internal_output;
+    break;
   }
 }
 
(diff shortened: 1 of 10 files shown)


src/cubeb_resampler.cpp
```diff
@@ -273,23 +281,31 @@ cubeb_resampler_speex<T, InputProcessor, OutputProcessor>::fill_internal_duplex(
...
+      input_processor->input(in_buffer, *input_frames_count);
+      size_t frames_resampled = 0;
+      resampled_input = input_processor->output(output_frames_before_processing,
+                                                &frames_resampled);
+      *input_frames_count = frames_resampled;
+    } else {
+      output_frames_before_processing =
+          std::min(output_frames_before_processing, *input_frames_count);
+      resampled_input = in_buffer;
+    }
```
kinetiknz, line 322: Do we need a `*input_frames_count = output_frames_before_processing;` here?

src/cubeb_utils.h
```diff
@@ -150,16 +160,16 @@ template <typename T> class auto_array {
   /** Keeps the storage, but removes all the elements from the array. */
   void clear() { length_ = 0; }
 
-  /** Change the storage of this auto array, copying the elements to the new
-   * storage.
+  /** Ensure the storage can hold at least `new_capacity` elements, reallocating
+   * if needed. Never shrinks.
    * @returns true in case of success
    * @returns false if the new capacity is not big enough to accomodate for the
    *                elements in the array.
```
kinetiknz: This never returns false now, can probably simplify by making it a `void` return and clean up the docs.

src/cubeb_resampler_internal.h
```diff
@@ -556,67 +432,44 @@ cubeb_resampler_create_internal(cubeb_stream * stream,
...
+                                       : cubeb_resampler_direction::OUTPUT);
 
   if (input_resampler && output_resampler) {
     LOG("Resampling input (%d) and output (%d) to target rate of %dHz",
         input_params->rate, output_params->rate, target_rate);
     return new cubeb_resampler_speex<T, cubeb_resampler_speex_one_way<T>,
                                      cubeb_resampler_speex_one_way<T>>(
         input_resampler.release(), output_resampler.release(), stream, callback,
-        user_ptr);
+        user_ptr, cubeb_resampler_direction::DUPLEX);
```
kinetiknz: nit: pass `direction` here rather than hard-coded `DUPLEX` for consistency with the other branches below

kinetiknz (approved):
LGTM, thanks!

src/cubeb_resampler.cpp
padenot: No: all the frames are now consumed. The param isn't useful anymore, but I'm not removing it just now because we need to coordinate w/ the rust bindings.

padenot:
@kinetiknz Didn't change this technically, but I can't run format w/ clang-format-15 anymore on ubuntu 26.04 so bumped it and format.

kinetiknz:
LGTM thanks, let's get this merged.

PR #847, opened 2026-06-10
resampler: Bound delay_line::output reads to buffered frames

kinetiknz:
The pointer-returning delay_line::output copied frames_to_samples(frames_needed) samples out of delay_input_buffer unconditionally. When the delay line holds fewer frames than requested - as happens at stream startup or during a capture underrun in a duplex stream that resamples only its output side - this read past the end of the buffer.

Copy only the buffered samples, zero-pad the remainder, and report the number of frames actually consumed, mirroring the underrun handling already present in cubeb_resampler_speex_one_way::output.

### DIFF
--- a/src/cubeb_resampler_internal.h
+++ b/src/cubeb_resampler_internal.h
@@ -432,15 +432,24 @@ public:
    * hold onto the pointer. */
   T * output(uint32_t frames_needed, size_t * input_frames_used)
   {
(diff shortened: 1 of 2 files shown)


padenot:
We should merge https://github.com/mozilla/cubeb/pull/841 instead, no?

kinetiknz:
> We should merge #841 instead, no?

Good point, I thought this seemed familiar but I assumed #841 was already merged and this was missed or something... bad memory on my part!

PR #857, opened 2026-08-26
Fix hang while waiting for state changes.

aosmond:
In wait_for_state_change, it loops until the new state matches the desired state. Sometimes the state machine would skip the expected state to another. This would happen in aaudio_stream_stop_locked, where we were in STARTING, expected to move to STARTED, and then it would transition into STOPPING/STOPPED/DISCONNECTED instead. This will cause the thread to block forever.

This patch replaces wait_for_state_change into two new methods, with the existing callers updated appropriately:

wait_for_stable_state, which doesn't verify what state we expect to move into, but simply until we leave the transitional state to a stable state.

refresh_state, which updates the internal state and reports it without blocking.

--

This should fix [bug 1831033](https://bugzilla.mozilla.org/show_bug.cgi?id=1831033), where the MediaDecoderStateMachine was hung waiting on a response from the remoted cubeb forever. It fixes a slew of intermittent failures for test_autoplay_* variants under dom/media/autoplay/test/mochitest.

### DIFF
--- a/src/cubeb_aaudio.cpp
+++ b/src/cubeb_aaudio.cpp
@@ -387,21 +387,53 @@ struct cubeb {
   struct cubeb_stream streams[MAX_STREAMS];
 };
 
-// Returns when aaudio_stream's state is equal to desired_state.
+static aaudio_stream_state_t
+get_stable_state(aaudio_stream_state_t state);
+
+// Whether the stream is on its way to some stable state.
+static bool
+is_transitional_state(aaudio_stream_state_t state)
+{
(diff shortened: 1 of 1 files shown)


aosmond:
green try where the intermittents are: https://treeherder.mozilla.org/jobs?repo=try&revision=16f9a0de517264138c9dbd317006c5806ba5c37b

aosmond:
@kinetiknz ^^

kinetiknz (approved):
LGTM, thanks!

